PR #170, opened 2020-08-11
array pop working with new Result-based signature

dfellis:
Part of work on #80

### DIFF
--- a/runtime/src/vm/memory.rs
+++ b/runtime/src/vm/memory.rs
@@ -268,14 +268,23 @@ impl HandlerMemory {
   }
 
   /// removes the last value of the array in the address and returns it
-  pub fn pop_fractal(self: &mut HandlerMemory, addr: i64) -> Result<i64, HandlerMemory> {
+  pub fn pop_fractal(self: &mut HandlerMemory, addr: i64) -> Result<HandlerMemory, String> {
+    // There's probably a more elegant way of doing this, but...
     let arr = &mut self.fractal_mem[self.either_mem[addr as usize] as usize];
-    let decision = arr.either_mem.pop().unwrap();
-    if decision < 0 {
-      return Ok(arr.mem.pop().unwrap());
+    if arr.mem.len() == 0 {
+      return Err("cannot pop empty array".to_string());
     } else {
-      arr.mem.pop();
-      return Err(arr.fractal_mem.pop().unwrap());
+      let decision = arr.either_mem.pop().unwrap();
+      if decision < 0 {
+        let mut frac = HandlerMemory::new(None, 1);
+        frac.write_fixed(0, arr.mem.pop().unwrap());
+        // This is a really shitty side-channel signal that it's fixed data
+        frac.either_closure_mem[0] = 1;
+        return Ok(frac);
+      } else {
+        arr.mem.pop();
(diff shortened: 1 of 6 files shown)


runtime/src/vm/opcode.rs
```diff
@@ -1828,10 +1828,38 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
   });
   cpu!("poparr", |args, hand_mem, _, _| {
     let last = hand_mem.pop_fractal(args[0]);
+    hand_mem.new_fractal(args[2]);
```
depombo, line 1831: Non blocking, but I wonder if it would be possible to DRY some of the Result/Maybe logic from opcodes

runtime/src/vm/memory.rs
```diff
@@ -268,14 +268,23 @@ impl HandlerMemory {
...
+    if arr.mem.len() == 0 {
+      return Err("cannot pop empty array".to_string());
     } else {
-      arr.mem.pop();
-      return Err(arr.fractal_mem.pop().unwrap());
+      let decision = arr.either_mem.pop().unwrap();
+      if decision < 0 {
+        let mut frac = HandlerMemory::new(None, 1);
+        frac.write_fixed(0, arr.mem.pop().unwrap());
+        // This is a really shitty side-channel signal that it's fixed data
```
depombo, line 281: Why is it shitty?

runtime/src/vm/memory.rs
```diff
@@ -268,14 +268,23 @@ impl HandlerMemory {
...
-    if decision < 0 {
-      return Ok(arr.mem.pop().unwrap());
+    if arr.mem.len() == 0 {
+      return Err("cannot pop empty array".to_string());
     } else {
-      arr.mem.pop();
-      return Err(arr.fractal_mem.pop().unwrap());
+      let decision = arr.either_mem.pop().unwrap();
+      if decision < 0 {
+        let mut frac = HandlerMemory::new(None, 1);
```
depombo, line 279: Hmm we can probably do this in other places actually

runtime/src/vm/opcode.rs
dfellis: Oh, DRYing up a lot of stuff is on my TODO list. Most of the duplication surrounds handling Pascal-style strings in HandlerMemory objects. I hope to centralize it so we can eventually find a more efficient way to do it.

The other part of DRYing it up is with Result and Maybe types being replicated within the opcodes, yes, so making helper functions for that is probably good. But DRY also makes it harder to "do something like this, but slightly different" so I prefer to duplicate for a bit and then DRY it up afterwards once the patterns emerge.

runtime/src/vm/memory.rs
dfellis: It's an abuse of the closure arguments to signal if the HandlerMemory should be considered fixed or not. There's probably a better way to do this.

runtime/src/vm/memory.rs
dfellis: That's true, and then we could simplify the opcodes, but lower priority, I think?

depombo: yes, not blocking either

PR #172, opened 2020-08-12
Update README.md

dfellis:
It has one dependency, now, but it is pure JS.


depombo:
While we are at it, let's also add the bit on the top level README about wanting to make sure contributions to the compiler stay open, but all libraries are Apache 2 licensed so you can use them in your proprietary codebases

dfellis:
This is a different README, though. Just for the `js-runtime`, not the outermost README.

I can do that in a different PR, though? (I did it by just clicking the "edit" button on the README, I don't have this branch local.)

PR #173, opened 2020-08-12
Reduce, Filter, Concat

dfellis:
The final PR to resolve \#80

Our opcode closure generation is very broken, though :(

### DIFF
--- a/bdd/spec/011_array_spec.sh
+++ b/bdd/spec/011_array_spec.sh
@@ -403,7 +403,7 @@ true"
           // TODO: Lots of improvements needed for closures passed directly to opcodes. This one-liner is ridiculous
           test.reduce(fn (acc: string, i: int): string = (acc == '' && i % 2 == 1) ? i.toString() : (i % 2 == 1 ? (acc + ', ' + i.toString()) : acc), '').print()
           // TODO: Even more ridiculous when you want to allow parallelism
-          test.reducePar(fn (acc: string, i: int): string = (acc == '' && i % 2 == 1) ? i.toString() : (i % 2 == 1 ? (acc + ', ' + i.toString()) : acc), fn (acc: string, cur: string): string = acc + ', ' + cur, '').print()
+          test.reducePar(fn (acc: string, i: int): string = (acc == '' && i % 2 == 1) ? i.toString() : (i % 2 == 1 ? (acc + ', ' + i.toString()) : acc), fn (acc: string, cur: string): string = (acc != '' && cur != '') ? (acc + ', ' + cur) : (acc != '' ? acc : cur), '').print()
 
           emit exit 0
         }
@@ -433,7 +433,6 @@ reduce as filter and concat test
     End
 
     It "runs agc"
-      Pending array-support
       When run alan-runtime run temp.agc
       The output should eq "$ADVARRAYOUTPUT"
     End
--- a/bdd/spec/011_array_spec.sh
+++ b/bdd/spec/011_array_spec.sh
@@ -260,14 +260,14 @@ Hello, World!"
     End
   End
 
-  Describe "repeat and mapl"
+  Describe "repeat and mapLin"
     before() {
       sourceToAll "
         from @std/app import start, print, exit
 
         on start {
           const arr = [1, 2, 3] * 3
-          const out = arr.mapl(fn (x: int64): string = x.toString()).join(', ')
+          const out = arr.mapLin(fn (x: int64): string = x.toString()).join(', ')
           print(out)
           emit exit 0
         }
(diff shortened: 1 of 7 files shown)


runtime/src/vm/memory.rs
```diff
@@ -16,12 +16,12 @@ pub struct HandlerMemory {
   /// Global memory reference
   gmem: &'static Vec<u8>,
   /// Memory of the handler for fixed size data types
-  mem: Vec<i64>,
+  pub mem: Vec<i64>,
```
depombo, line 19: Do we want to go back to making these private at some point or we are ok making all of the memory fields public?

std/root.ln
```diff
@@ -321,14 +321,38 @@ export fn push(arr: Array<bool>, val: bool) {
...
-export each // opcode with signature `fn each(arr: Array<any>, cb: function): void`
+export each // parallel opcode with signature `fn each(arr: Array<any>, cb: function): void`
+export fn eachLin(arr: Array<any>, cb: function): void = eachl(arr, cb)
 export map // parallel opcode with signature `fn map(arr: Array<any>, cb: function): Array<any>`
-export mapl // serial opcode with signature `fn mapl(arr: Array<any>, cb: function): Array<any>`
-export reduce // opcode with signature `fn reduce(arr: Array<any>, cb: function): any`
+export fn mapLin(arr: Array<any>, cb: function): Array<anythingElse> = mapl(arr, cb)
+export fn reduce(arr: Array<any>, cb: function): any = reducel(arr, cb)
+export fn reducePar(arr: Array<any>, cb: function): any = reducep(arr, cb)
+type InitialReduce<T, U> {
```
depombo, line 330: let's leave a comment explaining the different types of reduce and reducePar signatures?

dfellis: Sounds good.

runtime/src/vm/memory.rs
dfellis: I personally prefer having these fields public. There are a couple of methods in here that support only a single opcode because that logic couldn't live in the opcode with the struct fields being private. That causes `HandlerMemory` to have to house logic that it itself "should care about" and makes the separation of concerns blurry.

Of course, public fields can also cause issues in the other direction, which is why I'm not pushing for making everything public. All of the logic involving global memory addresses makes fields involving that pretty convoluted and something the opcodes should *not* need to deal with.

bdd/spec/011_array_spec.sh
```diff
@@ -373,27 +373,37 @@ true"
...
-          (test + test2).each(print)
+          test.concat(test2).map(fn (val: int64): string {
+            return toString(val)
+          }).join(', ').print()
+          (test + test2).map(fn (val: int64): string {
+            return toString(val)
+          }).join(', ').print()
+
+          print('reduce as filter and concat test')
+          // TODO: Lots of improvements needed for closures passed directly to opcodes. This one-liner is ridiculous
```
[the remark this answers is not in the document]
dfellis: So, closures passed directly to opcodes cannot use any opcodes themselves that are IO-bound, or trigger opcode stack changes like `condfn` does. That means conditionals are out, but also something like `urls.map(fn (url: string) = http.get(url))` will *also* not work.

Then there's the issue with return types being required for these closures but not required anywhere else (including closures passed in to user-defined functions) which makes for weird error messages.

Then there's the issue of closures being converted to microstatements *too early* so the microstatement types may be wrong, making it nearly impossible to write a closure used by multiple opcodes.

Then there's the issue of closure selection not occurring correctly (it just picks the first match by name if passed in by name instead of inline declared) so you can't do something like `arr.each(print)` because it will choose the "wrong" print function and fail.

Then there's the issue of not being able to use an opcode itself as your closure function, so `const sum = arr.reduce(add)` doesn't work.

"Lots of improvements". Each of these should probably be turned into a Github Issue, but I don't want all of that in the code comments. :/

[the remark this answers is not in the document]
depombo: Wow, got it. It is good that we have it in this PR at least.

runtime/src/vm/memory.rs
depombo: I agree that global memory should not be public, it is just becoming harder to understand where abstraction is drawn. e.g. what should be public and what shouldn't.

dfellis: True. I want to do an RFC at some point in the future (after launch) where we go back and streamline the memory model with both performance and clarity. Similar to the "remove ANTLR from the compiler" work that's nice to have but doesn't block functionality.

depombo: 💯 agree!

PR #174, opened 2020-08-13
README rewrite

dfellis:


README.md
```diff
@@ -1,8 +1,47 @@
...
+npm i -g alan-js-runtime
+```
+
+To install `alan-runtime`, run:
+
+```bash
+cargo install alan-runtime
+```
+
+### Source Installation
```
depombo, line 41: I would mention this is the installation for developing or contributing to alan

README.md
```diff
@@ -42,4 +106,24 @@ make bdd testfile=bdd/spec/001_event_spec.sh
...
+
+If one does need to make changes to the compiler or runtime, obeying the terms is as simple as merely forking this repository and making the changes there. You do not even need to try to upstream those changes. If you use the compiler or runtime without changes, all that is required of you is to provide a link to this repository *if you are asked for it.*
+
+We believe this structure will encourage the greatest growth in the ecosystem. Any kind of software, from fully-freely licensed (Apache, BSD, MIT) to reciprocally-freely licensed ([A]GPL 2/3) to proprietary can be developed in the Alan language, and there are no concerns that Alan Technologies will try to apply patent infringement lawsuits on you, while the compiler and primary runtime are reciprocally-freely licensed to ensure advancements from all parties are available to be used.
```
depombo, line 129: Such a long, unrequested explanation would be off putting to me as a developer, unless it was within a license FAQ. I would also not mention suing or any form legal action, even if it is to guarantee that it won't happen, anywhere throughout. Also off putting if they don't fully understand the licensing world and I believe our target user does not. 

Most people wouldn't make the effort to fully understand it either, but just dismiss it. I would also be more explicit about informing the user what GCC is and why it's licensing is relevant. Our target user will probably not know about GCC's license. I didn't. I was thinking about something more concise such as:

The Apache 2.0 license is freely combinable with the GPL 3 series of licenses as well as with proprietary software, so the standard library and Javascript runtime library are freely combinable with your own software projects without any requirement to open source it.

The AGPL 3.0 license requires that any changes to the code are published and publicly accessible. This is to make sure that any advancements to the compiler and runtime are available to all. The licensing of these tools does *not* affect the licensing of the code they compile or run. Similarly the [GCC compiler collection](https://gcc.gnu.org/) is GPL 3.0 licensed, but code compiled by it is not GPL 3.0 licensed.

alan
```diff
@@ -1,77 +0,0 @@
-#!/bin/bash
```
depombo, line 1: nice!

README.md
dfellis: Sounds good.

PR #182, opened 2020-08-14
Convert registers from tree traversal to pointers

dfellis:
This was originally going to also include the refactoring of the 'resfrom' opcode, but I decided it's a big enough and complex enough change that it should stand on its own right. Supports work for \#164

### DIFF
--- a/runtime/src/vm/memory.rs
+++ b/runtime/src/vm/memory.rs
@@ -8,6 +8,11 @@ use crate::vm::program::Program;
 // -2^63
 pub const CLOSURE_ARG_MEM_START: i64 = -9223372036854775808;
 pub const CLOSURE_ARG_MEM_SIZE: usize = 4;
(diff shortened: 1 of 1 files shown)


runtime/src/vm/memory.rs
```diff
@@ -9,6 +8,9 @@ use crate::vm::program::Program;
 // -2^63
 pub const CLOSURE_ARG_MEM_START: i64 = -9223372036854775808;
 pub const CLOSURE_ARG_MEM_SIZE: usize = 4;
+const NORMAL_ADDR: i8 = 0;
+const FX_PTR_ADDR: i8 = 1;
+const HM_PTR_ADDR: i8 = 2;
```
depombo, line 13: nit: can we add a small comment of how each works?

dfellis: Sure.

PR #183, opened 2020-08-15
HashMap on main runtime

dfellis:
Completely reworks registers to be pointers and closure memory as a nested HandlerMemory.

It's working, It's working!

Resolves \#164

### DIFF
--- a/runtime/src/vm/memory.rs
+++ b/runtime/src/vm/memory.rs
@@ -24,21 +24,18 @@ pub struct HandlerMemory {
   gmem: &'static Vec<u8>,
   /// Memory of the handler for fixed size data types
   pub mem: Vec<i64>,
-  /// Memory space used for closure arguments
-  closure_args: Vec<i64>,
-  fractal_closure_args: Vec<HandlerMemory>,
   /// Fractal memory storage for variable-length data types to an instance of the same struct
   pub fractal_mem: Vec<HandlerMemory>,
-  /// Helper for fractal_mem or fractal_closure_args to lookup the actual location of the relevant data, since
-  /// instantiating HandlerMemory instances for each memory usage need is expensive. -1 means the
-  /// data is in mem, otherwise it's mapping to the index in fractal_mem that houses the relevant
-  /// data.
+  /// Helper for fractal_mem to lookup the actual location of the relevant data, since instantiating
+  /// HandlerMemory instances for each memory usage need is expensive. -1 means the data is in mem,
+  /// otherwise it's mapping to the index in fractal_mem that houses the relevant data.
   either_mem: Vec<i64>,
-  pub either_closure_mem: Vec<i64>,
   /// Flag indicating the memory stored is actually a pointer
   registers_ish: Vec<i8>,
   /// Temporary hack
   pub is_fixed: bool,
+  /// Memory space used for closure arguments
+  closure_args: Vec<HandlerMemory>, // Option doesn't work for some reason
 }
 
 impl fmt::Display for HandlerMemory {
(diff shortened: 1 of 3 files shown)


runtime/src/vm/opcode.rs
```diff
@@ -1724,8 +1724,8 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
   cpu!("copytov", |args, hand_mem, _, _| {
     // args = [arr_addr, inner_addr, outer_addr]
     // copy data from outer_addr to inner_addr in arr_addr
-    let inner = hand_mem.read_fixed(args[2]);
-    hand_mem.copy_to_fractal_mem(args[0], args[1], inner);
+    let inner = hand_mem.read_fixed(args[1]);
+    hand_mem.copy_to_fractal_mem(args[0], args[2], inner);
```
dfellis, line 1728: This only worked before when we had no pointers because the only thing we ever copy is an empty string, which "looked" the same as fixed data.

runtime/src/vm/memory.rs
```diff
@@ -24,21 +24,18 @@ pub struct HandlerMemory {
...
-  /// data is in mem, otherwise it's mapping to the index in fractal_mem that houses the relevant
-  /// data.
-  either_mem: Vec<i64>,
-  pub either_closure_mem: Vec<i64>,
+  /// Helper for fractal_mem to lookup the actual location of the relevant data, since instantiating
+  /// HandlerMemory instances for each memory usage need is expensive. -1 means the data is in mem,
+  /// otherwise it's mapping to the index in fractal_mem that houses the relevant data.
+  pub either_mem: Vec<i64>,
   /// Flag indicating the memory stored is actually a pointer
   registers_ish: Vec<i8>,
```
depombo, line 34: nit: `register_ptrs` are probably a better name now

runtime/src/vm/memory.rs
```diff
@@ -305,7 +403,7 @@ impl HandlerMemory {
         let mut frac = HandlerMemory::new(None, 1);
         frac.write_fixed(0, arr.mem.pop().unwrap());
         // This is a really shitty side-channel signal that it's fixed data
```
depombo, line 405: is this still shitty though or ideally how would we do it?

runtime/src/vm/memory.rs
```diff
@@ -111,67 +106,171 @@ impl HandlerMemory {
...
-      self.mem[reg] = ptr;
+      let arr = self.get_mut_fractal(outer_addr);
+      arr.mem[0] = 0;
+      self.push_nested_fractal_mem(outer_addr, out);
     }
   }
 
   /// The address provided can be a directly nested fractal or a registerish address that points to
   /// a fractal. Either returns a reference to a HandlerMemory
   pub fn get_fractal(self: &HandlerMemory, addr: i64) -> &HandlerMemory {
```
depombo, line 255: can we collapse `read_fractal` and `get_fractal` what is the difference?

runtime/src/vm/memory.rs
dfellis: I'm not sure, yet, but I don't like the idea of a HandlerMemory object being spawned to house a single integer.

runtime/src/vm/memory.rs
dfellis: Yes, but I didn't want to introduce even more churn in this PR that's adding functionality. Would rather have a "cleanup" PR that intentionally changes no behavior for this.

runtime/src/vm/memory.rs
dfellis: `get_fractal` returns a reference to the actual HandlerMemory, `read_fractal` returns a copy of it. Sometimes one is useful versus the other. I do agree that I want to trim this API again, but I intentionally wanted to keep that to a "cleanup" PR as mentioned above where you know all of the changes are supposed to produce identical behavior.

PR #184, opened 2020-08-16
Add a 'clone' function that works for all data types. Built by taking over the 'assign' function and fixing some bugs in it

dfellis:
Resolves \#165


dfellis:
> awesome, let's update the docs too? :)

I am planning to do a docs update rollup soon-ish. :) Part of the delay was on whether or not I get the HashMap syntactic sugar implemented by Friday or not.

PR #188, opened 2020-08-18
Array of KeyVal pairs to HashMap converter function, useful for opcode interop

dfellis:
Resolves \#185

### DIFF
--- a/runtime/src/vm/opcode.rs
+++ b/runtime/src/vm/opcode.rs
@@ -730,6 +730,43 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
     None
   });
 
+  cpu!("absi8", |args, hand_mem, _, _| {
+    let a = hand_mem.read_fixed(args[0]) as i8;
+    let out = a.abs() as i64;
+    hand_mem.write_fixed(args[2], out);
+    None
+  });
+  cpu!("absi16", |args, hand_mem, _, _| {
+    let a = hand_mem.read_fixed(args[0]) as i16;
+    let out = a.abs() as i64;
+    hand_mem.write_fixed(args[2], out);
+    None
+  });
+  cpu!("absi32", |args, hand_mem, _, _| {
+    let a = hand_mem.read_fixed(args[0]) as i32;
+    let out = a.abs() as i64;
+    hand_mem.write_fixed(args[2], out);
+    None
+  });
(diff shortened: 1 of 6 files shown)


runtime/src/vm/opcode.rs
```diff
@@ -2036,28 +2073,28 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
     });
     None
   });
-  unpred_cpu!("eachl", |args, hand_mem, frag, ins_sched| {
-    let arr = hand_mem.get_fractal(args[0]);
+  unpred_cpu!("eachl", |args, mut hand_mem, frag, ins_sched| {
```
dfellis, line 2076: Don't ask me why Rust can infer that `hand_mem` has to be mutable in all the other opcodes but not here after my refactor. It doesn't make sense to me.

std/root.ln
```diff
@@ -129,6 +129,13 @@ export fn negate(n: int64) = negi64(n)
 export fn negate(n: float32) = negf32(n)
 export fn negate(n: float64) = negf64(n)
 
+export fn abs(n: int8) = absi8(n)
+export fn abs(n: int16) = absi16(n)
+export fn abs(n: int32) = absi32(n)
+export fn abs(n: int64) = absi64(n)
+export fn abs(n: float32) = absf32(n)
+export fn abs(n: float64) = absf64(n)
```
dfellis, line 137: A long time ago I had already defined these opcodes in the compiler, but forgot to implement them everywhere else, in case you're curious why `compile/src/lntoamm/opcodes.ts` didn't need a change.

std/root.ln
```diff
@@ -611,6 +617,19 @@ export fn newHashMap(firstKey: Hashable, firstVal: any): HashMap<Hashable, any>
   return hm.set(firstKey, firstVal)
 }
 
+export fn asHashMap(kva: Array<KeyVal<Hashable, any>>) {
```
dfellis, line 620: If you can think of a better name I'm all ears. :)

runtime/src/vm/opcode.rs
depombo: oh rust

std/root.ln
depombo: I prefer `toHashMap`

dfellis: Got it. Changing.

PR #227, opened 2020-08-20
Http client opcodes

depombo:


compiler/src/lntoamm/opcodes.ts
```diff
@@ -338,6 +338,8 @@ addopcodes({
   gtef32: [{ a: t('float32'), b: t('float32'), }, t('bool')],
   gtef64: [{ a: t('float64'), b: t('float64'), }, t('bool')],
   gtestr: [{ a: t('string'), b: t('string'), }, t('bool')],
+  httpget: [{ a: t('string')}, t('Result<any>')],
+  httppost: [{ a: t('string'), b: t('string')}, t('Result<any>')],
```
dfellis, line 342: This signature is *definitely* going to have to change once we add support for HTTP Headers.

runtime/src/vm/instruction.rs
```diff
@@ -48,14 +50,18 @@ impl InstructionScheduler {
...
-        return async_func(&ins.args, &mut hand_mem, &mut frag, self);
+        return async_func(ins.args.clone(), mem.clone());
       }).collect();
       task::spawn(async move {
-        // Poll futures concurrently, but not in parallel, using a single thread.
-        // This is akin to Promise.all in JavaScript Promises.
         join_all(futures).await;
+        let deref_res = Arc::try_unwrap(mem);
+        if deref_res.is_err() {
+          panic!("Arc for handler memory passed to io opcodes has more than one strong reference.");
```
dfellis, line 62: I hope we're never at this disco...

runtime/src/vm/opcode.rs
```diff
@@ -3063,9 +3063,123 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
   });
 
   // "Special" opcodes
```
dfellis, line 3065: Can we just rename this to "io opcodes"?

runtime/src/vm/opcode.rs
```diff
@@ -3063,9 +3063,123 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    let fut = async move {
+      let hand_mem = mem.read().await;
+      let ms = hand_mem.read_fixed(args[0]) as u64;
+      drop(hand_mem); // drop read lock
+      delay_for(Duration::from_millis(ms)).await;
+    };
+    return Box::pin(fut);
+  });
+  io!("httpget", |args, mem| {
+    let fut = async move {
```
dfellis, line 3076: bol

compiler/src/lntoamm/opcodes.ts
depombo: yes, I wanted to not have to think of it for now. Would we do the 3-arg trick and add a `Maybe<Array<KeyVal<Hashable, Any>>>`?

runtime/src/vm/run.rs
```diff
@@ -83,6 +83,7 @@ pub fn exec(fp: &str) {
   let mut rt = runtime::Builder::new()
     .basic_scheduler()
     .enable_time()
+    .enable_io()
```
dfellis, line 86: This one line contains multitudes.

runtime/src/vm/instruction.rs
dfellis: panic! at the disco

runtime/src/vm/opcode.rs
dfellis: futbol

runtime/src/vm/run.rs
dfellis: multitudes

compiler/src/lntoamm/opcodes.ts
dfellis: > yes, I wanted to not have to think of it for now. Would we do the 3-arg trick and add a `Maybe<Array<KeyVal<Hashable, Any>>>`?

The 3-arg trick is only when there's no return value.

We have to do the trick done for the 3 and 4 arg reduce functions where we wrap the arguments into a user-defined type and then pass that down.

dfellis: Also, on the type up there, the `KeyVal` would be a `KeyVal<string, string>` since we know the types involved, and we can probably eliminate the `Maybe` by just using multiple dispatch to have a function that does or does not define a header payload and update the opcode's request object appropriately.

runtime/src/vm/run.rs
dfellis: 1. It contains multitudes of source code behind it.
2. It turns on IO in Tok.io, an IO scheduling library. What was it before? Just Tok? (It contradicts itself, too!)

runtime/src/vm/instruction.rs
depombo: ahahah, yeah we should never go there..

runtime/src/vm/opcode.rs
depombo: yes! ⚽️ definitely comes to mind before 🏈 for me

dfellis: Well, in this case, because the `fut`ure is not a `foot`ure, otherwise definitely the latter ;)

runtime/src/vm/run.rs
depombo: that multitude alluded me for longer than I really wish it had...

bdd/spec/023_http_spec.sh
```diff
@@ -8,7 +8,7 @@ Describe "@std/http"
         from @std/http import get
 
         on start {
-          print(get('https://raw.githubusercontent.com/alantech/hellodep/aea1ce817a423d00107577a430a046993e4e6cad/index.ln'))
+          print(get('https://raw.githubusercontent.com/alantech/hellodep/aea1ce817a423d00107577a430a046993e4e6cad/index.ln').getOr('default'))
```
dfellis, line 11: This shouldn't be necessary, `print` can handle a `Result<Stringifiable>` just fine.

compiler/yarn.lock
```diff
@@ -113,6 +113,7 @@ ajv@^6.5.5:
 "alan-js-runtime@file:../js-runtime":
   version "0.0.1"
   dependencies:
+    isomorphic-fetch "^2.2.1"
```
depombo, line 116: shouldn't this be a `binary` file in `.gitattributes`?

bdd/spec/023_http_spec.sh
```diff
@@ -30,4 +30,34 @@ Describe "@std/http"
...
   End
+
+  Describe "basic post"
+    before() {
+      sourceToAll "
+        from @std/app import start, print, exit
+        from @std/http import post
+
+        on start {
+          print(post('https://reqbin.com/echo/post/json', '{"test":"test"}').getOr('default'))
```
dfellis, line 41: Two things here:

1. You need to escape the `"` inside of that JSON with backslashes.
2. You don't need the `.getOr('default')` part, again, because `print` can handle `Result<Stringifiable>` types.

bdd/spec/023_http_spec.sh
depombo: I like that

dfellis: Transparently handling `Result` makes a lot of things simpler :) And when we Result-ify integer math, that property is going to persist :) going to make sure the built-in functions for dealing with integers can handle raw integers or Result-wrapped ones, so you can do some math and then print it and see if it worked or returned an error in the process no problem.

compiler/yarn.lock
dfellis: Yeah, probably. :) I just always forget that sort of thing.

bdd/spec/023_http_spec.sh
```diff
@@ -52,12 +52,12 @@ Describe "@std/http"
 
     It "runs js"
       When run node temp.js
-      The output should eq '{"success":"true"}'
+      The output should eq "{\"success\":\"true\"}"
```
dfellis, line 55: So here you probably didn't need to do the escaping, but this is fine. :)

compiler/src/lntoamm/opcodes.ts
dfellis: Oh crap, how did I not noticed this? These opcodes are wrongly defined and `.getOr` was accidentally covering that up.

The return type should be `Result<string>` and there needs to be a line at the top of this file defining that type in opcode scope (around line 28)

dfellis: And we also need to make a ticket to figure out how `getOr` can override the return type instead of cause a compilation error here.

depombo: ah, got it. Just pushed the fix, do you want to create the issue since you have more context?

dfellis: > ah, got it. Just pushed the fix, do you want to create the issue since you have more context?

Yes, though I discussed last night in chat before I passed out what went wrong here. :) The fix is unfortunately something we have to do after some other already planned work is done.

Issue #156, opened 2020-08-03
Implement @std/http's http server using Node's built-in http server in js-runtime


dfellis:
Two bdd tests: one without headers and one for headers

PR #232, opened 2020-08-20
js-runtime HTTP Server

dfellis:
Eventually resolves \#156


std/http.ln
```diff
@@ -2,5 +2,103 @@
...
+      status = 200 // If not set by the user, assume they meant it to be good
+      headers = newHashMap('Content-Length', '0') // If not set by the user, assume no data
+      body = '' // If not set by the user, assume no data
+      connId = conn.connId
+    }
+  }
+}
+
+// The listen function tells the http server to start up and listen on the given port
+// For now only one http server per application, a macro system is necessary to improve this
```
depombo, line 76: nit: call this out as a TODO

std/http.ln
```diff
@@ -2,5 +2,103 @@
...
+// The send function converts the response object into an internal response object and passed that
+// back to the HTTP server. A Result type with either an 'ok' string or an error is returned
+export fn send(res: Response): Result<string> {
+  const ires = new InternalResponse {
+    status = res.status
+    headers = res.headers.keyVal
+    body = res.body
+    connId = res.connId
+  }
+  return httpsend(ires)
```
depombo, line 103: opcode?

dfellis: Yes, there are two opcodes and one event being added with this first-order-version of the HTTP server :)

std/http.ln
dfellis: I was actually looking for where we discussed the macro function idea to link to but I couldn't find it. Do you remember where we discussed that? I *thought* it was an RFC.

depombo: We discussed that before we had RFC, but I just checked the meta repo and it is not there. 🤔

compiler/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "test": "yarn run prepare && yarn run bundle && cypress run",
     "build": "tsc",
-    "prepare": "tsc &&  cp -r ../std std",
+    "prepare": "tsc && mkdir -p std && cp -r ../std/* ./std/",
```
depombo, line 8: yes, this definitely also got me a few times!

Issue #233, opened 2020-08-21
Implement @std/http with some performant Rust web server


dfellis:
Splitting off from #156 it's pretty big all on its own.

PR #239, opened 2020-08-25
Web server in rust runtime

dfellis:
Resolves \#233 and also cleans up the string manipulation logic in the opcodes.

### DIFF
--- a/runtime/src/vm/opcode.rs
+++ b/runtime/src/vm/opcode.rs
@@ -1,11 +1,10 @@
 use std::collections::HashMap;
-use std::convert::{Infallible, TryInto};
+use std::convert::Infallible;
 use std::future::Future;
 use std::hash::Hasher;
 use std::net::SocketAddr;
 use std::pin::Pin;
 use std::process::Command;
-use std::slice;
 use std::str;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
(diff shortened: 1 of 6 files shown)


runtime/src/vm/opcode.rs
```diff
@@ -1,25 +1,38 @@
...
 use tokio::time::delay_for;
 use twox_hash::XxHash64;
 
-use crate::vm::event::{EventEmit, HandlerFragment};
+use crate::vm::event::{BuiltInEvents, EventEmit, HandlerFragment};
 use crate::vm::instruction::InstructionScheduler;
 use crate::vm::memory::{CLOSURE_ARG_MEM_START, HandlerMemory};
+use crate::vm::run::{EVENT_TX};
+
+static HTTP_RESPONSES: Lazy<Arc<Mutex<HashMap<i64, HandlerMemory>>>> = Lazy::new(|| {
```
depombo, line 33: why is a Mutex necessary, wouldn't a RwLock suffice?

runtime/src/vm/opcode.rs
```diff
@@ -3074,110 +2560,188 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    let event_tx = EVENT_TX.get().unwrap();
+    if event_tx.send(event_emit).is_err() {
+      eprintln!("Event transmission error");
+      std::process::exit(3);
+    }
+    // Get the HTTP responses lock and periodically poll it for responses from the user code
+    // TODO: Swap from a timing-based poll to getting the waker to the user code so it can be
+    // woken only once and reduce pressure on this lock.
+    let responses = Arc::clone(&HTTP_RESPONSES);
+    let response_hm = poll_fn(|cx: &mut Context<'_>| -> Poll<HandlerMemory> {
```
depombo, line 2668: oh rust..

runtime/src/vm/opcode.rs
dfellis: https://stackoverflow.com/questions/50704279/when-or-why-should-i-use-a-mutex-over-an-rwlock#50704283

I am pretty sure HandlerMemory needs the extra safeguards. I am willing to be proven wrong here, though. I am also expecting us to eventually see this as a bottleneck and switch to dashmap or something like that.

runtime/src/vm/opcode.rs
dfellis: I am glad I found that thing, though! I wasn't sure how I was going to implement this without something of this nature.

And I should be able to make it more efficient (and not really a poll at all) in the future.

We could also have timeout support at some point to make sure that there aren't dead connections piling up.

PR #240, opened 2020-08-25
More compiler fixes

dfellis:
Discovered this this morning, the standard library is not being included in the published compiler so the globally-installed one doesn't work. Creating an [.npmignore](http://npm.github.io/publishing-pkgs-docs/publishing/the-npmignore-file.html) file appears to be the answer. Can only test by trying to publish this change, though.


dfellis:
```
damocles@Talyn:~/alantech/alan/compiler(more-compiler-fixes)$ alan-compile temp.ln temp.js
Done in 5.337s
damocles@Talyn:~/alantech/alan/compiler(more-compiler-fixes)$ node temp.js
Hello, World!
```

Confirmed that it now works.

PR #245, opened 2020-08-28
Add Windows CI testing, finally!

dfellis:


dfellis:
Going to have to abort this for now. It looks like we'll have to migrate the BDD tests from shellspec to something less UNIX-y, despite Bash for Windows existing. I cannot tell where or how the test suite is failing, or if shellspec itself really is getting stuck in a loop in Windows -- nothing is printed until I cancel the job.

Issue #249, opened 2020-08-31
alan-runtime run with an non existing file panics on a unwrap instead of an error message


LoPoHa:
If the alan-runtime gets called with a file that doesn't exist it unwraps (src\vm\run.rs:98:32)
I'm not sure how you want such errors to be handled so i didn't create a pr.
But something simple like .expect("The given file was not found") would be better than the unwrap here.

dfellis:
Absolutely something we should tackle! We have a [test just for compiler errors](https://github.com/alantech/alan/blob/main/bdd/spec/019_compiler_error_spec.sh), we should make a similar one for runtime errors.

PR #250, opened 2020-08-31
Better File-not-found error for the runtime

dfellis:
Resolves \#249


dfellis:
```
7.1) Unexpected output to stderr occurred

            main: line 11349: 12189 Terminated              node temp.js
```

Github Actions *really did* change the behavior around `kill`ing processes. It wasn't related to my changes in the Windows CI!

I am both relieved and annoyed. Relieved that I did the right thing with the Windows CI PR, but annoyed that I lost ~12 hours to fixing that!

Issue #251, opened 2020-09-01
Compilation error parsing `app.print(2 / (3))`


ogier:
How to reproduce:

1. Visit alan-lang.com.
2. Replace "Hello, World!" with `2 / (3)`.
3. Execute code snippet.

Result:

```
Cannot read property 'basicassignables' of null
```

dfellis:
Hi @ogier thank you for reporting this. I can confirm that if you put `2 / (1 + 2)` in there it *does* work, so I think I know what's going on here. :)

PR #254, opened 2020-09-01
Fix bug with operator groups of one value

dfellis:
Resolves \#251


runtime/Cargo.toml
```diff
@@ -5,7 +5,7 @@ license = "AGPL-3.0"
 homepage = "https://alan-lang.org"
 documentation = "https://docs.alan-lang.org"
 repository = "https://github.com/alantech/alan"
-version = "0.1.2"
+version = "0.1.3"
```
depombo, line 8: why do we need to bump the runtime?

dfellis: I made a change to the runtime yesterday but I forgot to version bump and publish it.

Issue #238, opened 2020-08-24
Implement shared mutable caching layer with implicit locking mechanisms


dfellis:
[008 - Safe Global Data Storage RFC](https://github.com/alantech/alan/blob/main/rfcs/008%20-%20Safe%20Global%20Data%20Storage%20RFC.md)

PR #256, opened 2020-09-02
@std/datastore (js-runtime only for now)

dfellis:
Beginning to implement RFC 008

Part 1 of 2 for #238

### DIFF
--- a/js-runtime/index.js
+++ b/js-runtime/index.js
@@ -459,9 +459,9 @@ module.exports = {
   hashf,
   hashv,
 
-  // In Node.js these opcodes don't have to be IO opcodes, but in the Rust runtime they do, because
-  // of the multithreaded nature of the Rust runtime. Not sure if they should be "fake" async here
-  // or not.
+  // In Node.js the datastore opcodes don't have to be IO opcodes, but in the Rust runtime they do,
+  // because of the multithreaded nature of the Rust runtime. Not sure if they should be "fake"
+  // async here or not.
   dssetf:  (ns, key, val) => {
     ds[`${ns}:${key}`] = val
   },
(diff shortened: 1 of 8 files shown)


js-runtime/index.js
```diff
@@ -456,6 +459,39 @@ module.exports = {
...
+      return [ false, 'namespace-key pair not found', ]
+    }
+  },
+  dsgetv:  (ns, key) => {
+    const fullKey = `${ns}:${key}`
+    if (ds.hasOwnProperty(fullKey)) {
+      return [ true, ds[`${ns}:${key}`], ]
+    } else {
+      return [ false, 'namespace-key pair not found', ]
+    }
```
depombo, line 492: non-blocking nit: is there any way to dry this up easily?

js-runtime/index.js
```diff
@@ -456,6 +459,39 @@ module.exports = {
   hashf,
   hashv,
 
+  // In Node.js these opcodes don't have to be IO opcodes, but in the Rust runtime they do, because
```
depombo, line 462: non-blocking nit: In Node.js these opcodes -> In Node.js the datastore opcodes

js-runtime/index.js
dfellis: Unfortunately, no, since they're defined inline with the object creation. I can't just use the prior definition. The only way is to restructure it *all* so I define the opcodes as separate variables and then have the object definition reference these variables when constructing said object.

This would *almost double* the size of the file since most of the time the opcodes are one-liners.

js-runtime/index.js
dfellis: Will fix

dfellis:
So, because the compiler has a dependency on the js-runtime (for the web bundle generated) I need to publish the js-runtime *before* landing this, then update the compiler to point at the published updated js-runtime. :/

Really need to figure out a better way to handle this.

PR #259, opened 2020-09-04
Add datastore support to the rust runtime

dfellis:
Resolves \#238


std/datastore.ln
```diff
@@ -6,14 +6,14 @@
...
 // The set function to store shared data
+export fn set(ns: string, key: string, val: any) = dssetv(ns, key, val)
 export fn set(ns: string, key: string, val: int8) = dssetf(ns, key, val)
 export fn set(ns: string, key: string, val: int16) = dssetf(ns, key, val)
 export fn set(ns: string, key: string, val: int32) = dssetf(ns, key, val)
 export fn set(ns: string, key: string, val: int64) = dssetf(ns, key, val)
 export fn set(ns: string, key: string, val: float32) = dssetf(ns, key, val)
 export fn set(ns: string, key: string, val: float64) = dssetf(ns, key, val)
 export fn set(ns: string, key: string, val: bool) = dssetf(ns, key, val)
-export fn set(ns: string, key: string, val: any) = dssetv(ns, key, val)
```
depombo, line 16: this a no-op?

depombo (approved):

std/datastore.ln
dfellis: No it's not. It really does matter. When multiple functions match a given set of argument types, *the most recently defined wins*. This is to make it possible for users to define their own overloading functions and have it actually called instead of the original in their own modules.

But it also means if you define multiple functions within the same module, the most general function should be defined first and the more specific later.

It didn't matter in the js-runtime since `dssetv` and `dssetf` do the same thing there, but it does matter for the rust runtime, and I caught that mistake pretty early on.

dfellis: We could have had a "more specific type wins" approach, but that would prevent overloading by users in some scenarios.

It's also ambiguous if one function has the first argument as an interface type and the second a specific type, and another function has the first argument a specific type and the second an interface type and both could apply for this call -- which is "more specific?"

depombo: That makes sense, thank you for explaining

Issue #260, opened 2020-09-04
Fix first stage compiler weirdness around declarations and assignments


dfellis:
This is something that has been a TODO for a while, and I was going to roll it up in the "Eliminate ANTLR" PR, but it's worse than I originally anticipated based on a report from @ogier

Currently the parser thinks this is valid:

```ln
import @std/app

on app.start {
    app.oops
}
```

This was already cleaned up in the AMM grammar file, so fixing that part shouldn't be hard, but there are a lot of references throughout the codebase to the current structure that will need cleanup.

PR #261, opened 2020-09-04
Fail hard on invalid statements

dfellis:
Resolves \#260


depombo (approved):
I love how many TODOs this PR takes care of :)

Issue #203, opened 2020-08-19
Make an `alan` command to wrap compiler, runtime, package management, and etc.


dfellis:
We recently removed the shell script that was doing this, but we probably want this to be a binary that can work across OSes, so likely should be rewritten in Rust or Alan itself.

PR #267, opened 2020-09-09
Start building an 'alan' command that wraps the compile and AVM together

dfellis:
Resolves \#203


dfellis:
Screenshot from 2020-09-09 14-42-40

depombo:
This is amazing! What if we have the `alan` command by default take a `.ln` file to compile and run? `alan test.ln` feels familiar coming from `python test.py` and `node test.js`

dfellis:
> This is amazing! What if we have the `alan` command by default take a `.ln` file to compile and run? `alan test.ln` feels familiar coming from `python test.py` and `node test.js`

You can have a default behavior if no subcommand is specified I think, so that should be fine. I don't want to do that, yet, because compilation is so slow, though?

depombo:
> > This is amazing! What if we have the `alan` command by default take a `.ln` file to compile and run? `alan test.ln` feels familiar coming from `python test.py` and `node test.js`
> 
> You can have a default behavior if no subcommand is specified I think, so that should be fine. I don't want to do that, yet, because compilation is so slow, though?

I think that is okay and we should do it anyway because it is in line with the ethos of this PR/task which is making Alan easier to use, or in other words, reducing the amount of steps to use it

depombo (approved):

avm/src/main.rs
```diff
@@ -0,0 +1,56 @@
+use std::env;
+
+use clap::{App, SubCommand, crate_name, crate_version};
+
+use crate::benchmark::run::benchmark;
+use crate::compile::compile::compile;
```
dfellis, line 6: compile compile compile

PR #286, opened 2020-09-18
Lazily and multiple-y dispatched closure functions

### LINKED ISSUE #194: Get closures passed to opcodes to use multiple dispatch
Currently a hacky solution that assumes the list of potential functions has a length of 1.

### LINKED ISSUE #195: Evaluate closure functions lazily instead of eagerly
The current behavior makes it impossible to define a closure function with interfaces instead of "normal" types, so it's impossible to use a generalized function like `print` as closure function (`arr.each(print)` doesn't work).

dfellis:
Resolves #194 and resolves #195 simultaneously. Wasn't able to cleanly separate them when I actually dug into it.

### DIFF
--- a/bdd/spec/011_array_spec.sh
+++ b/bdd/spec/011_array_spec.sh
@@ -401,9 +401,9 @@ true"
 
           print('reduce as filter and concat test')
           // TODO: Lots of improvements needed for closures passed directly to opcodes. This one-liner is ridiculous
-          test.reduce(fn (acc: string, i: int): string = (acc == '' && i % 2 == 1) ? i.toString() : (i % 2 == 1 ? (acc + ', ' + i.toString()) : acc), '').print()
+          test.reduce(fn (acc: string, i: int): string = ((acc == '') && (i % 2 == 1)) ? i.toString() : (i % 2 == 1 ? (acc + ', ' + i.toString()) : acc), '').print()
           // TODO: Even more ridiculous when you want to allow parallelism
-          test.reducePar(fn (acc: string, i: int): string = (acc == '' && i % 2 == 1) ? i.toString() : (i % 2 == 1 ? (acc + ', ' + i.toString()) : acc), fn (acc: string, cur: string): string = (acc != '' && cur != '') ? (acc + ', ' + cur) : (acc != '' ? acc : cur), '').print()
+          test.reducePar(fn (acc: string, i: int): string = ((acc == '') && (i % 2 == 1)) ? i.toString() : (i % 2 == 1 ? (acc + ', ' + i.toString()) : acc), fn (acc: string, cur: string): string = ((acc != '') && (cur != '')) ? (acc + ', ' + cur) : (acc != '' ? acc : cur), '').print()
 
           emit exit 0
         }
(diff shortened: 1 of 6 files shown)


bdd/spec/018_closure_spec.sh
```diff
@@ -55,7 +55,7 @@ Describe "Closure Functions"
 
         on start {
           const numbers = [1, 2, 3, 4, 5]
-          numbers.map(double).map(fn (x: int64): string = x.toString()).join(', ').print()
```
dfellis, line 58: I am *so* happy about being able to finally eliminate these weird wrapper functions!

depombo: amazing! :D

bdd/spec/011_array_spec.sh
```diff
@@ -401,9 +401,9 @@ true"
 
           print('reduce as filter and concat test')
           // TODO: Lots of improvements needed for closures passed directly to opcodes. This one-liner is ridiculous
-          test.reduce(fn (acc: string, i: int): string = (acc == '' && i % 2 == 1) ? i.toString() : (i % 2 == 1 ? (acc + ', ' + i.toString()) : acc), '').print()
```
dfellis, line 404: This change is a bit horrifying to me, because I don't understand how it compiled correctly before looking at the operator precedence table.

But that also makes me wonder if the precedence levels have been set correctly. The previous version looks better than this one.

dfellis:
### Summary of the changes

A new microstatement type is declared, `CLOSUREDEF` which simply marks that a closure function has been defined somewhere within the function being processed. The creation of a CLOSURE is now shifted to when the closure is used. This is required to unblock closures with interface types, because the type realization and the determination of which functions should be used within the closure can only be done at the call sites.

For closures used by other user functions that's all self-contained in the relatively straightforward Microstatement.ts changes.

Closures that are passed to the opcodes, however, require more work. Closures in user-defined code are basically inlined into the execution graph, and the multiple dispatch code will figure out which one "fits" if there are multiple implementations of that closure function name with different arguments.

For opcodes, though, the way the closure function is used is "hidden" from the rest of the compiler, so the primitive `function` type doesn't work there. The long-term solution is to eventually allow function types to work very similarly to function declarations within interfaces, but for now it's been patched together by hardwiring the decision logic based on the "class" of opcode that has `microstatementInlining` being called on it.

This is unfortunately further complicated by special logic already existing for opcodes to handle interfaces and generics, so there are three sets of similar-but-not-quite-the-same implementations of the interface-type-realization and multiple-dispatch-selection logic for these opcodes.

### Future Work

Eventual conversion to something more sane will take some effort, though, because some opcodes, like `map` and `each` accept mutliple function signatures, and figuring out how to handle that overloading (or deciding to undo that overloading and just give the different variants different opcode names) is yet-to-be-determined. There's also a giant question around how the two `recurse`-related opcodes will work within the type system. I wrote code that correctly pulls the necessary types, but it does require that the recursive function does *not* use interfaces and I don't know how to solve that.

There is also one small issue with closure functions and return type inference not working that I want to solve, but I would like to do that in a separate PR and close this off so I can release this already much-needed improvement to closures and continue working on other higher-priority issues.

depombo (approved):

PR #288, opened 2020-09-18
Update download link in readme

LoPoHa:
Always link to the latest release instead of a fixed one.


dfellis (approved):
Thank you so much! I wasn't aware of a 'latest' URL.

PR #291, opened 2020-09-26
Tree RFC

dfellis:


rfcs/009 - Tree RFC.md
```diff
@@ -0,0 +1,110 @@
...
+
+`reduce` is similarly interesting -- it could be parallelized by "dependency layer", all of the nodes with no children, then all of the nodes in the first layer above those, then the layer above that, etc, until you reach the root node. This means that serializing a data structure built on top of a Tree (like JSON, for instance) would be an automatically parallel operation and could therefore be more performant by default than other languages.
+
+`filter` could have two implementations, one that filters in a way that a failed filter node *and all of its children* are removed, and one where the failed nodes are deleted and their children (that pass the filter) are re-parented on the failed node's parent, where the latter would provide an array of potentially more than one tree if the root node is filtered away. What they should be called is not clear.
+
+`every` and `some` could simply be shorthand for something like `someTree.toNodeArray().every(...)` or just excluded instead.
```
depombo, line 67: seems weird, probably just excluded

rfcs/009 - Tree RFC.md
```diff
@@ -0,0 +1,110 @@
...
+This is the iffiest idea because it's done through an abuse of the operator syntax, but it looks pretty natural for defining tree constants. Fluent style isn't so bad, though:
+
+```ln
+const myTree = newTree("foo")
+  .addChild(newTree("bar")
+    .addChild("baz"))
+  .addChild("bay")
+```
+
+With the only downside being keeping track of the parens.
```
depombo, line 93: This should go first since this will be the default way of defining a tree initially

rfcs/009 - Tree RFC.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Author
+
+- David Ellis <david@alantechnologies.com>
+
+## Summary
+
+The ability to define and reason about data that has a tree-like structure is necessary to handle a host of different problems, from parsers and serializers, to manipulation of an HTML DOM or JSON object.
+
+This is an integral piece of any language. Most choose to allow recursive data types and then deal with the complexities around infinitely-defined types (one way or the other). For Alan, we instead define a new Tree type to handle this and provide a collection of tools to work with it.
```
depombo, line 30: I was hoping the rfc would include some code samples into what it would look like to code/solve these kinds of  problem statements with `Tree`. That seems more relevant that the short hard notation which takes on a considerable amount of the rfc.

rfcs/009 - Tree RFC.md
dfellis: The only one I'm not doing, because of offline discussions. Since we decided that mirroring the Array method names wherever possible was the right choice, excluding some array methods for no apparent reason doesn't seem right.

rfcs/009 - Tree RFC.md
dfellis: Added :)

depombo (approved):

Issue #224, opened 2020-08-19
Get closure opcodes to support handlers with multiple fragments and/or IO opcodes


dfellis:
Currently closures passed to array methods can only be purely computation, without IO or even conditional branching, so the following doesn't work:

```ln
fn loadUrls(urls: Array<string>): Array<string> {
  return urls.map(fn (url: string): string = http.get(url))
}
```

PR #292, opened 2020-10-01
Tree data structure (js-runtime only for now)

dfellis:


dfellis:
Getting this to work on the AVM requires tackling the tech debt mentioned in #224 as well as dealing with some issues with the memory model. I did some of the work on the AVM's memory model here but it demonstrated to me that just patching it over is going to make `memory.rs` even more difficult to follow, so I want to rework it entirely. Will create a new issue for that.

std/root.ln
```diff
@@ -631,6 +631,154 @@ export fn toHashMap(kva: Array<KeyVal<Hashable, any>>) {
...
+}
+
+export type Node<T> {
+  id: int64
+  tree: Tree<T>
+}
+
+export fn newTree(rootVal: any): Tree<any> = new Tree<any> {
+  vals = new Array<any> [ rootVal ]
+  parents = new Array<int64> [ -1 ]
```
depombo, line 648: nit - can we have a few short comments on how the initialization to -1 works and how `Node` and `Tree` relate to each other?

depombo (approved):

depombo (approved):

PR #295, opened 2020-10-13
add basic alan install

depombo:


avm/src/main.rs
```diff
@@ -10,6 +11,19 @@ mod benchmark;
 mod compile;
 mod vm;
 
+fn run(source_file: &str) {
```
dfellis, line 14: Why not put this in a separate source file like the others and just call it? I expect a *lot* of subcommands eventually, and mixing them and their dependencies together is gonna make it hard to follow.

dfellis (approved):

PR #296, opened 2020-10-14
Redo the AVM memory model

dfellis:
Currently very broken, but it's hit midnight and that's when I decided to call it. Will continue tomorrow after an appointment.

### DIFF
--- a/avm/src/vm/memory.rs
+++ b/avm/src/vm/memory.rs
@@ -206,7 +206,7 @@ impl HandlerMemory {
     let mem = self.read_fractal(fractal_addr);
     let (a, b) = mem[offset_addr as usize];
     if a < std::usize::MAX {
-      self.set_addr(addr, a, b.try_into().unwrap());
+      self.set_addr(addr, a, b as usize);
     } else {
       let (a, _) = self.addr_to_idxs(fractal_addr);
       self.set_addr(addr, a, offset_addr as usize);
--- a/avm/src/vm/memory.rs
+++ b/avm/src/vm/memory.rs
@@ -185,7 +185,12 @@ impl HandlerMemory {
   pub fn push_register(self: &mut HandlerMemory, addr: i64, other_addr: i64) {
     let (a, b) = self.addr_to_idxs(other_addr);
     let mem = self.read_mut_fractal(addr);
-    mem.push((a, b.try_into().unwrap()));
+    mem.push((a, b as i64));
+  }
+
+  pub fn push_idxs(self: &mut HandlerMemory, addr: i64, a: usize, b: usize) {
+    let mem = self.read_mut_fractal(addr);
+    mem.push((a, b as i64));
   }
 
   pub fn pop(self: &mut HandlerMemory, addr: i64) -> Result<(usize, i64), String> {
--- a/avm/src/vm/memory.rs
+++ b/avm/src/vm/memory.rs
@@ -29,21 +29,32 @@ fn addr_type(addr: i64) -> i8 {
 #[derive(Clone)]
 #[derive(Debug)]
 pub struct HandlerMemory {
-  /// The set of memory blocks. Each block consists of a tuple of two values. If the first value is
-  /// zero then the second value is the actual data, if it is non-zero then the two values together
-  /// represent a virtual pointer to another memory block and value (if the second value is zero,
-  /// it represents a pointer to a nested array of data, otherwise a pointer to an explicit value)
+  /// The set of memory blocks. The first (zeroth) block hosts global memory and all blocks
+  /// afterwards host memory created by the handler. Each block consists of tuples of two values,
+  /// representing either a virtual pointer or raw data, with three classes of values:
+  /// 1. `(usize::MAX, any value)` - The first value indicates that the second value is actual raw
+  ///    data
+  /// 2. `(< usize::MAX, usize::MAX)` - The first value indicates that this is a virtual pointer to
+  ///    nested memory. The second value indicates that the pointer is to an entire block of
+  ///    memory, not an explicit value.
+  /// 3. `(< usize::MAX, < usize::MAX)` - The first value indicates that this is a virtual pointer
+  ///    to nested memory. The second value indicates that the pointer is to an explicit value
+  ///    within that block of memory.
+  /// Virtual pointers are simply the indexes into the `mems` field.
   mems: Vec<Vec<(usize, i64)>>,
-  /// The address spaces for the handler memory. The first  is the "normal" memory space, and the
-  /// second is the args memory space.
+  /// The address spaces for the handler memory that the handler can mutate. The first is the
+  /// "normal" memory space, and the second is the args memory space. Global addresses are fixed
+  /// for the application and do not need a mutable vector to parse.
   addr: (Vec<(usize, usize)>, Vec<(usize, usize)>),
-  /// Specifies which memory block to push "normal" values into
+  /// Specifies which memory block to push "normal" values into.
   mem_addr: usize,
-  /// Specifies which memory block to push "args" into
+  /// Specifies which memory block to push "args" into.
   args_addr: usize,
 }
 
 impl HandlerMemory {
+  /// Constructs a new HandlerMemory. If given another HandlerMemory it simply adjusts it to the
+  /// expected memory needs, otherwise constructs a new one with said memory requirements.
   pub fn new(payload_mem: Option<HandlerMemory>, mem_req: i64) -> HandlerMemory {
     if payload_mem.is_none() {
       let mut hm = HandlerMemory {
@@ -61,6 +72,8 @@ impl HandlerMemory {
     }
   }
 
+  /// Grabs the relevant data for the event and constructs a new HandlerMemory with that value in
+  /// address 0, or returns no HandlerMemory if it is a void event.
   pub fn alloc_payload(
     event_id: i64,
     curr_addr: i64,
@@ -77,6 +90,7 @@ impl HandlerMemory {
     }
   }
 
+  /// Takes a given address and looks up the `mems` indexes relevant to it.
   pub fn addr_to_idxs(self: &HandlerMemory, addr: i64) -> (usize, usize) {
     return if addr >= 0 {
      self.addr.0[addr as usize]
@@ -87,6 +101,7 @@ impl HandlerMemory {
     }
   }
 
+  /// Reads fixed data from a given address.
   pub fn read_fixed(self: &HandlerMemory, addr: i64) -> i64 {
     let (a, b) = self.addr_to_idxs(addr);
     return if a == std::usize::MAX {
@@ -96,6 +111,7 @@ impl HandlerMemory {
     }
   }
 
+  /// Reads an array of data from the given address.
   pub fn read_fractal(self: &HandlerMemory, addr: i64) -> &[(usize, i64)] {
     let (a, b) = self.addr_to_idxs(addr);
     if addr_type(addr) == GMEM_ADDR {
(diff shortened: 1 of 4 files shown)


avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
 
-  pub fn push_fractal_fixed(self: &mut HandlerMemory, addr: i64, val: i64) {
-    // This implementation uses the `mem` vector as a way to keep track of the total length of the
-    // array, as well. It's simple but wastes space when the inserted value is variable-length
-    // (such as other strings or other arrays), however it greatly simplifies addressing and
-    // lookup, particularly for `Array<any>`, which is also what user-defined types are transformed
-    // into. In the future we could have an address translation layer and pack the data as tightly
-    // as we can, assuming that doesn't impose a large performance penalty, while this simple
-    // solution only adds an extra key's worth of space usage, but does have memory copy issues due
-    // to the constant resizing.
```
depombo, line 426: 😁

avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
-    } else if reg == FX_PTR_ADDR {
-      unsafe {
-        let ptr = usize::from_ne_bytes(self.mem[addr as usize].to_ne_bytes()) as *const i64;
-        let val = *ptr;
-        return val;
-      }
-    } else {
-      panic!("Trying to get fixed value from fractal pointer");
-    }
+  pub fn register_in(self: &mut HandlerMemory, orig_addr: i64, fractal_addr: i64, offset_addr: i64) {
```
depombo, line 224: nit - register_in_fractal?

avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
 
-  pub fn push_nested_fractal(self: &mut HandlerMemory, addr: i64, val: HandlerMemory) {
-    let arr = self.get_mut_fractal(addr);
-    let idx = arr.fractal_mem.len() as i64;
-    //println!("push nested: @{}[{}]: {}", addr, arr.mem.len(), val);
-    arr.mem.push(0);
-    arr.either_mem.push(idx);
-    arr.fractal_mem.push(val);
-    arr.registers_ish.push(0);
+  pub fn push_idxs(self: &mut HandlerMemory, addr: i64, a: usize, b: usize) {
```
depombo, line 205: nit push_addr_idxs ?

avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
-      let closure = &self.closure_args[0];
-      return closure.read_fractal_mem(addr - CLOSURE_ARG_MEM_START);
-    }
-    let reg = self.registers_ish[addr as usize];
-    if reg == NORMAL_ADDR {
-      let a = addr as usize;
-      let arr = &self.fractal_mem[self.either_mem[a] as usize];
-      let res = arr.mem.as_slice();
-      return res.to_vec();
+  pub fn register_out(self: &mut HandlerMemory, fractal_addr: i64, offset_addr: i64, out_addr: i64) {
```
depombo, line 230: nit - register_to_fractal or a quick one liner on the semantics of register_out?

avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
-      return closure.write_fractal(addr - CLOSURE_ARG_MEM_START, payload);
-    } else {
-      let idx = self.fractal_mem.len() as i64;
-      self.either_mem[addr as usize] = idx;
-      self.fractal_mem.push(payload);
-      self.registers_ish.push(0);
+  pub fn dupe(self: &mut HandlerMemory, orig_addr: i64, dest_addr: i64) {
+    // This *should be possible with something like this:
+    // HandlerMemory::transfer(self, orig_addr, self, dest_addr);
+    // But Rust's borrow checker doesn't like it, so we basically have to replicate the code here
```
depombo, line 331: ugh..

avm/src/vm/memory.rs
```diff
@@ -661,30 +397,23 @@ impl HandlerMemory {
...
-    let s_len = self.mem[0] as usize;
-    unsafe {
-      let s_u8 = slice::from_raw_parts(self.mem[1..].as_ptr().cast::<u8>(), s_len*8);
-      let s = str::from_utf8(&s_u8[0..s_len]).unwrap();
-      s.to_string()
+  pub fn fractal_to_string(f: &[(usize, i64)]) -> String {
+    let s_len = f[0].1 as usize;
+    let mut s_bytes: Vec<u8> = Vec::new();
+    for i in 1..f.len() {
+      let mut b = f[i].1.clone().to_ne_bytes().to_vec();
```
depombo, line 413: hmm might be missing something, but [`to_vec()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.to_vec) is effectively a clone. Can we get away without the additional `clone()` or does the borrow checker complain? If so, can we leave a comment about it

avm/src/vm/opcode.rs
```diff
@@ -1454,152 +1454,152 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-        // this could be a string or fixed data type
-        let val = arr.read_fixed(idx);
-        mem.write_fixed(CLOSURE_ARG_MEM_START + 1, val);
-      } else {
-        // more nested arrays
-        let arr_el = arr.read_fractal(idx);
-        mem.write_fractal(CLOSURE_ARG_MEM_START + 1, arr_el);
-      }
+      mem.set_addr(CLOSURE_ARG_MEM_START + 1, arr[idx].0, arr[idx].1 as usize);
+      mem.write_fixed(CLOSURE_ARG_MEM_START + 2, idx as i64);
```
depombo, line 1602: amazing!

avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
-  pub is_fixed: bool,
-  /// Memory space used for closure arguments
-  closure_args: Vec<HandlerMemory>, // Option doesn't work for some reason
+  /// The set of memory blocks. Each block consists of a tuple of two values. If the first value is
+  /// zero then the second value is the actual data, if it is non-zero then the two values together
+  /// represent a virtual pointer to another memory block and value (if the second value is zero,
+  /// it represents a pointer to a nested array of data, otherwise a pointer to an explicit value)
+  mems: Vec<Vec<(usize, i64)>>,
+  /// The address spaces for the handler memory. The first  is the "normal" memory space, and the
+  /// second is the args memory space.
```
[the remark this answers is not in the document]
dfellis: So, you don't have to explicitly declare a `HandlerMemory` is for a closure anymore, and I also use this space as a "scratch" space for certain opcodes, so that's why I don't mention "closure" in the comments.

avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
-  /// Flag indicating the memory stored is actually a pointer
-  registers_ish: Vec<i8>,
-  /// Temporary hack
-  pub is_fixed: bool,
-  /// Memory space used for closure arguments
-  closure_args: Vec<HandlerMemory>, // Option doesn't work for some reason
+  /// The set of memory blocks. Each block consists of a tuple of two values. If the first value is
+  /// zero then the second value is the actual data, if it is non-zero then the two values together
+  /// represent a virtual pointer to another memory block and value (if the second value is zero,
+  /// it represents a pointer to a nested array of data, otherwise a pointer to an explicit value)
```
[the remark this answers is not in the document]
dfellis: The comment is also out of date. The special value is now `usize::MAX` not zero. I will fix and try to incorporate something like what you've said here.

avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
+  mems: Vec<Vec<(usize, i64)>>,
+  /// The address spaces for the handler memory. The first  is the "normal" memory space, and the
+  /// second is the args memory space.
+  addr: (Vec<(usize, usize)>, Vec<(usize, usize)>),
+  /// Specifies which memory block to push "normal" values into
+  mem_addr: usize,
+  /// Specifies which memory block to push "args" into
```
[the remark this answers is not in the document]
dfellis: Same reply: you don't have to explicitly toggle one into a "closure" form, anymore. :)

avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
-  /// Temporary hack
-  pub is_fixed: bool,
-  /// Memory space used for closure arguments
-  closure_args: Vec<HandlerMemory>, // Option doesn't work for some reason
+  /// The set of memory blocks. Each block consists of a tuple of two values. If the first value is
+  /// zero then the second value is the actual data, if it is non-zero then the two values together
+  /// represent a virtual pointer to another memory block and value (if the second value is zero,
+  /// it represents a pointer to a nested array of data, otherwise a pointer to an explicit value)
+  mems: Vec<Vec<(usize, i64)>>,
+  /// The address spaces for the handler memory. The first  is the "normal" memory space, and the
```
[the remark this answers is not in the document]
dfellis: They're both virtualized, though?

avm/src/vm/memory.rs
```diff
@@ -1,654 +1,390 @@
...
+      hm
     }
-    formatter.write_str(&out_str)
   }
-}
 
-impl HandlerMemory {
-  /// Allocates a payload for the given event id from the address within the HandlerMemory
-  /// provided to a new HandlerMemory. Called by "emit to" opcode.
   pub fn alloc_payload(
```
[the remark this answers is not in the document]
dfellis: We had kinda made it overly complex by not reusing our `new` method, but I agree all of the different forms of memory was making this hard to comprehend.

avm/src/vm/memory.rs
dfellis: :tada:

avm/src/vm/memory.rs
dfellis: No, all of the `push_<thing>` functions are pushing into some nested memory. This function is for pushing the virtual pointer itself, which made certain operations faster *and* simpler. Unless you think they should *all be `push_addr_<thing>`?

avm/src/vm/memory.rs
dfellis: I think that would make it more confusing? Would `register_in_fractal` only work if the value inside is also a fractal?

I can add comments, though.

`register` - Just aliases one address for another.
`register_in` - Points to a value stored inside of a nested memory address
`register_out` Creates a pointer in the nested memory address to the outer memory address (inverse of register_in)

avm/src/vm/memory.rs
dfellis: I'll add comments for all three.

avm/src/vm/memory.rs
dfellis: I know. I don't even understand why it's a problem. :/ I hope it's fixed in a later version of Rust.

avm/src/vm/memory.rs
dfellis: So I did get problems here, but looking at this I think I can simplify it. `to_vec` was required to solve it because it has some sort of "copy-on-write"-like behavior where the basic `clone` doesn't work (because it's just a pointer and cloning a pointer still points at the same underlying data).

This is probably one of the functions I finished up around midnight and once it compiled and worked correctly I basically ignored since.

avm/src/vm/opcode.rs
dfellis: :smile:

avm/src/vm/memory.rs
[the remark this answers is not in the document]
depombo: hmm fair

avm/src/vm/memory.rs
depombo: got it, those comments make things a lot clearer

avm/src/vm/memory.rs
depombo: since it is a public function this particular method does seem confusing since idx could mean anything so a part of my does want to go with `push_addr_<thing>` for all, but more of a nit. fine leaving it as-is with a comment

depombo (approved):

PR #299, opened 2020-10-29
Add fork and join methods to HandlerMemory to make parallel map work, plus some other cleanup

dfellis:


avm/src/vm/memory.rs
```diff
@@ -426,6 +426,51 @@ impl HandlerMemory {
...
+  /// Joins two HandlerMemory structs back together. Assumes that the passed in handler memory was
+  /// generated by a fork call. This process consumes the forked HandlerMemory, moving over the
+  /// records created in the fork into the original and then "stitches up" the virtual memory
+  /// pointers for anything pointing at newly-created data. This mechanism is faster but will keep
+  /// around unreachable memory for longer. Whether or not this is the right trade-off will have to
+  /// be seen by real-world usage.
+  pub fn join(self: &mut HandlerMemory, mut hm: HandlerMemory) {
+    let s = hm.mem_addr; // The initial block that will be transferred (plus all following blocks)
+    let s2 = self.mems.len(); // The new address of the initial block
+    let offset = s2 - s; // Assuming it was made by `fork` this should be positive or zero
```
depombo, line 451: should we maybe panic if offset < 0?

avm/src/vm/opcode.rs
```diff
@@ -1617,7 +1612,7 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
     }).collect();
     hand_mem.write_fractal(args[2], &Vec::new());
     for mem in output {
-      HandlerMemory::transfer(&mem, CLOSURE_ARG_MEM_START, hand_mem, CLOSURE_ARG_MEM_START);
```
depombo, line 1620: do we still need `transfer`?

dfellis: It's used by the HTTP and Datastore opcodes, and in those cases we really do need a copy and we really do want just the relevant data instead of wholesale swiping of memory blocks, so we do still need it.

avm/src/vm/memory.rs
dfellis: The type is `usize` so I thought Rust would do that automatically?

dfellis: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=fn%20main()%20%7B%0A%20%20%20%20let%20avec%20%3D%20vec!%5B1%2C%202%2C%203%5D%3B%0A%20%20%20%20let%20bvec%20%3D%20vec!%5B1%2C%202%2C%203%2C%204%5D%3B%0A%20%20%20%20let%20a%20%3D%20avec.len()%3B%0A%20%20%20%20let%20b%20%3D%20bvec.len()%3B%0A%20%20%20%20let%20c%20%3D%20a%20-%20b%3B%0A%20%20%20%20print!(%22%7B%7D%22%2C%20c)%3B%0A%7D

depombo (approved):

Issue #197, opened 2020-08-19
ammtoaga doesn't define dependencies correctly


dfellis:
This has been a lower priority since the runtime mostly sticks to the user-defined order whenever possible, but complex IO operations may be reordered incorrectly without this being fixed.

PR #302, opened 2020-11-03
Move to using Tokio only, and get nested closures fully working in the AVM

dfellis:
Almost done with this. Resolves \#224

However, because of \#197 I had to roll back the IO parallelism work for now. Hope to restore that soon!


avm/src/vm/memory.rs
```diff
@@ -72,6 +72,15 @@ impl HandlerMemory {
     }
   }
 
+  /// Because of constraints in Tokio's RWLock, this method moves the contents of `self` into the
+  /// `target`. The `self` is destroyed in the process
+  pub fn replace(self: HandlerMemory, target: &mut HandlerMemory) {
```
depombo, line 77: hmmm, not sure how I feel about this. my intuition tells me maybe there is a better way to do this?

avm/src/vm/event.rs
```diff
@@ -188,29 +183,61 @@ impl HandlerFragment {
...
-    // Next insert the new handler where we want it to start (at zero)
-    self.handlers.insert(0, HandlerFragmentID {
-      event_id,
-      handler_idx: 0,
-      // Finally, because of how `get_next_fragment` works, it will assume the first fragment
-      // in the new subhandler has already been run, which is not true, so we start it out at None
-      // and from there increment it to 0.
-      fragment_idx: None,
-    });
+  pub async fn run(mut self: HandlerFragment, mut hand_mem: HandlerMemory) -> HandlerMemory {
```
depombo, line 186: this method needs a comment about how concurrency via the tokio api is working for us. Mentioning how spin up a new task for every fragment and then we block in place for cpu-bound fragment or spin another task if io bound and why

depombo (commented):
amazing work and re-architecture for the AVM!

avm/src/vm/event.rs
```diff
@@ -72,6 +80,8 @@ impl EventHandler {
         self.fragments.push(frag);
       }
     } else {
+      // TODO: Restore this logic. For now just turn it into a new fragment by itself
```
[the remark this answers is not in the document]
dfellis: Yes, I was planning to. :)

avm/src/vm/memory.rs
dfellis: I tried updating tokio which appeared to have the function I was looking for, but then it failed to compile because the latest Tokio depends on a super-recent version of Rust, so I went back and wrote this. At only 6 LOC I'm not too bothered, and the perf impact is minimal.

avm/src/vm/event.rs
dfellis: Absolutely. :)

avm/src/vm/memory.rs
dfellis: (A super recent version of Rust that we aren't targeting, I mean.)

dfellis: But on that front, I am thinking about potentially changing the `io!` opcode definition as the current form with the Box-wrapped async function seems heavyweight and I wonder if we can do it better (I don't see a single `io!` opcode not following that pattern, for instance).

depombo (approved):

PR #304, opened 2020-11-04
Get the Tree data structure working on the AVM

dfellis:


avm/src/vm/memory.rs
```diff
@@ -251,38 +251,15 @@ impl HandlerMemory {
   }
 
   /// Creates an alias for data at one address in another address.
-  pub fn register(self: &mut HandlerMemory, addr: i64, orig_addr: i64) {
+  pub fn register(self: &mut HandlerMemory, addr: i64, orig_addr: i64, is_variable: bool) {
```
depombo, line 254: much better!

depombo (approved):

PR #306, opened 2020-11-10
Implement @std/seq's recurse functionality in the AVM

dfellis:
Resolves \#265

### DIFF
--- a/avm/src/vm/memory.rs
+++ b/avm/src/vm/memory.rs
@@ -5,7 +5,7 @@ use crate::vm::program::Program;
 
 // -2^63
 pub const CLOSURE_ARG_MEM_START: i64 = -9223372036854775808;
-pub const CLOSURE_ARG_MEM_END: i64 = CLOSURE_ARG_MEM_START + 3;
+pub const CLOSURE_ARG_MEM_END: i64 = CLOSURE_ARG_MEM_START + 9001; // TODO: IT'S OVER 9000!
 // Flags for the registers_ish vector. The normal address flag indicates that the data is stored
 // normally in either the memory or fractal memory structures. The fixed pointer address flag
 // indicates that the value in the memory structure is actually a pointer to an i64 value. The
--- a/avm/src/vm/memory.rs
+++ b/avm/src/vm/memory.rs
@@ -5,6 +5,10 @@ use crate::vm::program::Program;
 
 // -2^63
 pub const CLOSURE_ARG_MEM_START: i64 = -9223372036854775808;
+// The closure arg memory end has been extended to handle disambiguating nested closure arguments
+// being used deep in the scope hierarchy. The quickest solution was to just increase that memory
+// space to a large constant range, but the proper solution is to make this no longer a constant
+// and determine the range based on the side of the global memory.
 pub const CLOSURE_ARG_MEM_END: i64 = CLOSURE_ARG_MEM_START + 9001; // TODO: IT'S OVER 9000!
 // Flags for the registers_ish vector. The normal address flag indicates that the data is stored
 // normally in either the memory or fractal memory structures. The fixed pointer address flag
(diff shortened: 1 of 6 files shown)


avm/src/vm/memory.rs
```diff
@@ -5,7 +5,7 @@ use crate::vm::program::Program;
 
 // -2^63
 pub const CLOSURE_ARG_MEM_START: i64 = -9223372036854775808;
-pub const CLOSURE_ARG_MEM_END: i64 = CLOSURE_ARG_MEM_START + 3;
+pub const CLOSURE_ARG_MEM_END: i64 = CLOSURE_ARG_MEM_START + 9001; // TODO: IT'S OVER 9000!
```
depombo, line 8: can we add some more context about why this is 9001 right now?

avm/src/vm/opcode.rs
```diff
@@ -2468,7 +2468,37 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      let mut hand_mem = mem.write().await;
+      let mut hm = hand_mem.fork();
+      let (a, b) = hm.addr_to_idxs(args[0]);
+      hm.set_addr(CLOSURE_ARG_MEM_START + 1, a, b);
+      let (a, b) = hm.addr_to_idxs(args[1]);
+      hm.set_addr(CLOSURE_ARG_MEM_START + 2, a, b);
+      let slf = hm.read_fractal(args[0]);
+      let (mut seq, _) = hm.read_either_idxs(slf[0].0, slf[0].1 as usize);
+      let recurse_fn = HandlerFragment::new(slf[1].1, 0);
+      if seq[0].1 < seq[1].1 {
```
depombo, line 2482: this is not really part of this PR, but I'm not a big fan of tuples just because they hurt legibility. also what does `slf` stand for?

compiler/src/ammtoaga.ts
```diff
@@ -135,9 +135,28 @@ const getHandlersMem = (handlers: LPNode[]) => handlers
     return handlerMem
   })
 
-const closuresFromDeclaration = (declaration: LPNode, closureMem: object, eventDecs: object) => {
+const closuresFromDeclaration = (
+  declaration: LPNode,
+  closureMem: object,
+  eventDecs: object,
+  addressMap: object,
+  argRerefOffset: number,
```
depombo, line 143: can we leave a short comment about how argRerefOffset works?

depombo (approved):
Looks great, just minor comments!

avm/src/vm/memory.rs
dfellis: Because it's over 9000!?

What!? 9000!?

compiler/src/ammtoaga.ts
dfellis: Sure

avm/src/vm/opcode.rs
dfellis: So, the tuples are the core of the memory model now. Rather than add some more memory model function calls, I just dug into it in just this one opcode as a trade-off.

And `slf` is `self` but the Rust compiler wouldn't let me call it that because the `self` variable is a reserved keyword only allowed inside of `impl` blocks for some insane reason.

avm/src/vm/memory.rs
dfellis: I can add more, I just couldn't fit it after the constant so it'll have to be a new line.

avm/src/vm/opcode.rs
depombo: > So, the tuples are the core of the memory model now. Rather than add some more memory model function calls, I just dug into it in just this one opcode as a trade-off.
> 
> And `slf` is `self` but the Rust compiler wouldn't let me call it that because the `self` variable is a reserved keyword only allowed inside of `impl` blocks for some insane reason.

ahh got it!

depombo (approved):

depombo:
Pending tests passing

PR #318, opened 2020-11-13
add telemetry

depombo:


avm/src/vm/telemetry.rs
```diff
@@ -0,0 +1,26 @@
...
+
+const AMPLITUDE_API_KEY: &str = "ae20dafe801eddecf308c6ce643e19d1";
+const AMPLITUDE_URL: &str = "https://api.amplitude.com/2/httpapi";
+// https://doc.rust-lang.org/cargo/reference/environment-variables.html
+const ALAN_VERSION: Option<&'static str> = option_env!("CARGO_PKG_VERSION");
+const NO_TELEMETRY: Option<&'static str> = option_env!("ALAN_TELEMETRY_OFF");
+const OS: &str = std::env::consts::OS;
+
+pub async fn log() {
+  let no_telemetry = json!(NO_TELEMETRY.unwrap_or("false")).as_bool().unwrap_or(false);
```
dfellis, line 11: Why not just `let no_telemetry = NO_TELEMETRY.unwrap_or("false") == "true";`

avm/src/vm/run.rs
```diff
@@ -99,6 +100,7 @@ pub fn exec(fp: &str, delete_after_load: bool) {
     let mut vm = VM::new();
     let start = EventEmit { id: i64::from(BuiltInEvents::START), payload: None };
     vm.add(start);
+    telemetry::log().await;
```
dfellis, line 103: Doesn't this only log on application start, not all of the various commands?

dfellis:
General thoughts: maybe instead of putting the hook in the run command itself, put it in `avm/src/main.rs` and we can easily hook onto all of the different subcommands, including compilation and dependency management?

avm/src/vm/telemetry.rs
```diff
@@ -0,0 +1,26 @@
+use serde_json::json;
+
+const AMPLITUDE_API_KEY: &str = "ae20dafe801eddecf308c6ce643e19d1";
```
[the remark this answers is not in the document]
depombo: public key!

avm/src/vm/run.rs
depombo: Yes :/

depombo:
> General thoughts: maybe instead of putting the hook in the run command itself, put it in `avm/src/main.rs` and we can easily hook onto all of the different subcommands, including compilation and dependency management?

Unfortunately, I could not hook up to subcommands from `avm/src/main.rs` because there is no async / tokio environment defined at that point so it would only be possible with a blocking call. I want to start out with this which is already a great start.

avm/src/vm/run.rs
depombo: https://github.com/alantech/alan/pull/318#issuecomment-727245916

dfellis (approved):

PR #320, opened 2020-11-14
implement min and max built-in functions

MareinK:


dfellis (approved):
This is great :) Can we get `string`, too?

MareinK:
Done!

dfellis (approved):

PR #322, opened 2020-11-19
Refactor ammtoaga to make dependency graph determination more correct

dfellis:
Resolves \#197

Well, sorta. We need to make the dependency resolution recursive for dependencies very far removed
from the original definition in terms of blocks, but it should be very hard to cause such a problem
*and* we can build on this base to do it in a single function (instead of being spread all over the
place) later, while unblocking restoration of graph-based concurrency in the AVM.


depombo (approved):
LGTM pending test suite passing! 👍

PR #324, opened 2020-11-20
do not track avm runs on test runs and use new metric name to get clean count and stay under event limit

depombo:


dfellis (approved):

bdd/bdd.sh
```diff
@@ -27,6 +27,11 @@ PHYS_DIR=`pwd -P`
 
 PATH="${PHYS_DIR}:${PHYS_DIR}/../shellspec:${PHYS_DIR}/../node_modules/.bin:${PHYS_DIR}/../avm/target/release:${PATH}"
 export PATH
+
+# turn off telemetry
+ALAN_TELEMETRY_OFF="true"
+export ALAN_TELEMETRY_OFF
+
```
depombo, line 34: this works locally, but not in Github's CI

PR #326, opened 2020-11-20
Add Orderable interface, use to simplify min/max implementation.

dfellis:
I was not able to reproduce the issue pointed out by @MareinK. Will need to ask what the difference
is with this and what was tried before.

### DIFF
--- a/bdd/spec/011_array_spec.sh
+++ b/bdd/spec/011_array_spec.sh
@@ -383,8 +383,8 @@ true"
           const test2 = [ 4, 5, 6 ]
           print('reduce test')
           test.reduce(fn (a: int, b: int): int = a + b).print()
-          test.reduce(fn (a: int, b: int): int = min(a, b)).print()
-          test.reduce(fn (a: int, b: int): int = max(a, b)).print()
+          test.reduce(min).print()
+          test.reduce(max).print()
 
           print('filter test')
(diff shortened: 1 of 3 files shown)


depombo (approved):

dfellis:
Hmm... it does fail, but only when it's being used inside of a closure passed to an array method, not when it's used directly inside of a handler function.

MareinK:
For me, these work on 0.1.19:

```alan
app.print(reduce([1, 2, 3], min))
app.print(reduce([1, 2, 3], add))
```

But these fail:

```alan
app.print([1, 2, 3].reduce(min))
app.print([1, 2, 3].reduce(add))
```

With these errors:

```
Cannot resolve operators with remaining statement
[1, 2, 3].reduce(min)
<Array<int64>> <int64>
```

```
Cannot resolve operators with remaining statement
[1, 2, 3].reduce(add)
<Array<int64>> <int64>
```

dfellis:
Cannot resolve *operators*? What operator is it even referring to.

I will try to reproduce and see how it is getting confused like this, but probably not today. Housework on the weekends, usually.

MareinK:
I found an extra hint: the below do work.

```alan
app.print(([1, 2, 3]).reduce(min))
app.print(([1, 2, 3]).reduce(add))
```

PR #328, opened 2020-11-22
add toInt and toFloat type coercion aliases

MareinK:
I'm not sure if this is the best way to do it. I couldn't find a way to simply do `toFloat = toFloat64`. Let me know if there is an easier way! I didn't add tests because there were no tests for the existing coercion functions.

### DIFF
--- a/bdd/spec/007_comparator_spec.sh
+++ b/bdd/spec/007_comparator_spec.sh
@@ -421,5 +421,39 @@ false"
       The output should eq "$GREATERTHANOREQUAL"
     End
   End
+
+  Describe "Type Coercion Aliases"
+    before() {
+      sourceToAll "
+        from @std/app import start, print, exit
+
+        on start {
+          print(toInt(0) == toInt64(0));
+          print(toFloat(0.0) == toFloat(0.0));
+
+          emit exit 0;
+        }
+      "
+    }
+    BeforeAll before
+
+    after() {
+      cleanTemp
+    }
+    AfterAll after
+
+    TYPECOERCIONALIASES="true
+true"
+
+    It "runs js"
+      When run test_js
+      The output should eq "$TYPECOERCIONALIASES"
+    End
+
+    It "runs agc"
+      When run test_agc
+      The output should eq "$TYPECOERCIONALIASES"
+    End
+  End
 End
 
--- a/std/root.ln
+++ b/std/root.ln
@@ -39,6 +39,12 @@ export interface Orderable {
   gt(Orderable, Orderable): bool,
   gte(Orderable, Orderable): bool,
 }
+export interface canFloat64 {
+  toFloat64(canFloat64): float64
+}
+export interface canInt64 {
+  toInt64(canInt64): int64
+}
 
 // Type conversion functions
 export fn toFloat64(n: int8) = i8f64(n);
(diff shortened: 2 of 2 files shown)


dfellis:
So being able to alias functions wasn't ever expected to be a common solution, so what you have done is right.

It could be shortened a little bit by something like:

```
interface canFloat64 {
  toFloat64(canFloat64): float64
}
export toFloat(n: canFloat64): float = toFloat64(n) 
```

But I can see why you might be shy about using interfaces. ;)

MareinK:
I was actually thinking whether something like this might be possible, but I wasn't sure how to do it! Changed the implementation to be like that :)

MareinK:
Hmm, this is building without error but tests are failing. The first test to fail is at `bdd/spec/003_math_spec.sh:12`, which tests `int8` addition, with the error "Undefined function called: add". The `add` aliases are defined directly after the `toFloat` and `toInt` aliases.

Also, the 'browser check' test is failing with "Error: Operator || is not defined" at `bundle.js:12029:27`.

Not sure how I can proceed from here, I'll leave it for someone else to have a look sometime :)

dfellis:
> Hmm, this is building without error but tests are failing. The first test to fail is at `bdd/spec/003_math_spec.sh:12`, which tests `int8` addition, with the error "Undefined function called: add". The `add` aliases are defined directly after the `toFloat` and `toInt` aliases.
> 
> Also, the 'browser check' test is failing with "Error: Operator || is not defined" at `bundle.js:12029:27`.
> 
> Not sure how I can proceed from here, I'll leave it for someone else to have a look sometime :)

So the browser check has recently become very flaky for reasons that I haven't had time to look into, yet. I presume something to do with changes to cypress or one of its dependencies. (I would like to drop cypress as it feels like overkill but I haven't found a good alternative, yet.)

I will dig in on what's going on with the rest of the test suite. The root scope is very special in the compiler, so there might be some interaction with the compiler itself that's causing an issue here.

dfellis:
I think I got it working.

1. Forgot to export the interfaces.
2. Forgot to label the new functions as `fn` and it hit the same error you pointed out earlier about the compiler not properly erroring out on exporting an assignment (which is what it thought it was) which then cascaded to the rest of the file being invalid.

dfellis:
Screenshot from 2020-11-23 11-42-23

These are the exact changes I made if you want to make them and get this PR merged @MareinK :)

If not, I can make a new PR tomorrow based on your's. (But would prefer you get the credit.)

MareinK:
I'll be happy to when I get time in the coming days! I'm also fine with someone else taking it up.

MareinK:
Changes applied, seems good now except for the browser check!

dfellis:
I triggered the test suite to re-run last night and it looks good. There's something flaky about it that we need to dig into (I suspect some node dep change because this is relatively recent behavior without any changes to that code).

dfellis:
@MareinK can we get a test added to the BDD tests, perhaps [this test group](https://github.com/alantech/alan/blob/main/bdd/spec/007_comparator_spec.sh) to exercise `toInt` and `toFloat`?

MareinK:
I added a very simple test, I couldn't think of a counterexample to test. Let me know if you can think of more to add!

There is a seemingly unrelated test failing on macos, at `011_array_spec.sh:447` which I can't explain.

dfellis (approved):
I am still approving this despite the CI failure. We have seen that on another branch and weren't sure if it was a latent race condition or if it was caused by the changes in that PR.

Your change has nothing to do with it, so that confirms the latent race condition theory.

dfellis:
The bug revealed by the OSX CI has been squashed, so I'm going to merge this now.

dfellis:
Thank you for your patience on this!

MareinK:
No problem at all, thanks for your help! :) Nice to see it merged.

Issue #330, opened 2020-11-23
Parser bug on arrays  being used in method call


dfellis:
https://github.com/alantech/alan/pull/326#issuecomment-731750937

Thanks again to @MareinK for reporting this!

These work:

```ln
app.print(reduce([1, 2, 3], add))
app.print(([1, 2, 3]).reduce(add))
```

and this does not:

```ln
app.print([1, 2, 3].reduce(add))
```

and gives a nonsensical error message about operators.

This is probably a bug in the ANTLR AST definition, but need to investigate.

dfellis:
To be clear, I think this is another aspect of #98 making it a blend of both AST generation and later parts of the first stage of the compiler.

dfellis:
Confirmed, btw, right here's the definition: https://github.com/alantech/alan/blob/main/compiler/src/ln/Ln.g4#L77

And it only allows constants (true, false, numbers, strings) or paren-wrapped values to be used in the method syntax right now. So I'm going to be redoing all of this anyways, and will take care of that at the same time.

PR #332, opened 2020-11-25
Major first-stage refactor. Faster compilation, method syntax works predictably, and semicolons!

dfellis:
Resolves \#98 and \#330

Currently only reproduces the current behavior with new ANTLR grammar, but the new grammar should allow new tests to confirm that it's fixed.


dfellis:
This is probably going to be the main thing I work on this week.

Also the current draft of the ANTLR grammar looks correct to me, but it's also much slower than the existing grammar, so I may completely rework this again before it's landed -- I would like to *reduce* the amount of backtracking, not increase it, in the parser.

dfellis:
Creating this comment to keep track of which test suites are working and which still aren't. Will just update this comment as they're resolved now.

- [x] 001_event_spec.sh
- [x] 002_printing.sh
- [ ] 003_math_spec.sh
- [x] 004_bitwise_spec.sh
- [x] 005_boolean_spec.sh
- [ ] 006_string_spec.sh
- [x] 007_comparator_spec.sh
- [x] 008_function_spec.sh
- [ ] 009_conditional_spec.sh
- [ ] 010_object_literal_spec.sh
- [ ] 011_array_spec.sh
- [ ] 012_hash_spec.sh
- [x] 013_generics_spec.sh
- [ ] 014_interface_spec.sh
- [ ] 015_maybe_result_either_spec.sh
- [x] 016_type_spec.sh
- [x] 017_custom_event_spec.sh
- [ ] 018_closure_spec.sh
- [x] 019_compiler_error_spec.sh
- [x] 020_module_constants_spec.sh
- [ ] 021_trig_spec.sh
- [x] 022_deps_spec.sh
- [ ] 023_http_spec.sh
- [ ] 024_clone_spec.sh
- [x] 025_runtime_error_spec.sh
- [ ] 026_datastore_spec.sh (currently hangs)
- [ ] 027_seq_spec.sh
- [ ] 028_tree_spec.sh
- [x] 029_err_printing_spec.sh
- [x] 030_cmd_spec.sh

dfellis:
It's all passing locally, now!

Next steps:

- [ ] Add test cases for syntax this work is meant to unlock
- [ ] Make function return type determination lazy so we actually get the perf gain this is meant to provide
- [ ] Rework the website and docs to scatter semicolons across them :/

dfellis:
Soo... I can't reproduce this error locally, but it appears to be consistent across the three environments. Trying to figure out why.

std/root.ln
```diff
@@ -660,212 +660,217 @@ export fn toHashMap(kva: Array<KeyVal<Hashable, any>>) {
...
+export fn addChild(n: Node<any>, val: Tree<any>): Node<any> = n.addChild(val.getRootNode());
 
-export fn getOr(n: Node<any>, default: any): any = getOr(n.tree.vals[n.id], default)
+export fn getOr(n: Node<any>, default: any): any = getOr(n.tree.vals[n.id], default);
 
-export fn toNodeArray(t: Tree<any>): Array<Node<any>> = map(t.vals, fn (val: any, i: int64): Node<any> = t.getNodeById(i))
+export fn toNodeArray(t: Tree<any>): Array<Node<any>> = map(
+  t.vals,
+  fn (val: any, i: int64): Node<any> = t.getNodeById(i)
+);
```
dfellis, line 844: Take a look at this. It seems trivial, but this kind of thing wasn't parsing correctly before!

dfellis:
> Soo... I can't reproduce this error locally, but it appears to be consistent across the three environments. Trying to figure out why.

Pulled out an old laptop and rebuilt and tested on there and also can't reproduce the error here on that branch.

I'm going to temporarily turn off the xargs-based parallel execution and see if that helps, because I'm at a loss.

dfellis:
This error is maddeningly consistent and nonsensical! The line numbers its pointing to don't line up with any assertions in the shellspec file!

dfellis:
Discovered an issue with the new type inference code in this branch with code that uses an outer scope variable failing.

This *must* be resolved before this PR is landed. Putting this back into draft mode.

std/trig.ln
```diff
@@ -1,242 +1,242 @@
...
+export fn secant(x: float32) = toFloat32(secant(toFloat64(x)));
+
+export fn csc(x: float64) = 1.0 / sinf64(x);
+export fn csc(x: float32) = toFloat32(csc(toFloat64(x)));
+export fn cosecant(x: float64) = 1.0 / sinf64(x);
+export fn cosecant(x: float32) = toFloat32(cosecant(toFloat64(x)));
+
+export fn cot(x: float64) = 1.0 / tanf64(x);
+export fn cot(x: float32) = toFloat32(cot(toFloat64(x)));
+export fn cotangent(x: float64) = 1.0 / tanf64(x);
```
depombo, line 41: 👀

compiler/src/lntoamm/Statement.ts
```diff
@@ -88,64 +91,15 @@ class Statement {
...
   static isAssignablePure(assignableAst: any, scope: Scope) { // TODO: Migrate off ANTLR
-    if (assignableAst.basicassignables() != null) {
-      return Statement.isBasicAssignablePure(assignableAst.basicassignables(), scope)
-    }
-    if (assignableAst.withoperators() != null) {
-      return Statement.isWithOperatorsPure(assignableAst.withoperators(), scope)
-    }
-    // This should never be reached
-    throw new Error("Impossible assignment situation")
+    // TODO: Redo this
```
depombo, line 95: nit / not blocking this PR: what is "this" and what is a potential Redo? I actually have hit this error a few times

compiler/src/lntoamm/Microstatement.ts
```diff
@@ -1728,23 +1192,486 @@ ${constdeclarationAst.getText()} on line ${constdeclarationAst.start.line}:${con
...
   }
 
+  static fromBaseAssignableAst(
+    baseAssignableAsts: any, // TODO: Eliminate ANTLR
+    scope: Scope,
+    microstatements: Array<Microstatement>,
+  ) {
+    // The base assignables array are a lightly annotated set of primitives that can be combined
+    // together to produce an assignable value. Certain combinations of these primitives are invalid
+    // and TODO provide good error messaging when these are encountered. A state machine of valid
```
depombo, line 1202: this is awesome!!!

depombo (approved):

compiler/src/lntoamm/Statement.ts
dfellis: I deleted everything and just marked all assignables pure all the time. So it needs to be completely redone and I thought it was self-evident that `return true` is not something you'd ever actually want from a function call.

We aren't yet using function/statement purity in the compiler, so it's not a huge deal, but if/when we want the compiler to automatically downgrade a `map` to a `mapLin` because you're mutating outer state, this will need to be working and correct.

I can update the TODO comment with all of this if you want?

dfellis: Since it's not blocking, though, I can do it in a followup PR.

PR #336, opened 2020-12-01
revive io parallelization

depombo:


dfellis (approved):
Assuming tests pass, this is great news! :D

PR #344, opened 2020-12-07
getOrExit function

dfellis:
After see @ledwards' advent of code submissions, realized that this option is missing for CLI-type applications that really do want to halt on error instead of continuing. Providing a default value that is completely nonsensical will just produce nonsensical outputs.

### DIFF
--- a/bdd/spec/025_runtime_error_spec.sh
+++ b/bdd/spec/025_runtime_error_spec.sh
@@ -9,4 +9,43 @@ Describe "Runtime Errors"
       The error should include "File not found:"
     End
   End
+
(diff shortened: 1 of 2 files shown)


depombo:
I know I mentioned 'getOrHalt', but I just thought about 'getOrExit' which might be more descriptive and bodes better with the fact that we have an 'exit' event

dfellis:
> I know I mentioned 'getOrHalt', but I just thought about 'getOrExit' which might be more descriptive and bodes better with the fact that we have an 'exit' event

SGTM

depombo (approved):

PR #345, opened 2020-12-08
Fix the findl opcode when mutating outer scope

dfellis:
Thanks to @cdmistman for spotting this!

### DIFF
--- a/avm/src/vm/opcode.rs
+++ b/avm/src/vm/opcode.rs
@@ -1724,7 +1724,7 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
         let mut hm = hand_mem.clone();
         hm.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
         hm = subhandler.clone().run(hm).await;
-        // The sequential version of `each` is allowed to have side-effects
+        // The sequential version of `find` is allowed to have side-effects
         hm.replace(&mut hand_mem);
         let val = hand_mem.read_fixed(CLOSURE_ARG_MEM_START);
         if val == 1 {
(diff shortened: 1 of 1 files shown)


avm/src/vm/opcode.rs
```diff
@@ -1717,14 +1717,16 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      let arr = hand_mem.read_fractal(args[0]).to_vec();
       let len = arr.len();
       let subhandler = HandlerFragment::new(args[1], 0);
       for i in 0..len {
-        let mut hm = hand_mem.fork();
+        let mut hm = hand_mem.clone();
         hm.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
         hm = subhandler.clone().run(hm).await;
-        let val = hm.read_fixed(CLOSURE_ARG_MEM_START);
+        // The sequential version of `each` is allowed to have side-effects
```
depombo, line 1727: `find`?

this also opens the non-blocking question of whether `map` should be allowed to have side-effects?

depombo (approved):

avm/src/vm/opcode.rs
dfellis: Oh, yes, let me fix that.

And `find` and `map` can't do that because they run in parallel and mutating outer scope would be nondeterministic.

That's what the `pure` flag on functions is about -- to try to track that and auto-downgrade to these linear opcodes when we can't safely run code in parallel -- but that's also not working yet, hence the `eachLin`, `findLin`, etc being directly exposed (though I'm also not opposed to keeping those exposed long-term for other reasons (like predictable execution ordering when communicating to some other server, for instance).

depombo (approved):

Issue #343, opened 2020-12-07
Multiple problems with the AVM and multiply-nested array method calls


dfellis:
The AVM currently eager schedules parallel array operations, which creates a fork of the handler memory space so they can safely run in parallel.
With a large enough number of arrays this can OOM the AVM even though only up to `n` (cpu core count) instances can run at a time. Slowing this down and tossing up batches of elements to process at a time should help here.

Relatedly, explicitly using a linear version of these array methods runs slower than the JS transpiled version of the Alan code. This has to be an issue with how the Tokio tasks are scheduled in the AVM right now and causing a lot of unnecessary overhead. Reusing the existing Tokio thread for the iterations in the linear form should be used, instead.

Hat tip to @ledwards for pointing this out.

depombo:
> The AVM currently eager schedules parallel array operations, which creates a fork of the handler memory space so they can safely run in parallel.
> With a large enough number of arrays this can OOM the AVM even though only up to `n` (cpu core count) instances can run at a time. Slowing this down and tossing up batches of elements to process at a time should help here.
WIll be fixed by https://github.com/alantech/alan/issues/354

> Relatedly, explicitly using a linear version of these array methods runs slower than the JS transpiled version of the Alan code. This has to be an issue with how the Tokio tasks are scheduled in the AVM right now and causing a lot of unnecessary overhead. Reusing the existing Tokio thread for the iterations in the linear form should be used, instead.
> 
Fixed by https://github.com/alantech/alan/pull/346

depombo:
https://github.com/alantech/alan/pull/384 and https://github.com/alantech/alan/pull/374 allow multiply nested array methods to work and have decent performance which is why I closed this out. It is worth pointing out tossing up batches of elements, which is called out in the description is still not done, but is not necessary to fix the bug and will be tracked in https://github.com/alantech/alan/issues/354

PR #346, opened 2020-12-09
AVM Perf Cleanup Part 1

dfellis:
Example code from @ledwards demonstrated a couple of pathological cases in the AVM:

1. When dropping a value that is an array of nested arrays, the nested arrays stay around until the handler completes in a semi-detached state, so code that causes that in a tight loop can quickly OOM the AVM.
2. Many opcodes are needlessly `clone`-ing the handler memory data structure for convenience, putting `memcpy`s in a tight loop; coupled with (1) that means each iteration of the loop takes longer than the last.

On my machine, this makes an explicitly linear version of the example code to go from >2 hours runtime (I just killed it after it had been running that long) to ~1 minute. However, transpiling to Javascript and running it takes ~4 seconds, so there's still a ways to go.

Breaking this up into multiple PRs to keep it reviewable. Future work includes:

* Array methods that use a user-provided function to automatically detect if the function is 100% CPU bound and perform the loop in-thread instead of scheduling in Tokio.
* Try to "uninfect" such io opcodes to avoid unnecessary threading when possible
* Actual parallel opcodes (eg `map` vs `mapl`) to only schedule up to `n` threads where `n` is the number of CPU cores, rather than creating a Tokio task per array element, and each Tokio thread consuming `m/n` elements where `m` is the total number of elements in the array being operated on.
* Rethink the AVM memory model itself to make deletions as cheap as insertions

It's expected that just the first one should get us within margin-of-error of the JS transpiled form, while the second one *should* get us slightly faster than JS, the third makes the AVM demonstrably better than normal dynamic languages and competitive with JVM, and the fourth to avoid a whole class of issues that probably exist but we don't have example code for, yet.

This PR only tackles half of what's described in #343

### DIFF
--- a/avm/src/vm/event.rs
+++ b/avm/src/vm/event.rs
@@ -6,7 +6,7 @@ use tokio::task;
 
 use crate::vm::instruction::Instruction;
 use crate::vm::memory::HandlerMemory;
-use crate::vm::opcode::EmptyFuture;
+use crate::vm::opcode::HMFuture;
 use crate::vm::program::Program;
 use crate::vm::run::{EVENT_TX};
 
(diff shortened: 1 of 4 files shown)


avm/src/vm/telemetry.rs
```diff
@@ -22,5 +22,7 @@ pub async fn log() {
     ]
   });
   let client = reqwest::Client::new();
-  client.post(AMPLITUDE_URL).json(&body).send().await;
+  if client.post(AMPLITUDE_URL).json(&body).send().await.is_ok() {
+    // Do nothing
+  }
```
dfellis, line 27: This is to eliminate a warning from the rust compiler so we get back to zero warning on compilation.

avm/src/vm/memory.rs
```diff
@@ -207,9 +207,33 @@ impl HandlerMemory {
...
+          drop(old_fractal);
+          self.mems.pop();
+          break;
+        }
+      }
+      self.mems[a] = val.to_vec().clone();
+    } else {
+      let a = self.mems.len();
+      self.mems.push(val.to_vec().clone());
+      self.set_addr(addr, a, std::usize::MAX);
```
depombo, line 235: I'm a big fan of keeping this contained to one method!

avm/src/vm/opcode.rs
```diff
@@ -46,12 +45,12 @@ static DS: Lazy<Arc<DashMap<String, HandlerMemory>>> = Lazy::new(|| {
 /// For more information see:
 /// https://stackoverflow.com/questions/58354633/cannot-use-impl-future-to-store-async-function-in-a-vector
 /// https://stackoverflow.com/questions/51485410/unable-to-tokiorun-a-boxed-future-because-the-trait-bound-send-is-not-satisfie
-pub type EmptyFuture = Pin<Box<dyn Future<Output = ()> + Send>>;
+pub type HMFuture = Pin<Box<dyn Future<Output = HandlerMemory> + Send>>;
```
depombo, line 48: 😎

avm/src/vm/memory.rs
```diff
@@ -477,15 +501,20 @@ impl HandlerMemory {
...
+      hm.mems.drain(..s); // Remove the irrelevant memory blocks
+      self.mems.append(&mut hm.mems); // Append the relevant ones to the original HandlerMemory
+      // Set the return address on the original HandlerMemory to the acquired indexes, potentially
+      // offset if it is a pointer at new data
+      if a < std::usize::MAX && a >= s {
+        self.set_addr(CLOSURE_ARG_MEM_START, a + offset, b);
+      } else {
+        self.set_addr(CLOSURE_ARG_MEM_START, a, b);
+      }
     } else {
```
depombo, line 515: when exactly does this branch run?

depombo (approved):

avm/src/vm/memory.rs
dfellis: Especially when this special casing should be removed in the future and shouldn't affect the public API

avm/src/vm/memory.rs
dfellis: For any IO opcode that *doesn't* provide a return value. Eg `httpget` just writes the result into the `arg[2]` address. That's actually most io opcodes, but before we weren't using fork and join on opcodes directly.

avm/src/vm/event.rs
```diff
@@ -189,56 +186,77 @@ impl HandlerFragment {
...
-        // io-bound fragment
-        if !instructions[0].opcode.pred_exec {
-          let mem = Arc::new(RwLock::new(hand_mem));
-          let futures: Vec<EmptyFuture> = instructions.iter().map(|ins| {
+  pub async fn run_local(mut self: HandlerFragment, mut hand_mem: HandlerMemory) -> HandlerMemory {
+    let mut instructions = self.get_instruction_fragment();
+    loop {
+      // io-bound fragment
+      if !instructions[0].opcode.pred_exec {
+        if instructions.len() > 1 {
```
depombo, line 194: why is this necessary?

dfellis: It's not, anymore, but it was a nice shortcut when we were going to run them in parallel since it would limit the number of forks (memcpys) created when it wasn't necessary at all.

dfellis: pretty much 100% of the perf boost of this PR is just eliminating useless memcpys.

depombo (approved):

Issue #354, opened 2020-12-14
Fan out parallel array operations across cores


depombo:
AVM to avoid spinning up more than `n` (cpu core count) instances. `map`, `each`, `find` and all parallel array opcodes should generate `n` tasks for each cpu core instead of `m` tasks for each element in the array

depombo:
https://github.com/udoprog/unicycle
https://docs.rs/unicycle/0.7.0/unicycle/

depombo:
Implementation should take into account the results of https://github.com/alantech/alan/issues/387

PR #357, opened 2020-12-15
upgrade rust and tokio + set max threads to num of cpus

depombo:


dfellis:
This doesn't actually fix the real issue of #354 which is that we're still creating a tokio task per array element, which explodes the AVM with an OOM on large arrays unnecessarily (also because we duplicate the working memory for each task right now, so we get array size to the array size amount of memory consumption)

avm/Cargo.toml
```diff
@@ -5,7 +5,7 @@ license = "AGPL-3.0"
 homepage = "https://alan-lang.org"
 documentation = "https://docs.alan-lang.org"
 repository = "https://github.com/alantech/alan"
-version = "0.1.22"
+version = "0.1.48"
```
dfellis, line 8: Is this the Rust version? Or why did you change it?

avm/Cargo.toml
```diff
@@ -5,7 +5,7 @@ license = "AGPL-3.0"
...
 authors = ["Luis de Pombo <luis@alantechnologies.com>", "David Ellis <david@alantechnologies.com>"]
 edition = "2018"
```
dfellis, line 10: I assumed *this* was the field you need to change?

avm/Cargo.toml
depombo: this is the Rust version

avm/Cargo.toml
depombo: Initially, I thought so too

avm/Cargo.toml
dfellis: Not according to: https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field

dfellis: It appears you *can't* specify the Rust compiler version in Cargo? https://github.com/emk/heroku-buildpack-rust/issues/11 That explains why we've had such difficulties with the submitted projects compiling on our laptops.

dfellis: Tracing through several years of github issues leads me to: https://github.com/rust-lang/rustup/pull/1172

We apparently need to add a `rust-toolchain` file that contains, as far as I can tell, only the version of rust that we want to use?

dfellis:
It looks like this upgrade also broke the HTTP client and server code :/

depombo:
> This doesn't actually fix the real issue of #354 which is that we're still creating a tokio task per array element, which explodes the AVM with an OOM on large arrays unnecessarily (also because we duplicate the working memory for each task right now, so we get array size to the array size amount of memory consumption)

I see. To some extent I did not understand the issue correctly. It should not spin one thread per array element, but also it should not spin on task per array element. Since tokio uses a work stealing mechanism this addresses the former, but not the latter.

depombo:
> It looks like this upgrade also broke the HTTP client and server code :/

Yeah, I think this is a bug between `enable_all` not actually turning on `enable_io`

dfellis:
> > It looks like this upgrade also broke the HTTP client and server code :/
> 
> Yeah, I think this is a bug between `enable_all` not actually turning on `enable_io`

...that's crazy! What else would you mean by "enable_all"?

depombo:
> > > It looks like this upgrade also broke the HTTP client and server code :/
> > 
> > 
> > Yeah, I think this is a bug between `enable_all` not actually turning on `enable_io`
> 
> ...that's crazy! What else would you mean by "enable_all"?

So it turns out that it was silently failing because the latest version of `hyper` and `reqwst` are still using tokio 0.2. `hyper`'s master branch already made the change, but until their next release we can't upgrade. I'm going to abandon and make a new PR where I fix the actual issue of spinning up a tokio task per array element and set max threads without upgrading tokio

PR #361, opened 2020-12-20
Attempt to fix OSX CI failure

dfellis:
Reading the implementation of the opcodes, the CI failure *shouldn't* be happening -- unless the `futures` dependency isn't actually keeping the parallel requests in order? So trying to fix it with a version bump.


dfellis:
The actual issue was a bug in the AVM maintaining associativity guarantees of the `reducePar` implementations. It was showing up only on the OSX tests because Github recently increased the number of available cores from 2 to 3, causing an odd number of elements to reduce to appear.

depombo (approved):

PR #364, opened 2020-12-22
Result-ify math

dfellis:
To get it to work I needed to finally convert `int64` in the browser to `BigInt`, so there's also changes with the compiler to support that and updating seemingly unrelated js-runtime opcodes.

Array accesses, bitwise and comparison operators natively accept Result-wrapped values. Most of the time you won't have to worry about or unwrap the results of the math operations, but when you do need to make sure you actually have a number, you'll need to at that point.


bdd/spec/012_hash_spec.sh
```diff
@@ -29,15 +29,9 @@ Describe "Hashing"
...
-    JSHASHOUTPUT="-1058942856030168400
--5016367128657348000
--1058942856030168400
-6288867289231076000
--1521185239552941000" # TODO: Rounding should disappear once we can use BigInt consistently in JS
-
     It "runs js"
       When run test_js
-      The output should eq "$JSHASHOUTPUT"
+      The output should eq "$TOHASHOUTPUT"
```
dfellis, line 34: Very happy with this change. :)

bdd/spec/011_array_spec.sh
```diff
@@ -232,7 +232,7 @@ Hello, World!"
 
         on start {
           const count = [1, 2, 3, 4, 5]; // Ah, ah, ahh!
-          const byTwos = count.map(fn (n: int64): int64 = n * 2);
+          const byTwos = count.map(fn (n: int64): int64 = n * 2 || 0);
```
dfellis, line 235: This keeps the types the same, but I think if I just popped off the return type from this map function and removed the typed wrapper function around `toString` below it will also just work since you can print `Result<Stringifiable>` types.

Debating which to do here.

bdd/spec/021_trig_spec.sh
```diff
@@ -15,33 +15,33 @@ Describe "@std/trig"
         print(trig.log(e));
 
         'Basic Trig functions'.print();
-        print(trig.sin(tau / 6.0));
-        print(trig.cos(tau / 6.0));
-        print(trig.tan(tau / 6.0));
-        print(trig.sec(tau / 6.0));
-        print(trig.csc(tau / 6.0));
-        print(trig.cot(tau / 6.0));
+        print(trig.sin(tau / 6.0 || 0.0));
```
dfellis, line 18: I intend to undo all of this and instead edit the `std/trig.ln` file to accept `Result`-wrapped floats.

bdd/spec/026_datastore_spec.sh
```diff
@@ -27,7 +27,7 @@ Describe "@std/datastore"
       on waitAndInc fn (ms: int64) {
         wait(ms);
         let i = namespace('foo').getOr('inc', 0);
-        i = i + 1;
+        i = i + 1 || 0;
```
dfellis, line 30: Incrementing is annoying with Result-wrapped math outputs. The saturating math operators in the RFC would probably go better here.

compiler/src/ammtojs.ts
```diff
@@ -65,6 +65,13 @@ const assignableToJsText = (assignable: LPNode, indent: string) => {
     outText += assignable.get('variable').t
   } else if (assignable.has('value')) {
     outText += assignable.get('value').t
+    try {
+      const t = assignable.get('value').t
+      if (!/"/.test(t) && t !== 'true' && t !== 'false' && !/\./.test(t)) {
+        parseInt(assignable.get('value').t)
+        outText += 'n'
+      }
+    } catch (e) { }
```
dfellis, line 74: Since the kinds of values that can be assigned in AMM are so small, this works, even if it looks crazy. But it only works because right now you can't define integer constants that aren't `int64`. If we solve that, this breaks.

This may also need to be refactored from producing things like `2n` to producing `BigInt("2")` if Safari *still* doesn't have bigint literal support.

js-runtime/index.js
```diff
@@ -66,7 +78,19 @@ const hashv = arr => {
...
+  try {
+    return JSON.parse(JSON.stringify(a))
+  } catch (e) {
+    if (typeof a[0] === 'bigint') {
+      return a.map(v => BigInt(v))
+    } else {
+      return a.map(v => copyarr(v))
+    }
+  }
 }
```
dfellis, line 94: *Please* let me know if you can think of a better way. To deep clone in Javascript it's *recommended* to use JSON.parse/JSON.stringify even though that fails silently for Date objects and crashes on BigInt. This is my solution to that problem, but I know it's going to kill JS perf any time you need to clone an array of data.

js-runtime/index.js
```diff
@@ -142,59 +166,469 @@ module.exports = {
...
-  addf32:  (a, b) => a + b,
-  addf64:  (a, b) => a + b,
-
-  subi8:   (a, b) => a - b,
-  subi16:  (a, b) => a - b,
-  subi32:  (a, b) => a - b,
-  subi64:  (a, b) => a - b,
-  subf32:  (a, b) => a - b,
-  subf64:  (a, b) => a - b,
+  addi8:   (ra, rb) => {
```
dfellis, line 169: These new opcode implementations follow a pattern.

js-runtime/index.js
```diff
@@ -142,59 +166,469 @@ module.exports = {
...
+    if (!ra[0]) return ra
+    if (!rb[0]) return rb
```
dfellis, line 171: First, if either result-wrapped integer is an error, return that error result.

js-runtime/index.js
```diff
@@ -142,59 +166,469 @@ module.exports = {
...
+    const a = ra[1]
+    const b = rb[1]
```
dfellis, line 173: Otherwise extract the integers

js-runtime/index.js
```diff
@@ -142,59 +166,469 @@ module.exports = {
...
+    if (a === 0 || b === 0) return [1, a + b]
+    if (a < 0 && b > 0) return [1, a + b]
+    if (a > 0 && b < 0) return [1, a + b]
```
dfellis, line 176: For cases where it's impossible to overflow/underflow, return a result-wrapped addition (or subtraction, multiplication, etc)

js-runtime/index.js
```diff
@@ -142,59 +166,469 @@ module.exports = {
...
+    if (a > 0 && b > 0) {
+      if (a > INT8MAX - b) return [0, 'overflow']
+      return [1, a + b]
+    }
```
dfellis, line 180: When you *might* have a case that could fail, do the math to check it, and return an error result if it would have failed.

js-runtime/index.js
```diff
@@ -142,59 +166,469 @@ module.exports = {
...
+  addf32:  (ra, rb) => {
+    if (!ra[0]) return ra
+    if (!rb[0]) return rb
+    const a = ra[1]
+    const b = rb[1]
+    const out = a + b
+    if (out === Number.POSITIVE_INFINITY) return [0, 'overflow']
+    if (out === Number.NEGATIVE_INFINITY) return [0, 'underflow']
+    return [1, out]
+  },
```
dfellis, line 246: While it's kinda dumb for floating point numbers, since they can't actually underflow/overflow/etc from a CPU hardware perspective, doing this keeps it consistent with the integers and forces you to deal with error situations where not even 64-bit floats could represent your number.

dfellis: And the proposed "saturating math" operators restore the "normal" floating point behavior and give integers a similar behavior.

js-runtime/index.js
```diff
@@ -296,14 +730,14 @@ module.exports = {
   // String opcodes
   catstr:  (a, b) => a.concat(b),
   split:   (a, b) => a.split(b),
-  repstr:  (a, b) => new Array(b).fill(a).join(''),
+  repstr:  (a, b) => new Array(parseInt(b.toString())).fill(a).join(''),
```
dfellis, line 733: Various random opcodes that take `int64` inputs need to convert to JS `Number`s to actually use them inside of the runtime. Technically this means that some program behaviors that are valid in the AVM do not work in the JS runtime, but generally for things that aren't *practically* valid in the AVM, either, because they involve arrays greater than 2^53 elements long, so I'm not worried about this for the next, uh, century?

js-runtime/index.js
```diff
@@ -647,7 +1088,7 @@ module.exports = {
   },
   httplsn:  async (port) => {
     const server = http.createServer((req, res) => {
-      const connId = hashf(Math.random().toString())
+      const connId = Number(hashf(Math.random().toString()))
```
dfellis, line 1091: You apparently can't use a `BigInt` as an object key, even though `BigInt` has a `toString` method. 🤨

std/deps.ln
```diff
@@ -12,8 +12,8 @@ export event install: void
 export fn add(remote: string) {
   // TODO implement proper error handling
   const parts = remote.split('/');
-  const repo = parts[length(parts) - 1] || '';
-  const group = parts[parts.length() - 2] || '';
+  const repo = parts[length(parts) - 1 || 0] || '';
+  const group = parts[parts.length() - 2 || 0] || '';
```
dfellis, line 16: Array accesses should also be allowed to be done from Result-wrapped `int64`s, since the array access itself can fail, it could fail based on the fact that the input value has already failed, and eliminate this need to unwrap the result.

This might need some compiler changes, though, since array accesses are manually converted into certain opcode calls.

std/root.ln
```diff
@@ -189,6 +525,7 @@ export fn max(x: Orderable, y: Orderable): Orderable {
 }
 
 // Boolean and bitwise functions
+// TODO: Should Result-wrapped integers be accepted by the bitwise functions?
```
dfellis, line 528: Maybe? In the same vein as the `abs` functions above, where it doesn't result-wrap if both inputs are "bare"?

std/root.ln
```diff
@@ -232,6 +569,7 @@ export fn xnor(a: int64, b: int64) = xnori64(a, b);
 export fn xnor(a: bool, b: bool) = xnorboo(a, b);
 
 // Equality and order functions
+// TODO: Similarly, should equality/orderability functions accept Result-wrapped values?
```
dfellis, line 572: Maybe, but then we get into a slippery slope of implicit type conversions and seemingly impossible comparison results.

Possibilities here:
1. Keep it as-is -- the user has to manually unwrap the result-wrapped integer to do a comparison, eg `if j < (i - 1 || 0) ...` This adds syntactic noise but is clearer.
2. Result-wrap the comparison if Result-wrapped integers are involved, but this requires unwrapping the final comparison `if j < i - 1 || false ...` This makes it clearer that on failure you're explicitly choosing the success/failure of the conditional, but it still looks really weird.
3. Comparisons with result-wrapped integers always return an unwrapped boolean `if j < i - 1 ...` looks like before this PR, but has implicit rules to memorize, like "if either comparison input is an error type, returns false" and can cause weird things like `if j < i - 1 ... else if i - 1 < j ...` both being `false`.

The third behavior is most like how Javascript does things. The first behavior is most like Rust, while the middle is an "infectious" style kinda like Java exception bubbling. (The error opcode is simply passed up to something else that cares about it.)

I am personally leaning towards (3) as the code looks the closest to other popular languages and the comparison rules are also similar:

```
damocles@Talyn:/usr/local/alancorp/alan(resultify-math)$ node
> 3 < undefined
false
> undefined < 3
false
>
```
But we didn't discuss this issue in the RFC and so it would be nice for some thoughts from others.

std/root.ln
```diff
@@ -485,7 +823,7 @@ export fn set(arr: Array<bool>, idx: int64, val: bool) {
 
 // Ternary functions
 export fn pair(trueval: any, falseval: any) = new Array<any> [ trueval, falseval ];
-export fn cond(c: bool, options: Array<any>) = getR(options[1.sub(c.toInt64())]);
+export fn cond(c: bool, options: Array<any>) = getR(options[1.sub(c.toInt64()).getR()]);
```
dfellis, line 826: This `getR` won't be necessary if we adjust array accessors to accept Result-wrapped integers.

std/root.ln
```diff
@@ -638,7 +898,7 @@ export fn set(hm: HashMap<Hashable, any>, key: Hashable, val: any): HashMap<Hash
   const list = getR(hm.lookup[hash]);
   if list.length() == 8 {
     // Rebucket everything
-    const lookupLen = length(hm.lookup) * 2;
+    const lookupLen = length(hm.lookup) * 2 || 0;
     hm.lookup = new Array<Array<int64>> [ new Array<int64> [], ] * lookupLen;
```
dfellis, line 902: An unexpected impact of this is all of the other opcodes that take integers. Without unwrapping this integer you can't use it in the array construction line below. But if we allowed Result-wrapped values to be passed to the `repeat` function for arrays, it could fail besides OOM failure, and then you would need to unwrap the array.

Here it seems like adding Result-wrapped integer support would make things worse, not better, as the noise is unavoidable.

std/root.ln
```diff
@@ -791,7 +1051,7 @@ export fn prune(n: Node<any>): Tree<any> {
   // This is, unfortunately for now, a sequential algorithm. Hope to figure out a parallel version
   let nodeStack = new Array<int64> [ n.id ];
   let rmdIds = new Array<int64> [ ];
-  seqdo(newseq(pow(2, 62)), fn (): bool {
+  seqdo(newseq(pow(2, 62).getOr(0)), fn (): bool {
```
dfellis, line 1054: Similarly here, constructing a sequence type shouldn't fail, so this unwrapping is necessary.

std/root.ln
```diff
@@ -804,7 +1064,7 @@ export fn prune(n: Node<any>): Tree<any> {
       const childrenIds = childrenRes.getR();
       nodeStack = nodeStack.concat(childrenIds);
     }
-    const delIdx = nodeId - length(rmdIds);
+    const delIdx = nodeId - length(rmdIds) || 0;
     delete(n.tree.vals, delIdx);
     delete(n.tree.parents, delIdx);
     delete(n.tree.children, delIdx);
```
dfellis, line 1070: But array deletion can fail if the index is out-of-bounds, so it feels like this should accept a Result-wrapped index since it can just fail early if the input index is an error.

std/trig.ln
```diff
@@ -1,9 +1,11 @@
+// TODO: Probably change this from always unwrapping Result-wrapped values to passing along
+// result-wrapped errors
```
dfellis, line 2: Not probably, definitely. The usage of this version of `trig` in the test suite is terrible.

std/root.ln
depombo: I much prefer the third option to the rest so that sounds good!

std/root.ln
depombo: I have already expressed my recent reservations about how eliminating most runtime errors make things more annoying and less usable. I only realized this after I started writing Alan in the standard library for the `Tree` type. Undoubtedly usability will improve once the compiler is smart enough to unwrap a Result within an  `isErr` or `isOk` conditional, but it is not clear to me by how much. As an individual developer, I would love to have the guarantee of most runtime errors impossible but it depends on the cost of usability I have to pay for it.

It also feels like it would not be the main reason someone would use Alan. This is true for me at least. I think we definitely want to add it at some point, but it doesn't feel correct to have it part of our MVP. Basically to me automatic parallelization >> most runtime errors impossible.

compiler/src/lntoamm/Microstatement.ts
```diff
@@ -255,17 +255,44 @@ ${varAst.getText()} on line ${varAst.start.line}:${varAst.start.column}`)
           const lookup = microstatements[microstatements.length - 1]
           // TODO: Map support, which requires figuring out if the outer memory object is an array
           // or a map.
-          if (lookup.outputType.typename !== 'int64') {
-            throw new Error(`${segment.getText()} cannot be used in an array lookup as it is not an int64
+          if (lookup.outputType.typename === 'int64') {
```
depombo, line 258: will this affect aliasing to `int` at all?

js-runtime/index.js
depombo: ugh

depombo (approved):

compiler/src/lntoamm/Microstatement.ts
dfellis: No, the microstatements don't use the aliases.

js-runtime/index.js
dfellis: I need to think about that. The fewer deps the better, imo. Though I did confirm that it does work on BigInt even though it doesn't have any special case code for them. Not sure what magic they're doing there.

bdd/spec/027_seq_spec.sh
```diff
@@ -160,15 +160,15 @@ error: sequence out-of-bounds"
...
+              const prev = self.recurse(i - 1 || 0);
+              const prevPrev = self.recurse(i - 2 || 0);
               if prev.isErr() {
                 return prev;
               }
               if prevPrev.isErr() {
                 return prevPrev;
               }
-              return ok((prev || 1) + (prevPrev || 1));
+              return (prev || 0) + (prevPrev || 0);
```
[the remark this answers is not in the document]
dfellis: Oh, because I was silly here. Let me remove that.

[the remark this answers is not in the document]
dfellis: Wait, no. I think there's an actual issue here wrt the inferred type of the variable because of how `recurse` has that opaque `Self` type with no generics data to help, and that's why I put that in there.

I will update with a TODO on when recurse has proper type inference.

bdd/spec/027_seq_spec.sh
```diff
@@ -258,11 +258,11 @@ error: sequence out-of-bounds"
         import @std/app
         from @std/seq import seq, Self, recurse
 
-        fn triangularRec(x: int) : int = seq(x + 1).recurse(fn (self: Self, x: int) : Result<int> {
+        fn triangularRec(x: int) : int = seq(x + 1 || 0).recurse(fn (self: Self, x: int) : Result<int> {
           if x == 0 {
             return ok(x);
           } else {
-            return ok(x + (self.recurse(x - 1) || 0));
+            return x + (self.recurse(x - 1 || 0) || 0);
```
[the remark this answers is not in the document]
dfellis: Same issue here. Type inference inside of recursive functions is a bit broken and we need to fix it. But since we already want to redo this obtuse syntax I don't think this should be a blocker?

cdmistman (approved):

PR #368, opened 2021-01-01
Drop ANTLR

dfellis:
This is now ready. Replaced ANTLR with LP, added a couple more classes to LP to get it to work, simplified operators dramatically, and finally figured out what was slowing down LP and now we're ~30x faster than ANTLR at parsing the AST!


dfellis:
Will mark off the test suites as they are fixed up in this comment:

- [x] bdd/spec/001_event_spec.sh
- [x] bdd/spec/002_printing.sh
- [x] bdd/spec/003_math_spec.sh
- [x] bdd/spec/004_bitwise_spec.sh
- [x] bdd/spec/005_boolean_spec.sh
- [x] bdd/spec/006_string_spec.sh
- [x] bdd/spec/007_comparator_spec.sh
- [x] bdd/spec/008_function_spec.sh
- [ ] bdd/spec/009_conditional_spec.sh
- [ ] bdd/spec/010_object_literal_spec.sh
- [ ] bdd/spec/011_array_spec.sh
- [ ] bdd/spec/012_hash_spec.sh
- [ ] bdd/spec/013_generics_spec.sh
- [ ] bdd/spec/014_interface_spec.sh
- [ ] bdd/spec/015_maybe_result_either_spec.sh
- [ ] bdd/spec/016_type_spec.sh
- [ ] bdd/spec/017_custom_event_spec.sh
- [ ] bdd/spec/018_closure_spec.sh
- [x] bdd/spec/019_compiler_error_spec.sh
- [ ] bdd/spec/020_module_constants_spec.sh
- [ ] bdd/spec/021_trig_spec.sh
- [ ] bdd/spec/022_deps_spec.sh
- [ ] bdd/spec/023_http_spec.sh
- [x] bdd/spec/024_clone_spec.sh
- [x] bdd/spec/025_runtime_error_spec.sh
- [ ] bdd/spec/026_datastore_spec.sh
- [ ] bdd/spec/027_seq_spec.sh
- [ ] bdd/spec/028_tree_spec.sh
- [x] bdd/spec/029_err_printing_spec.sh
- [ ] bdd/spec/030_cmd_spec.sh

compiler/src/lntoamm/UserFunction.ts
```diff
@@ -33,92 +33,116 @@ class UserFunction implements Fn {
...
+      if (functionishAst.has('functions')) {
+        return UserFunction.fromFunctionsAst(functionishAst.get('functions'), scope)
+      }
+      if (functionishAst.has('functionbody')) {
+        return UserFunction.fromFunctionbodyAst(functionishAst.get('functionbody'), scope)
       }
-      if (functionishAst.functionbody() != null) {
-        return UserFunction.fromFunctionbodyAst(functionishAst.functionbody(), scope)
+      if (functionishAst.has('fnname')) {
+        // TODO: We didn't cover this path before?
```
cdmistman, line 56: does this need to get implemented right now?

compiler/src/lntoamm/Microstatement.ts
```diff
@@ -179,190 +180,35 @@ class Microstatement {
     return original
   }
 
-  // TODO: Eliminate ANTLR
-  static fromVarAst(varAst: any, scope: Scope, microstatements: Array<Microstatement>) {
```
dfellis, line 183: I had started to translate this, and then I noticed that absolutely nothing used it. It should've been removed in the statement refactor PR last month.

compiler/src/lntoamm/Statement.ts
```diff
@@ -43,106 +49,76 @@ class Statement {
...
-      s.exits().assignables()
-    )
-    if (s.emits() && s.emits().assignables()) return Statement.assignablesHasObjectLiteral(
-      s.emits().assignables()
-    )
-    // TODO: Cover conditionals
-    return false
-  }
-
-  static isCallPure(callAst: any, scope: Scope) { // TODO: Migrate off ANTLR
```
dfellis, line 68: I deleted this because while checking the purity of a function is desired so we know whether or not to automatically un-parallelize Array methods, the prior implementation didn't work and we weren't using it, so we should just add it back from a first-principles perspective in the future.

compiler/src/lntoamm/Type.ts
```diff
@@ -153,51 +154,71 @@ export class Interface {
...
+    ) {
+      const interfacelist = interfaceAst
+        .get('interfacedef')
+        .get('interfacebody')
+        .get('interfacelist')
+      const interfacelines = []
+      interfacelines.push(interfacelist.get('interfaceline'))
+      interfacelist.get('cdr').getAll().forEach(l => {
+        interfacelines.push(l.get('interfaceline'))
+      })
```
dfellis, line 181: This particular pattern to turn comma/semicolon/etc separated lists into arrays is all over the place. It unfortunately bloats up what was 1 LOC in ANTLR to 5 LOC in LP.

Figuring out a way to restore it in LP would be nice.

compiler/src/lntoamm/UserFunction.ts
dfellis: Well, can it be a follow-up? We apparently never ran into it before. I need to see what paths even lead to here, though. This is creating a function from the AST, so maybe it would be something as simple as `const myFn: function = someOtherFunction;`?

compiler/src/lp.ts
```diff
@@ -552,6 +565,138 @@ export class Or implements LPNode {
   }
 }
 
+export class ExclusiveOr implements LPNode {
```
dfellis, line 568: Didn't actually use this class, but it seems like a waste to delete?

depombo (approved):
This is amazing! Pending test passing of course!

PR #370, opened 2021-01-04
Upgrade the AVM to the latest version of tokio

### LINKED ISSUE #369: Upgrade the AVM to the latest version of tokio
- [ ] Replace reqwest in http client opcodes with hyper. This removes our dependency to reqwest which is pinned to tokio 0.2. This will make the AVM smaller and allow us to upgrade tokio to 0.3.
- [ ] Upgrade hyper to the latest version that uses tokio 0.3
- [ ] Upgrade tokio to 0.3 in the AVM

cdmistman:
Fixes #369


avm/src/vm/opcode.rs
```diff
@@ -2078,19 +2078,19 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
   });
   io!("httpget", |args, mut hand_mem| {
     Box::pin(async move {
-      let url = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[0]));
-      let http_res = reqwest::get(&url).await;
+      let url: Uri = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[0])).try_into().unwrap();
```
dfellis, line 2081: This will cause this particular opcode to potentially crash the AVM if it is fed an invalid URL, right?

Maybe we explicitly handle that and return an error result, instead?

avm/src/vm/opcode.rs
```diff
@@ -2078,19 +2078,19 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      let http_res = Client::builder().http2_only(true).build_http::<hyper::Body>().get(url.clone()).await;
```
dfellis, line 2082: Why is this forcing http 2.0?

avm/src/vm/opcode.rs
cdmistman: in the new commit, it returns an error if there's a problem parsing the URI

avm/src/vm/opcode.rs
cdmistman: removed in latest commit

cdmistman:
forgot about actually upgrading to tokio 1, next commit should finish this PR

cdmistman:
MacOS CI (and local build) fails because `mio:0.7.7` is using `libc:0.2.69`, whereas `libc:0.2.71` is the minimum required version to use `libc::TCP_KEEPINTVL` and `libc::TCP_CNT` (introduced in [this PR](https://github.com/rust-lang/libc/pull/1776)). See tokio-rs/mio#1433 for more info. Blocked on tokio-rs/mio#1434 merge, `mio:0.7.8`, and all mio-based dependencies converging on the updated crate.

dfellis:
> and all mio-based dependencies converging on the updated crate.

Is it not possible to override the dependencies of other dependencies in Rust? Worst-case by hand-editing `Cargo.lock`? I though that's what that file was for.

avm/src/vm/opcode.rs
```diff
@@ -2072,51 +2073,49 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      let (res, ok) = match TryInto::<Uri>::try_into(HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[0]))) {
+        Err(ee) => (format!("{}", ee), false),
+        Ok(url) => match Client::builder().build::<_, Body>(HttpsConnector::new()).get(url.clone()).await {
+          Err(ee) => (format!("{}", ee), false),
+          Ok(mut response) => match hyper::body::to_bytes(response.body_mut()).await {
+            Err(ee) => (format!("{}", ee), false),
+            Ok(body) => if let Ok(res) = String::from_utf8(body.to_vec()) {
+              (res, true)
+            } else {
+              (format!("could not parse"), false)
```
dfellis, line 2091: Nit: Does this need the `format!` here?

avm/src/vm/opcode.rs
```diff
@@ -2072,51 +2073,49 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
     })
   });
   io!("httppost", |args, mut hand_mem| {
     Box::pin(async move {
       let url = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[0]));
       let payload = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[1]));
-      let client = reqwest::Client::new();
-      let http_res = client.post(&url).body(payload.clone()).send().await;
+      let client = Client::builder().build::<_, Body>(HttpsConnector::new());
+      let http_res = client.request(Request::post(&url).body(payload.clone().into()).unwrap()).await;
```
dfellis, line 2107: I didn't catch this before, but this has a similar potential runtime crash with the `unwrap` here. Can we also guard this if it's a credible failure scenario? (Presumably again due to malformed URLs?)

dfellis (approved):

depombo (approved):

Issue #356, opened 2020-12-14
make `fork` and `join` cheaper in memcpys by changing the memory model


depombo:

dfellis:
![IMG_20201214_133140](https://user-images.githubusercontent.com/765551/102138196-d6424c00-3e10-11eb-9dd1-195cbb1422ea.jpg)

Here's a diagram of how it could work with minimal changes.

The HandlerMemory object has an optional parent pointer `Option<*HandlerMemory>` and the address arrays are now `Vec<Option<(usize, usize)>>`.

Reading data would first check if the optional is defined in the current memory space, otherwise return the parent's memory read (or fail if no parent). Writing data would always write into the current HandlerMemory space.

There would need to be new API work to read from fractal memory (perhaps as an iterator?) because now you can't guarantee that the virtual pointer logic is on the current HandlerMemory or the parent and where you need to read.

If you want to get a mutable version of data in the parent scope, it needs to copy it on the mutable read request so updates work as expected.

This makes `fork` a very cheap object creation, while `join` should be roughly the same complexity, except now it's easier to determine if any particular address was *actually* updated in the fork or not, meaning the guessing logic that was attempted before (and didn't really work) won't be necessary.

There's still the issue of unreachable data not being dropped until the handler finishes running, but for opcodes that are known to not mutate the outer scope (and almost always only return data through the `args` address space) there can be a special, simpler `join_retval` method that only rejoins the return value into the original scope to let the intermediate data die at that point.

depombo:
> There's still the issue of unreachable data not being dropped until the handler finishes running, but for opcodes that are known to not mutate the outer scope (and almost always only return data through the `args` address space) there can be a special, simpler `join_retval` method that only rejoins the return value into the original scope to let the intermediate data die at that point.

Do you mean this https://github.com/alantech/alan/blob/main/avm/src/vm/opcode.rs#L1611?

dfellis:
Yes, exactly. :)

PR #371, opened 2021-01-05
Trim and refactor the AVM's fractal memory abstraction

depombo:
preliminary PR that tries to reinforce the memory abstraction by making opcodes as unaware as possible of how memory indices works. This sets the field for the next PR which will solve https://github.com/alantech/alan/issues/356 and hopefully require very little to no changes to `opcode.rs`

### DIFF
--- a/avm/src/vm/memory.rs
+++ b/avm/src/vm/memory.rs
@@ -30,7 +30,7 @@ fn addr_type(addr: i64) -> i8 {
 }
 
 /// Memory representation of a fractal memory block within HandlerMemory
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug)]
 pub struct FractalMemory {
   // address in HandlerMemory if it is not a nested fractal
   hm_addr: Option<i64>,
@@ -62,6 +62,13 @@ impl FractalMemory {
       panic!("Trying to read raw data from memory when it is a pointer")
     }
     return self.block[idx].1;
+  }
+}
+
+impl PartialEq for FractalMemory {
+  fn eq(&self, other: &Self) -> bool {
+    // ignore hm_addr
+    self.block == other.block
   }
 }
 
@@ -217,9 +224,6 @@ impl HandlerMemory {
   /// values, and returns that value either as a vector or the singular value wrapped in a vector,
   /// and a boolean indicating if it is a fractal value or not.
   pub fn read_from_fractal(self: &HandlerMemory, fractal: &FractalMemory, idx: usize) -> (FractalMemory, bool) {
-    if fractal.hm_addr.is_none() {
-      panic!("HandlerMemory read called from a deeply nested Fractal");
-    }
     let (a, b) = fractal.block[idx];
     let b_usize = b as usize;
     return if a == std::usize::MAX {
@@ -242,11 +246,10 @@ impl HandlerMemory {
 
   /// Stores a nested fractal of data in a given address.
   pub fn write_fixed_in_fractal(self: &mut HandlerMemory, fractal: &mut FractalMemory, idx: usize, val: i64) {
-    if fractal.hm_addr.is_none() {
-      panic!("HandlerMemory write called from a deeply nested Fractal");
-    }
     fractal.block[idx].1 = val;
-    self.write_fractal(fractal.hm_addr.unwrap(), fractal);
+    if fractal.hm_addr.is_some() {
+      self.write_fractal(fractal.hm_addr.unwrap(), fractal);
+    }
   }
 
   /// Stores a fixed value in a given address. Determines where to place it based on the kind of
--- a/avm/src/vm/memory.rs
+++ b/avm/src/vm/memory.rs
@@ -319,10 +319,7 @@ impl HandlerMemory {
   pub fn pop(self: &mut HandlerMemory, addr: i64) -> Result<FractalMemory, String> {
     let mem = self.read_mut_fractal(addr);
     if mem.len() > 0 {
-      return Ok(FractalMemory{
-        block: vec![mem.pop().unwrap()],
-        hm_addr: Some(addr),
-      });
+      return Ok(FractalMemory::new(vec![mem.pop().unwrap()]));
     } else {
       return Err("cannot pop empty array".to_string());
     }
@@ -332,10 +329,7 @@ impl HandlerMemory {
   pub fn delete(self: &mut HandlerMemory, addr: i64, idx: usize) -> Result<FractalMemory, String> {
     let mem = self.read_mut_fractal(addr);
     if mem.len() > 0 && mem.len() > idx {
-      return Ok(FractalMemory{
-        block: vec![mem.remove(idx)],
-        hm_addr: Some(addr),
-      });
+      return Ok(FractalMemory::new(vec![mem.remove(idx)]));
     } else {
       return Err(format!("cannot remove idx {} from array with length {}", idx, mem.len()));
     }
--- a/avm/src/vm/memory.rs
+++ b/avm/src/vm/memory.rs
@@ -29,9 +29,51 @@ fn addr_type(addr: i64) -> i8 {
   }
 }
 
+/// Memory representation of a fractal memory block within HandlerMemory
+#[derive(Clone, Debug, PartialEq)]
+pub struct FractalMemory {
+  // address in HandlerMemory if it is not a nested fractal
+  hm_addr: Option<i64>,
+  // a memory block from HandlerMemory.mems
+  block: Vec<(usize, i64)>,
+}
+
+impl FractalMemory {
+  // Called to store actual data as opposed to fixed or nested data
+  pub fn new(block: Vec<(usize, i64)>) -> FractalMemory {
+    return FractalMemory {
+      hm_addr: None,
+      block: block,
+    }
+  }
+
+  fn new_nested(self: &FractalMemory, block: Vec<(usize, i64)>) -> FractalMemory {
+    return FractalMemory {
+      hm_addr: None,
+      block: block,
+    }
+  }
+
+  pub fn len(self: &FractalMemory) -> usize {
+    return self.block.len();
+  }
+
+  // Compare the blocks at given index between two FractalMemory
+  pub fn compare_at(self: &FractalMemory, idx: usize, other: &FractalMemory) -> bool {
+    return self.block[idx] == other.block[idx];
+  }
+
+  /// Reads fixed data from a given address.
+  pub fn read_fixed(self: &FractalMemory, idx: usize) -> i64 {
+    if self.block[idx].0 != usize::MAX {
+      panic!("Trying to read raw data from memory when it is a pointer")
+    }
+    return self.block[idx].1;
+  }
+}
+
 /// Memory representation of a handler call
-#[derive(Clone)]
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct HandlerMemory {
   /// The set of memory blocks. The first (zeroth) block hosts global memory and all blocks
   /// afterwards host memory created by the handler. Each block consists of tuples of two values,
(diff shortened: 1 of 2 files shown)


avm/src/vm/memory.rs
```diff
@@ -233,28 +309,33 @@ impl HandlerMemory {
...
 
   /// Pops a value off of the fractal. May be fixed data or a virtual pointer.
-  pub fn pop(self: &mut HandlerMemory, addr: i64) -> Result<(usize, i64), String> {
+  pub fn pop(self: &mut HandlerMemory, addr: i64) -> Result<FractalMemory, String> {
     let mem = self.read_mut_fractal(addr);
     if mem.len() > 0 {
-      return Ok(mem.pop().unwrap());
+      return Ok(FractalMemory{
+        block: vec![mem.pop().unwrap()],
+        hm_addr: Some(addr),
```
dfellis, line 324: Since it was popped off of the array, should it actually be getting the address here?

avm/src/vm/opcode.rs
```diff
@@ -1560,90 +1556,80 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
   cpu!("poparr", |args, hand_mem| {
     let last = hand_mem.pop(args[0]);
-    hand_mem.write_fractal(args[2], &Vec::new());
+    hand_mem.init_fractal(args[2]);
     if last.is_ok() {
       hand_mem.push_fixed(args[2], 1i64);
       let record = last.ok().unwrap();
-      let (a, b) = record;
-      hand_mem.push_idxs(args[2], a, b as usize);
+      hand_mem.push_register_from_fractal(args[2], &record, 0);
```
dfellis, line 1563: This is such a weird method name. I wonder if there's any better name, or if it would be better to figure out a way to make smaller, more composable functions here?

avm/src/vm/opcode.rs
```diff
@@ -1560,90 +1556,80 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      let len = arrv.len();
-      drop(arr);
+      let fractal = hand_mem.read_fractal(args[0]);
       let subhandler = HandlerFragment::new(args[1], 0);
-      hand_mem.write_fractal(args[2], &Vec::new());
-      for i in 0..len {
-        hand_mem.set_addr(CLOSURE_ARG_MEM_START + 1, arrv[i].0, arrv[i].1 as usize);
+      hand_mem.init_fractal(args[2]);
+      for i in 0..fractal.len() {
+        hand_mem.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1616: I don't think you need this. Just `hand_mem.register_out(args[0], i, CLOSURE_ARG_MEM_START + 1)`

dfellis: If it's only used in situations like this, we can probably drop the `register_from_fractal` method.

avm/src/vm/opcode.rs
```diff
@@ -1652,22 +1638,21 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      let arr = hand_mem.read_fractal(args[0]);
-      let len = arr.len();
+      let fractal = hand_mem.read_fractal(args[0]);
       let subhandler = HandlerFragment::new(args[1], 0);
       let mut runners = Vec::new();
-      for i in 0..len {
+      for i in 0..fractal.len() {
         let mut hm = hand_mem.fork();
-        hm.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
+        hm.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1655: Same here, `register_out` would also work.

avm/src/vm/opcode.rs
```diff
@@ -1677,11 +1662,10 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
   io!("eachl", |args, mut hand_mem| {
     Box::pin(async move {
-      let arr = hand_mem.read_fractal(args[0]).to_vec();
-      let len = arr.len();
+      let fractal = hand_mem.read_fractal(args[0]);
       let subhandler = HandlerFragment::new(args[1], 0);
-      for i in 0..len {
-        hand_mem.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
+      for i in 0..fractal.len() {
+        hand_mem.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1668: Same

avm/src/vm/opcode.rs
```diff
@@ -1690,65 +1674,61 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      let arr = hand_mem.read_fractal(args[0]);
-      let len = arr.len();
+      let fractal = hand_mem.read_fractal(args[0]);
+      let len = fractal.len();
       let subhandler = HandlerFragment::new(args[1], 0);
       let mut finders = Vec::new();
       for i in 0..len {
         let mut hm = hand_mem.fork();
-        hm.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
+        hm.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1683: 'nother one

avm/src/vm/opcode.rs
```diff
@@ -1690,65 +1674,61 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
   io!("findl", |args, mut hand_mem| {
     Box::pin(async move {
-      let arr = hand_mem.read_fractal(args[0]).to_vec();
-      let len = arr.len();
+      let fractal = hand_mem.read_fractal(args[0]);
       let subhandler = HandlerFragment::new(args[1], 0);
-      for i in 0..len {
-        hand_mem.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
+      for i in 0..fractal.len() {
+        hand_mem.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1708: here too

avm/src/vm/opcode.rs
```diff
@@ -1765,11 +1745,10 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
   io!("somel", |args, mut hand_mem| {
     Box::pin(async move {
-      let arr = hand_mem.read_fractal(args[0]).to_vec();
-      let len = arr.len();
+      let fractal = hand_mem.read_fractal(args[0]);
       let subhandler = HandlerFragment::new(args[1], 0);
-      for i in 0..len {
-        hand_mem.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
+      for i in 0..fractal.len() {
+        hand_mem.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1751: also true here

avm/src/vm/opcode.rs
```diff
@@ -1690,65 +1674,61 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      let arr = hand_mem.read_fractal(args[0]);
-      let len = arr.len();
+      let fractal = hand_mem.read_fractal(args[0]);
       let subhandler = HandlerFragment::new(args[1], 0);
       let mut somers = Vec::new();
-      for i in 0..len {
+      for i in 0..fractal.len() {
         let mut hm = hand_mem.fork();
-        hm.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
+        hm.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1731: and here

avm/src/vm/opcode.rs
```diff
@@ -1783,13 +1762,12 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
```
dfellis, line 1770: 'nother

avm/src/vm/opcode.rs
```diff
@@ -1806,11 +1784,10 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
   io!("everyl", |args, mut hand_mem| {
     Box::pin(async move {
-      let arr = hand_mem.read_fractal(args[0]).to_vec();
-      let len = arr.len();
+      let fractal = hand_mem.read_fractal(args[0]);
       let subhandler = HandlerFragment::new(args[1], 0);
-      for i in 0..len {
-        hand_mem.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
+      for i in 0..fractal.len() {
+        hand_mem.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1790: this one too

avm/src/vm/opcode.rs
```diff
@@ -1823,28 +1800,24 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
   io!("reducep", |args, mut hand_mem| {
     Box::pin(async move {
-      let arr = hand_mem.read_fractal(args[0]).to_vec();
+      let fractal = hand_mem.read_fractal(args[0]);
       let mut vals: Vec<HandlerMemory> = vec![];
-      for i in 0..arr.len() {
+      for i in 0..fractal.len() {
         let mut hm = HandlerMemory::new(None, 1);
-        hand_mem.set_addr(CLOSURE_ARG_MEM_START, arr[i].0, arr[i].1 as usize);
+        hand_mem.register_from_fractal(CLOSURE_ARG_MEM_START, &fractal, i);
```
dfellis, line 1820: and here

avm/src/vm/opcode.rs
```diff
@@ -1880,14 +1852,14 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      let fractal = hand_mem.read_fractal(args[0]);
+      if fractal.len() == 0 {
         return hand_mem;
       }
       let mut vals: Vec<HandlerMemory> = vec![];
-      for i in 0..arr.len() {
+      for i in 0..fractal.len() {
         let mut hm = HandlerMemory::new(None, 1);
-        hand_mem.set_addr(CLOSURE_ARG_MEM_START, arr[i].0, arr[i].1 as usize);
+        hand_mem.register_from_fractal(CLOSURE_ARG_MEM_START, &fractal, i);
```
dfellis, line 1862: and here too

avm/src/vm/opcode.rs
```diff
@@ -1907,19 +1879,16 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      let (c, d) = obj[1];
-      let (arr, _) = hand_mem.read_either_idxs(a, b as usize);
-      let arrv = arr.to_vec();
+      let (arr, _) = hand_mem.read_from_fractal(&obj, 0);
       let mut vals: Vec<HandlerMemory> = vec![];
-      for i in 0..arrv.len() {
+      for i in 0..arr.len() {
         let mut hm = HandlerMemory::new(None, 1);
-        hand_mem.set_addr(CLOSURE_ARG_MEM_START, arrv[i].0, arrv[i].1 as usize);
+        hand_mem.register_from_fractal(CLOSURE_ARG_MEM_START, &arr, i);
```
dfellis, line 1886: this one actually does need this function, though.

avm/src/vm/opcode.rs
```diff
@@ -1949,33 +1918,29 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
```
dfellis, line 1938: this one is also correct, can't switch to `register_out`

avm/src/vm/opcode.rs
```diff
@@ -1985,48 +1950,45 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      let arr = hand_mem.read_fractal(args[0]).to_vec();
-      let len = arr.len();
+      let fractal = hand_mem.read_fractal(args[0]);
+      let len = fractal.len();
       let subhandler = HandlerFragment::new(args[1], 0);
       let mut filters = Vec::new();
       for i in 0..len {
         let mut hm = hand_mem.fork();
-        hm.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
+        hm.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1965: back to `register_out` here, though

avm/src/vm/opcode.rs
```diff
@@ -1985,48 +1950,45 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      let arr = hand_mem.read_fractal(args[0]).to_vec();
-      let len = arr.len();
+      let fractal = hand_mem.read_fractal(args[0]);
+      let len = fractal.len();
       let subhandler = HandlerFragment::new(args[1], 0);
-      hand_mem.write_fractal(args[2], &Vec::new());
+      hand_mem.init_fractal(args[2]);
       for i in 0..len {
-        hand_mem.set_addr(CLOSURE_ARG_MEM_START + 1, arr[i].0, arr[i].1 as usize);
+        hand_mem.register_from_fractal(CLOSURE_ARG_MEM_START + 1, &fractal, i);
```
dfellis, line 1987: also can be replaced by `register_out`

dfellis (commented):
I think several places can be simplified, but this is great!

avm/src/vm/opcode.rs
depombo: I don't think we should make a composable function out of this. That pattern is what lends itself to have the memory abstraction to be leaked

The name is long and specific, but it is much easier to follow and imo not weird. Personally, when I saw `push_idxs` after not looking at the codebase for a while, it was not obvious what was going on. I am okay to shorten/change the name, but would prefer to keep it as-is.

avm/src/vm/opcode.rs
depombo: hmm I just realized that might be true, but they seem pretty equivalent. Your suggestion happens to work because of the way to we write into the closure argument memory space. The code that uses `register_from_fractal` to make more sense since we are iterating over the length of fractal? Your suggestion of using `register_out ` seems a bit more "magical" to follow to me or not a clear improvement at least?

avm/src/vm/memory.rs
depombo: good point!

avm/src/vm/opcode.rs
dfellis: what? But `register_out` was explicitly designed to tackle this?

* `register` -> make an alias from address A to address B
* `register_in` -> make an alias from address A to index X of address B
* `register_out` -> make an alias from index X of address A to address B

dfellis: I see it as less magical than a `register_from_fractal` that's gonna have to do some crazy magic once you add the HandlerMemory inheritance functionality to get the perf improvement we're after -- since that version is gonna have to figure out if the Fractal it was given is actually from a nested HandlerMemory and then copy it to the new HandlerMemory.

depombo: > I see it as less magical than a `register_from_fractal` that's gonna have to do some crazy magic once you add the HandlerMemory inheritance functionality to get the perf improvement we're after -- since that version is gonna have to figure out if the Fractal it was given is actually from a nested HandlerMemory and then copy it to the new HandlerMemory.

now that is a very good point that I hadn't factored in!

avm/src/vm/opcode.rs
depombo: hmmm `push_register_out` might be better

dfellis (approved):

cdmistman (approved):

PR #373, opened 2021-01-08
Easier Either, Option, and Result Destructuring RFC first draft

cdmistman:


rfcs/012 - Either, Option, and Result Destructuring.md
```diff
@@ -0,0 +1,110 @@
...
+	let okVal = myResult.getOr(-1);
+	print("should be 5: " + okVal);
+} else {
+	let errVal = myResult.getErr("expected error, but could not retrieve Error value");
+	print("unreachable, but I know that errVal is NOT 'expected error, but could not retrieve Error value', unless that's what myResult is");
+}
+```
+
+Note that, in both branches of the above `if` statement, the status (`ok` or `err`) of the `Result` is already known, but the user must then call either `getOr` or `getErr` in order to retrieve the expected value.
+(Note that `getOrDefault` is not yet implemented, but only makes this process marginally easier.)
```
dfellis, line 52: `getOrExit` could also be used and exists now, but it does imply that the user intentionally wants to crash their app if there was an error, which is semantically incorrect in this case.

rfcs/012 - Either, Option, and Result Destructuring.md
```diff
@@ -0,0 +1,110 @@
...
+if myEither.isMain() {
+	print("main rock: " + myEither); // 4
+} else {
+	print("alt rock: " + myEither); // 5
+}
+```
+
+In location 1, `myOption` should be of type `bool`.
+Meanwhile, in location 2, it might make sense to undefine the `myOption` var - this is fairly ambiguous and needs more discussion.
+Note that in location 3, `myEither` would technically be assigned the type `Either<string, void>`, but imagine it somehow assigned to a stringable value, say an `int64`.
```
dfellis, line 88: You could get the correct type signature if it came from a function that itself was dealing with some conditional, eg:

```
fn over9000(num: int64): Either<string, bool> {
  if num > 9000 {
    return main("It's over 9000!");
  } else {
    return alt(false);
  }
}
...
let myEither = over9000(9001);
```

rfcs/012 - Either, Option, and Result Destructuring.md
```diff
@@ -0,0 +1,110 @@
...
+The implementation should be relatively simple: when building the `amm` representation of the program, if the compiler detects any `if` statements (and other constructs that should have this feature too - if deemed worthy) with this pattern, then it should reassign the value at the beginning of the block.
+Preferably, this would be done with the `getR` opcode, which would allow for the value to be destructured from the `Result` efficiently.
+
+### Alternatives Considered
+
+The simplest alternative is to accept the current way of destructuring these types.
+However, this results in a lot of unnecessary overhead that any sane person will hate [(although there are fewer sane people, apparently)](https://github.com/golang/go/issues/32437).
+As such, a solution should be found that is good enough for the 99% use case (this proposal), while still providing an easy solution for every other case (the API).
+
+Another alternative would be to add ML-style pattern matching, but that might make the language quite distracted in its design and make itself inconsistent.
```
dfellis, line 100: I do want to eventually add something like that, but I don't think this issue should be gated on it, and I'm not entirely sure if my want to add pattern matching is a good idea because the more concepts in the language there are, the harder it will be to learn the language.

It feels like pattern matching could be a version 1.1 kind of thing (since it would be a backwards compatible addition of syntax).

rfcs/012 - Either, Option, and Result Destructuring.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Affected Components
+
+All code dealing with the destructuring of `Result`, `Option`, and `Either` will be broken.
```
dfellis, line 104: Nit: if destructuring within a conditional branch?

rfcs/012 - Either, Option, and Result Destructuring.md
```diff
@@ -0,0 +1,110 @@
...
+
+In location 1, `myOption` should be of type `bool`.
+Meanwhile, in location 2, it might make sense to undefine the `myOption` var - this is fairly ambiguous and needs more discussion.
+Note that in location 3, `myEither` would technically be assigned the type `Either<string, void>`, but imagine it somehow assigned to a stringable value, say an `int64`.
+In location 4, `myEither` should be of type `string`, and location 5 would be `int64` or such.
+
+The implementation should be relatively simple: when building the `amm` representation of the program, if the compiler detects any `if` statements (and other constructs that should have this feature too - if deemed worthy) with this pattern, then it should reassign the value at the beginning of the block.
+Preferably, this would be done with the `getR` opcode, which would allow for the value to be destructured from the `Result` efficiently.
+
+### Alternatives Considered
```
dfellis, line 94: We should also raise the consideration about complex conditionals that still prove that the value must be true and how we're avoiding that for now due to the far higher complexity in the compiler necessary to support that.

I would also like to add an alternative considered being early return from one of these conditional blocks implying the opposite state for the remainder of the outer scope as something considered but kept out not because it would be hard (we already isolate the remainder of the scope when there's an early return to make sure the dependency graph is executed correctly) but because it would cause hilarity when refactoring code.

dfellis: Oh, one last alternative. We could have a bare `get()` method that is only allowed within the conditional scope that has this conditional check and the user can unwrap safely without providing a default value, but it's really weird to have a method that can't always be called on the variable type it's designed for.

dfellis (approved):

rfcs/012 - Either, Option, and Result Destructuring.md
```diff
@@ -12,7 +12,7 @@ YYYY-MM-DD
 
 #### Approvers
 
-- Full Name <email@example.com>
+- David Ellis david@alantechnologies.com
```
dfellis, line 15: Please put the `<>` around the email. :) It makes the email address actually clickable with a `mailto:email@address` syntax in your browser.

depombo (approved):

PR #374, opened 2021-01-09
Fork HandlerMemory through a reference

depombo:


avm/src/vm/memory.rs
```diff
@@ -136,67 +144,120 @@ impl HandlerMemory {
...
   }
 
-  /// Takes a given address and looks up the `mems` indexes relevant to it.
-  fn addr_to_idxs(self: &HandlerMemory, addr: i64) -> (usize, usize) {
+  pub fn drop_parent(self: &mut HandlerMemory) {
+    if self.parent.is_some() {
+      let arc = self.parent.take();
+      drop(arc);
+    }
+  }
```
dfellis, line 152: What is this for? Wouldn't the parent be the one to consume the child?

depombo: Consuming the children is a mutable operation on the parent that cannot be performed on an object within an Arc. As a result, I have all the forked children drop their reference to the parent, take the parent out of the Arc and then join on its children.

avm/src/vm/memory.rs
```diff
@@ -136,67 +144,120 @@ impl HandlerMemory {
...
+      if self.parent.is_none() {
+        panic!("Memory address referenced in parent, but no parent pointer defined");
+      }
+      // fail if parent is present, but idx is not defined
+      let hm = self.parent.as_ref().unwrap();
+      let idxs = hm.addr_to_idxs_opt(addr);
+      if idxs.is_none() {
+        panic!("Memory address referenced in parent, but no indices correspond to the virtual address");
+      }
+      (idxs.unwrap(), true)
```
dfellis, line 193: This'll fail if you're multiple levels deep, eg double-nested `map`. You should just call `addr_to_idxs` on the parent hm in that case and let it handle it recursively, rather than having special case logic for the parent.

avm/src/vm/event.rs
```diff
@@ -213,9 +213,11 @@ impl HandlerFragment {
           for i in 0..instructions.len() {
             let ins = &instructions[i];
             let async_func = ins.opcode.async_func.unwrap();
+            //eprintln!("{} {} {} {}", ins.opcode._name, ins.args[0], ins.args[1], ins.args[2]);
```
depombo, line 216: leaving these here since we have similar ones in line 231 and they were very useful. Also ok removing all of them, but prefer to just keep them around

avm/src/vm/memory.rs
```diff
@@ -136,67 +149,132 @@ impl HandlerMemory {
...
-      self.mems[a][b].1
+      hm.mems[a][b].1
     }
   }
 
   /// Reads an array of data from the given address.
   pub fn read_fractal(self: &HandlerMemory, addr: i64) -> FractalMemory {
-    let (a, b) = self.addr_to_idxs(addr);
+    let ((a, b), hm_opt) = self.addr_to_idxs(addr);
+    // eprintln!("addr: {}, self?: {}, (a,b): ({},{})", addr, hm_opt.is_none(), a, b);
```
depombo, line 228: also leaving this comment here due to how useful it was

avm/src/vm/memory.rs
```diff
@@ -521,30 +628,29 @@ impl HandlerMemory {
...
+  /// newly-created data. This mechanism is faster but will keep around unreachable memory for longer.
+  /// Whether or not this is the right trade-off will have to be seen by real-world usage.
+  pub fn join(self: &mut HandlerMemory, hm: &mut HandlerMemory) {
     let s = hm.mem_addr; // The initial block that will be transferred (plus all following blocks)
     let s2 = self.mems.len(); // The new address of the initial block
     let offset = s2 - s; // Assuming it was made by `fork` this should be positive or zero
     if hm.addr.1.len() > 0 {
-      let (a, b) = hm.addr_to_idxs(CLOSURE_ARG_MEM_START); // The only address that can "escape"
+      let (a, b) = hm.addr_to_idxs_opt(CLOSURE_ARG_MEM_START).unwrap(); // The only address that can "escape"
+      // println!("a: {}, b: {}, s: {}, in_fork: {}, in_parent: {}", a, b, s, hm.is_idx_defined(a, b), self.is_idx_defined(a, b));
```
depombo, line 653: ditto

avm/src/vm/memory.rs
```diff
@@ -136,67 +149,132 @@ impl HandlerMemory {
...
+    let res = self.parent.as_ref().unwrap().hm_for_idxs(a, b);
+    return if res.is_none() { self.parent.as_ref() } else { res };
+  }
+
+  /// Takes a given address and looks up the fractal location and
+  /// `mems` indexes relevant to it if available
+  fn addr_to_idxs_opt(self: &HandlerMemory, addr: i64) -> Option<(usize, usize)> {
     return if addr >= 0 {
-     self.addr.0[addr as usize]
+      *self.addr.0.get(addr as usize).unwrap_or(&None)
```
dfellis, line 187: I'm really surprised this syntax works. If the address is defined, it'll unwrap it and return a pointer to it, but that somehow gets automatically turned into a `Some` Option? If it fails it's a pointer to the address of `None`, which kinda makes sense but still seems super indirect.

avm/src/vm/memory.rs
```diff
@@ -136,67 +149,132 @@ impl HandlerMemory {
...
+      let res = idxs.unwrap();
+      (res, self.hm_for_idxs(res.0, res.1))
+    };
   }
 
   /// Reads fixed data from a given address.
   pub fn read_fixed(self: &HandlerMemory, addr: i64) -> i64 {
-    let (a, b) = self.addr_to_idxs(addr);
+    let ((a, b), hm_opt) = self.addr_to_idxs(addr);
+    let hm = if hm_opt.is_none() { self } else { hm_opt.unwrap() };
```
dfellis, line 217: This line looks like it could just be in the `else` branch below on line 220. No need to check or unwrap the HM if the value we got back is a fixed value.

avm/src/vm/memory.rs
```diff
@@ -136,67 +149,132 @@ impl HandlerMemory {
...
   /// Provides a mutable array of data from the given address.
   fn read_mut_fractal(self: &mut HandlerMemory, addr: i64) -> &mut Vec<(usize, i64)> {
-    let (a, _) = self.addr_to_idxs(addr);
+    let ((a, _), hm_opt) = self.addr_to_idxs(addr);
+    if hm_opt.is_some() {
+      // copy necessary data from ancestor
+      let hm = Arc::clone(hm_opt.unwrap());
+      HandlerMemory::transfer(hm.as_ref(), addr, self, addr);
+      drop(hm);
+    }
```
dfellis, line 254: Nice!

avm/src/vm/memory.rs
```diff
@@ -208,15 +286,15 @@ impl HandlerMemory {
     if addr_type(addr) == NORMAL_ADDR {
       let addru = addr as usize;
       if self.addr.0.len() <= addru {
-        self.addr.0.resize(addru + 1, (std::usize::MAX, 0));
+        self.addr.0.resize(addru + 1, None);
```
dfellis, line 289: Can we confirm that this change is meant for child HMs to resize correctly?

avm/src/vm/memory.rs
```diff
@@ -260,9 +355,20 @@ impl HandlerMemory {
...
 
   /// Stores a nested fractal of data in a given address.
-  pub fn write_fractal(self: &mut HandlerMemory, addr: i64, val: &FractalMemory) {
+  pub fn write_fractal(self: &mut HandlerMemory, addr: i64, fractal: &FractalMemory) {
     let a = self.mems.len();
-    self.mems.push(val.block.clone());
+    if !fractal.belongs(self) {
+      if fractal.hm_addr.is_none() {
+        panic!("Writing a forked/read-only FractalMemory that is also deeply-nested is not possible");
+      }
```
dfellis, line 363: I feel like this one is gonna bite us at some point...

We can delay it for now, but I suspect that we'll need to clone the `Arc` here, `transfer` the memory, and then do the write.

depombo: Yes, but I didn't want to write the entire logic since it would be untested and this PR took long enough so I left a good error message for our future selves :)

avm/src/vm/memory.rs
depombo: Not sure I follow. This was actually the bug fix for `Tree` and `http` since opcodes would try to read a fractal in the child and instead of going to the parent they would look at this address. However, the empty address also works for parent HMs.

avm/src/vm/memory.rs
depombo: Yes, good point!

avm/src/vm/opcode.rs
```diff
@@ -2217,20 +2217,25 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+  io!("map", |args, hand_mem| {
     Box::pin(async move {
-      let fractal = hand_mem.read_fractal(args[0]);
+      let hand_mem_ref = Arc::new(hand_mem);
+      let fractal = hand_mem_ref.read_fractal(args[0]);
       let subhandler = HandlerFragment::new(args[1], 0);
       let mut mappers = Vec::new();
       for i in 0..fractal.len() {
-        let mut hm = hand_mem.fork();
+        let mut hm = HandlerMemory::fork(Arc::clone(&hand_mem_ref));
```
dfellis, line 2227: Unfortunate... :(

depombo: I know :(
I really tried to capture this logic within `memory.rs` but it was not possible

avm/src/vm/opcode.rs
```diff
@@ -2217,20 +2217,25 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
         hm.register_out(args[0], i, CLOSURE_ARG_MEM_START + 1);
         hm.write_fixed(CLOSURE_ARG_MEM_START + 2, i as i64);
         mappers.push(subhandler.clone().run(hm));
       }
-      let hms = join_all(mappers).await;
+      let mut hms = join_all(mappers).await;
+      for hm in &mut hms {
+        hm.drop_parent();
+      }
+      let mut hand_mem = Arc::try_unwrap(hand_mem_ref).unwrap();
```
cdmistman, line 2236: maybe do `expect` instead of `unwrap` so we get a panic message?

avm/src/vm/opcode.rs
```diff
@@ -2277,17 +2282,21 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
         hm.write_fixed(CLOSURE_ARG_MEM_START + 2, i as i64);
         runners.push(subhandler.clone().run(hm));
       }
-      join_all(runners).await;
-      hand_mem
+      let mut hms = join_all(runners).await;
+      for hm in &mut hms {
+        hm.drop_parent();
+      }
+      Arc::try_unwrap(hand_mem_ref).unwrap()
```
cdmistman, line 2299: ^

avm/src/vm/opcode.rs
```diff
@@ -2302,18 +2311,23 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+        let mut hm = HandlerMemory::fork(Arc::clone(&hand_mem_ref));
         hm.register_out(args[0], i, CLOSURE_ARG_MEM_START + 1);
         finders.push(subhandler.clone().run(hm));
       }
-      let hms = join_all(finders).await;
+      let mut hms = join_all(finders).await;
+      for hm in &mut hms {
+        hm.drop_parent();
+      }
+      let mut hand_mem = Arc::try_unwrap(hand_mem_ref).unwrap();
```
cdmistman: ^

avm/src/vm/opcode.rs
```diff
@@ -2351,25 +2365,28 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
         let val = hm.read_fixed(CLOSURE_ARG_MEM_START);
+        hm.drop_parent();
         if val == 1 {
-          hand_mem.write_fixed(args[2], 1);
-          return hand_mem;
+          ret_val = 1;
         }
       }
-      hand_mem.write_fixed(args[2], 0);
+      let mut hand_mem = Arc::try_unwrap(hand_mem_ref).unwrap();
```
cdmistman, line 2389: ^

avm/src/vm/opcode.rs
```diff
@@ -2390,25 +2407,28 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
         let val = hm.read_fixed(CLOSURE_ARG_MEM_START);
+        hm.drop_parent();
         if val == 0 {
-          hand_mem.write_fixed(args[2], 0);
-          return hand_mem;
+          ret_val = 0;
         }
       }
-      hand_mem.write_fixed(args[2], 1);
+      let mut hand_mem = Arc::try_unwrap(hand_mem_ref).unwrap();
```
cdmistman, line 2431: ^

avm/src/vm/opcode.rs
```diff
@@ -2584,26 +2604,33 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-        let hm = &hms[i];
+        let hm = &mut hms[i];
         let val = hm.read_fixed(CLOSURE_ARG_MEM_START);
+        hm.drop_parent();
         if val == 1 {
-          hand_mem.push_register_out(args[2], &fractal, i);
+          idxs.push(i);
         }
       }
+      let mut hand_mem = Arc::try_unwrap(hand_mem_ref).unwrap();
```
cdmistman, line 2630: ^

avm/src/vm/opcode.rs
```diff
@@ -3037,12 +3065,16 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
       let recurse_fn = HandlerFragment::new(slf.read_fixed(1), 0);
       let (mut seq, _) = hm.read_from_fractal(&slf, 0);
       let curr = seq.read_fixed(0);
+      let mut hand_mem;
       if curr < seq.read_fixed(1) {
         hm.write_fixed_in_fractal(&mut seq, 0, curr + 1);
         hm = recurse_fn.run(hm).await;
-        hand_mem.join(hm);
+        hm.drop_parent();
+        hand_mem = Arc::try_unwrap(hand_mem_ref).unwrap();
```
cdmistman, line 3074: ^

cdmistman (approved):

cdmistman (approved):

dfellis (approved):

avm/src/vm/memory.rs
depombo: Discussed offline, the `get` method called for `Vec` returns a `&Option<(usize, usize)>` versus before it was just `(usize, usize)`

avm/src/vm/opcode.rs
depombo: left this one out, since there was only one reference created from the arc

compiler/out.js
```diff
@@ -0,0 +1,290 @@
+const r = require('alan-js-runtime')
```
dfellis: I don't think you meant to add this to the PR...

depombo: yes, removed!

avm/src/vm/memory.rs
dfellis: It returns an `Option<&Option<(usize, usize)>>` to be precise. :)

PR #376, opened 2021-01-11
Drop 'test-server' compiler script and remove moderate severity security warning

dfellis:
This was a "feature" used to debug the compiler when built to run in the browser when Firefox refused to let `file://` loaded pages have access to some APIs for some reason. It was finally fixed a month or so ago and the particular implementation has a moderate security warning on it: https://github.com/alantech/alan/security/dependabot/compiler/yarn.lock/ecstatic/open

Since it was only a dev dependency it wasn't considered a big deal, but as the whole thing is irrelevant now, let's just remove it.


cdmistman (approved):

dfellis:
Forgot this was used in our cypress tests, going to take a different approach.

PR #378, opened 2021-01-12
Datastore Revisited RFC

dfellis:
This is now ready for review :)

### DIFF
--- a/rfcs/013 - Datastore Revisited RFC.md
+++ b/rfcs/013 - Datastore Revisited RFC.md
@@ -82,23 +82,27 @@ thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: JoinErro
 
 a hard crash of the runtime as it tries to access a number as if it was a string. With a stated goal of eliminating as many runtime errors as possible so you can be sure that your backend is going to continue running, having this unsafe casting footgun was absolutely not the intention of the `datastore` library.
 
-There are multiple ways to resolve this such that the actual datastore opcodes are left alone while the compiler guarantees that their usage is safe. I'm not actually sure which one(s) I like best, so I'm listing them all in here and we can debate the pros/cons.
+There are multiple ways to resolve this such that the actual datastore opcodes are left alone while the compiler guarantees that their usage is safe. For the sake of unblocking datastore usage as fast as possible, we're going to go through a two-step process to get it working, if a bit weird, and then using a new language feature to implement it cleanly. This does mean that the API will be changed twice, but since we are so far away from 1.0, this is considered fine for the adventurous who want to use this standard library that isn't documented publicly, yet.
 
-### Minimally-invasive with restored type introspection
+### Part 1: Minimally-invasive with restored type introspection
 
 We recently removed the ability to get the type of a variable as a string with the `type variableName` syntax, but if we restore that, it should be possible to force type safety by combining the type string with either the namespace or the key string, such that trying to get the same key name with a default value of a different type will find that it doesn't exist and return the provided default.
 
-This could provide zero actual API call changes except to the `del` function, which would need to be given an example of the type of data to delete in order to find it. This has the smallest change requirement, though it does require bringing back the string-based type manipulation code.
+This will provide zero actual API call changes except to the `has` and `del` functions, which would need to be given an example of the type of data to delete in order to find it. This has the smallest change requirement, though the API is a bit weird.
 
-### Generic Functions (in addition to the interface functions that currently exist)
+### Part 2: Generic Functions (in addition to the interface functions that currently exist)
 
-The maximum contrast to the prior is this particularly invasive change, where functions gain the ability to have one or more generic types, eg `fn foo<A>(arg1: int64, arg2: string): A {` such that they can be called with some type that is passed in during invocation but is not passed in as a variable, just the type, so `foo<bool>(1, 'bar')` would do something internally with the type.
+Functions gain the ability to have one or more generic types, eg `fn foo<A>(arg1: int64, arg2: string): A {` such that they can be called with some type that is passed in during invocation but is not passed in as a variable, just the type, so `foo<bool>(1, 'bar')` would do something internally with the type.
 
-In this case, the `namespace` function would take not only the name parameter, but the type info for the kind of variable to store in it, like `namespace<Array<int64>>('variousDigitsOfPi')`, then this type would internally be applied to the namespace through some syntax similar to the `type variableName` before so it can be stored in the datastore opcodes, but it would also be used to make sure writes and reads are sound, always providing values of the correct type.
+In this case, the `namespace` function would take not only the name parameter, but the type info for the kind of variable to store in it, like `namespace<Array<int64>>('variousDigitsOfPi')`, then this type would internally be applied to the namespace using a syntax similar to `type variableName` (perhaps as simple as `type GenericName`), so it can be stored in the datastore opcodes. This would remove the need to provide example values or type information to the `has` and `del` functions since it would already exist on the provided namespace.
 
-The biggest issue here is: it adds a new "color" of function, and there will be pressure to add something like an `fn foo<A as Orderable>(...` syntax so the generics defined for the function are also interface constrained so the function knows what to do with it besides pass it around or poke at it in some way or the other.
+It'll make the language much more complex, but the utility should surpass the complexity. This approach could be turned into the foundation for interface functions, such that `fn someFn<A as Orderable>(a: A, b: A): A` could be equivalent to `fn someFn(a: Orderable, b: Orderable): Orderable` with the latter becoming syntactic sugar. This would also allow us to remove duplicated interfaces like `anythingElse` because `fn map<A as Any, B as Any>(arr: Array<A>, mapper: fn(A): B): Array<B>` makes it clear that while `A` and `B` match the same interface, they are not the same actual type.
 
-It'll make the language much more complex, but will the utility gain be worth the complexity? It would make constructor functions much simpler, eg instead of `newHashMap(firstKey, firstVal)` we get `new<HashMap<KeyType, ValType>>()` without requiring the first key-val pair to be defined.
+It would make constructor functions much simpler, eg instead of `newHashMap(firstKey, firstVal)` we could get something like `new<HashMap<KeyType, ValType>>()` without requiring the first key-val pair to be defined.
+
+### Alternatives Considered
+
+Many alternatives were considered and rejected:
 
 ### Compiler-level hackery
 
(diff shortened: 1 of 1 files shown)


rfcs/013 - Datastore Revisited RFC.md
```diff
@@ -0,0 +1,137 @@
...
+
+There are multiple ways to resolve this such that the actual datastore opcodes are left alone while the compiler guarantees that their usage is safe. I'm not actually sure which one(s) I like best, so I'm listing them all in here and we can debate the pros/cons.
+
+### Minimally-invasive with restored type introspection
+
+We recently removed the ability to get the type of a variable as a string with the `type variableName` syntax, but if we restore that, it should be possible to force type safety by combining the type string with either the namespace or the key string, such that trying to get the same key name with a default value of a different type will find that it doesn't exist and return the provided default.
+
+This could provide zero actual API call changes except to the `del` function, which would need to be given an example of the type of data to delete in order to find it. This has the smallest change requirement, though it does require bringing back the string-based type manipulation code.
+
+### Generic Functions (in addition to the interface functions that currently exist)
```
cdmistman, line 93: I like this, it'll also remove the need for `any0`, `any1`, `any2` type variables. Something to keep in mind is how the generics would look with anonymous functions (since generics only apply to functions with params and/or a return type, we only have to worry about the `fn ([params])[: ret] { ... }` syntax here...)

```
fn foo<A>(arg1: int64, arg2: string): A { ... }
fn <A>(arg1: int64, arg2: string): A { ... }
```

Honestly, I'm not a fan of having the `<A>` right after `fn` in that scenario - my issue is with the usage of `<` and `>` without a name preceding it. It still makes sense and I'm not opposed to it, I just think if we can find a better syntax we should use it. FWIW, I do like the generics design for Go 2 (the first alternative that comes to mind), but I'm not sure how it would fit in with more Alan:
```
fn foo[A](arg1: int64, arg2: string): A { ... }
fn [A](arg1: int64, arg2: string): A { ... }
```

cdmistman: WRT `fn foo<A as Orderable>(...`, I don't think that's the worst thing. In fact, I think we can make the current interface-accepting syntax just sugar for that.
```
fn foo(arg1: Orderable, arg2: int64): string { ... }
// the above gets desugared into (and is equivalent) to this:
fn foo<A as Orderable>(arg1: A, arg2: int64): string { ... }
```

This way, none of the behavior changes, we've just introduced syntax for constraining multiple values to the same type. In fact, I think unconstrained generic type variables should have an implicit `as any`
```
fn foo<A>(arg1: int64, arg2: string): A { ... }
// the above gets desugared into (and is equivalent) to this:
fn foo<A as any>(arg1: int64, arg2: string): A { ... }
```

rfcs/013 - Datastore Revisited RFC.md
```diff
@@ -0,0 +1,137 @@
...
+
+In this case, the `namespace` function would take not only the name parameter, but the type info for the kind of variable to store in it, like `namespace<Array<int64>>('variousDigitsOfPi')`, then this type would internally be applied to the namespace through some syntax similar to the `type variableName` before so it can be stored in the datastore opcodes, but it would also be used to make sure writes and reads are sound, always providing values of the correct type.
+
+The biggest issue here is: it adds a new "color" of function, and there will be pressure to add something like an `fn foo<A as Orderable>(...` syntax so the generics defined for the function are also interface constrained so the function knows what to do with it besides pass it around or poke at it in some way or the other.
+
+It'll make the language much more complex, but will the utility gain be worth the complexity? It would make constructor functions much simpler, eg instead of `newHashMap(firstKey, firstVal)` we get `new<HashMap<KeyType, ValType>>()` without requiring the first key-val pair to be defined.
+
+### Compiler-level hackery
```
cdmistman, line 103: ew

rfcs/013 - Datastore Revisited RFC.md
```diff
@@ -0,0 +1,137 @@
...
+
+We could do it with neither of the above via effectively hackery within the compiler; the `ds*` opcodes get a special behavior that always appends the type information to the `namespace` string no matter what and we just have some magic behavior that "just works" but doesn't trust the end user to have access to that power.
+
+I'm including this for completeness, but that is the "Go Way" and in my book "The Go Way Is Stupid."
+
+### Rust-style impl-ish functionality
```
cdmistman, line 109: IMHO, this would cause the syntax of Alan to become a *lot* more noisy. I primarily see associated functions as a great way to solve namespacing functions (instead of having to do `newHashMap`), but with function shadowing I don't see the need for it (ie, `toString(5)` and `toString(true)` will always resolve correctly).

rfcs/013 - Datastore Revisited RFC.md
```diff
@@ -0,0 +1,137 @@
...
+
+This has the downside of requiring explicit boilerplate code for new types that want to be stored in datastore (and therefore can be explicitly written incorrectly and not actually do the right thing) and also breaking method syntax for these functions, but it doesn't require the magic of the Generic Functions or even the `type variableName` statement.
+
+This boilerplate could be covered up if/when we add macro functionality to the language, which would also take us further along the path towards Rust.
+
+### Namespace *is* Type
```
cdmistman, line 117: not too ecstatic about this, since it reduces functionality for users (two datastores that use the same type for different reasons)

rfcs/013 - Datastore Revisited RFC.md
```diff
@@ -0,0 +1,137 @@
...
+
+We could hide the namespace field from the end users and just make the namespace field the type. This would also require the `type variableName` syntax to be restored, but wouldn't require the special logic for the `set` and `has` functions that the mechanism maintaining the namespace would.
+
+### Stop avoiding it; change the ds* opcodes
```
cdmistman, line 121: Not a bad idea, this should probably be considered anyways, but not for the purpose of this RFC.

rfcs/013 - Datastore Revisited RFC.md
dfellis: This approach would still need the `type variableName` syntax of the first approach, too, in order to provide that data to the `ds*` opcodes, just to be clear. Only the very last option where we dig into redoing those opcodes would that potentially not be the case.

rfcs/013 - Datastore Revisited RFC.md
dfellis: Btw, what do you think is wrong/missing with those opcodes?

rfcs/013 - Datastore Revisited RFC.md
```diff
@@ -0,0 +1,137 @@
...
+thread 'tokio-runtime-worker' panicked at 'range end index 5 out of range for slice of length 0', src/vm/memory.rs:628:29
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(...)', src/vm/event.rs:260:14
+```
+
+a hard crash of the runtime as it tries to access a number as if it was a string. With a stated goal of eliminating as many runtime errors as possible so you can be sure that your backend is going to continue running, having this unsafe casting footgun was absolutely not the intention of the `datastore` library.
+
+There are multiple ways to resolve this such that the actual datastore opcodes are left alone while the compiler guarantees that their usage is safe. I'm not actually sure which one(s) I like best, so I'm listing them all in here and we can debate the pros/cons.
+
+### Minimally-invasive with restored type introspection
```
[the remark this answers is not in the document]
dfellis: The nice thing is that it's at compile time only, so there won't be any performance impact from it. But I agree that any sort of stringly-typed/branched code is generally bad, since it's very easy to accidentally not cover a relevant branch if you get an unexpected string out of it.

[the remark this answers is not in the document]
depombo: I am personally leaning towards this approach or rethinking / redoing the opcodes. Mostly because all approaches, except rethinking / redo-ing the opcodes, would still need the `type variableName` syntax and most of the other approaches are too hacky, or introduce quite a bit of noise/bloat into the syntax. I think generic functions could work, but `impl`-ish syntax feels like an overkill.

dfellis:
After keeping this in the back of my mind today to chew on while doing other things, I think I have to agree with... both of you. :)

Since we have to bring back basic (compile-time)ttype introspection anyways, we should start with the minimum impact changes where the type is added to the name space under the hood but the API is almost identical.

Aftee that, I am now convinced that we're going to be making generic functions in the future anyways for multiple reasons, so we may as well update this API a second time when that time comes, since we're still far enough away from 1.0.

This does mean that there's going to be extra work, but we unblock safe usage of datastore earlier for the adventurous and we can work on the generic function problem in parallel, since I believe that will be a multi-week effort, not 1-2 days.

@depombo @cdmistman if you are comfortable with that assessment, I will update this RFC with the two phase process.

Also, while we can continue to debate the exact syntax of the generic functions, I am leaning towards keeping <> since 1) it is consistent with generics everywhere else and 2) it's not Go-style. ;)

depombo (approved):

rfcs/013 - Datastore Revisited RFC.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Part 1: Minimally-invasive with restored type introspection
+
+We recently removed the ability to get the type of a variable as a string with the `type variableName` syntax, but if we restore that, it should be possible to force type safety by combining the type string with either the namespace or the key string, such that trying to get the same key name with a default value of a different type will find that it doesn't exist and return the provided default.
+
+This will provide zero actual API call changes except to the `has` and `del` functions, which would need to be given an example of the type of data to delete in order to find it. This has the smallest change requirement, though the API is a bit weird.
```
cdmistman, line 89: I think it's worth noting that this won't be necessary once part 2 is implemented

rfcs/013 - Datastore Revisited RFC.md
```diff
@@ -0,0 +1,142 @@
...
+
+In this case, the `namespace` function would take not only the name parameter, but the type info for the kind of variable to store in it, like `namespace<Array<int64>>('variousDigitsOfPi')`, then this type would internally be applied to the namespace using a syntax similar to `type variableName` (perhaps as simple as `type GenericName`), so it can be stored in the datastore opcodes. This would remove the need to provide example values or type information to the `has` and `del` functions since it would already exist on the provided namespace.
+
+It'll make the language much more complex, but the utility should surpass the complexity. This approach could be turned into the foundation for interface functions, such that `fn someFn<A as Orderable>(a: A, b: A): A` could be equivalent to `fn someFn(a: Orderable, b: Orderable): Orderable` with the latter becoming syntactic sugar. This would also allow us to remove duplicated interfaces like `anythingElse` because `fn map<A as Any, B as Any>(arr: Array<A>, mapper: fn(A): B): Array<B>` makes it clear that while `A` and `B` match the same interface, they are not the same actual type.
```
cdmistman, line 97: I think it should be noted that the desugaring of the second example of `someFn` wouldn't desugar into the first example. ie `someFn(a: Orderable, b: Orderable): Orderable` *shouldn't* desugar into `fn someFn<A as Orderable>(a: A, b: A): A`, but rather `fn someFn<A as Orderable, B as Orderable, R as Orderable>(a: A, b: B): R`

cdmistman (changes requested):

rfcs/013 - Datastore Revisited RFC.md
dfellis: Actually, you're wrong there, it *will* be equivalent to what I wrote in the RFC.

rfcs/013 - Datastore Revisited RFC.md
dfellis: Sure. But I'll note in in the second section.

dfellis: Wait, I did note that here: https://github.com/alantech/alan/pull/378/files#diff-5268a48223be94b011eb9b7b7042590372f2316a618e53850b6f9b0b69190635R95

rfcs/013 - Datastore Revisited RFC.md
cdmistman: I was thinking there might be a case where the desugaring as written might not be what's expected, but I can't think of one. Makes sense

cdmistman (approved):

PR #380, opened 2021-01-15
Opcode refactor

cdmistman:
Changes all of the opcode native functions to be static functions instead of closures, also changes the implementation of the opcode macros.

### DIFF
--- a/avm/src/vm/event.rs
+++ b/avm/src/vm/event.rs
@@ -191,7 +191,7 @@ impl HandlerFragment {
     loop {
       // io-bound fragment
       match &instructions[0].opcode.fun {
-        OpcodeFn::Io(async_func) => {
+        OpcodeFn::Io(io_func) => {
         if instructions.len() > 1 {
           // TODO: Revive IO parallelization, again, once AGA dependency graph *actually* works
           /*let futures: Vec<HMFuture> = instructions.iter().map(|ins| {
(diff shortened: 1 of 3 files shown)


avm/src/vm/opcode.rs
```diff
@@ -72,497 +66,544 @@ pub struct ByteOpcode {
...
   // Now insert the new name characters
-  for (i, c) in name.chars().enumerate() {
-    std::mem::replace(&mut name_vec[i], c as u8);
+  for (i, c) in name.chars().take(8).enumerate() {
+    std::mem::replace(&mut ascii_name[i], c as u8);
   }
-  let id = LittleEndian::read_i64(&name_vec);
+  let id = LittleEndian::read_i64(&ascii_name);
   return id;
 }
```
dfellis, line 94: This is the source of the issue; the `take(8)` probably did it.

dfellis: Nope, it was the `0u8` above :)

avm/src/vm/event.rs
```diff
@@ -190,7 +190,8 @@ impl HandlerFragment {
     let mut instructions = self.get_instruction_fragment();
     loop {
       // io-bound fragment
-      if !instructions[0].opcode.pred_exec {
+      match &instructions[0].opcode.fun {
```
dfellis, line 193: I've figured out why it's not working right. This `match` function matches the first instruction to determine if the fragment is an `io` or `cpu` collection, but then it uses the `async_func` or `func` from the *first instruction* on *every instruction call*, instead of using the correct function for each instruction call.

avm/src/vm/opcode.rs
```diff
@@ -47,16 +46,21 @@ static DS: Lazy<Arc<DashMap<String, HandlerMemory>>> = Lazy::new(|| {
...
-  &[i64],
-  &mut HandlerMemory,
-) -> Option<EventEmit>;
+
+/// A function to be run for an opcode.
+pub(crate) enum OpcodeFn {
+  Cpu(fn(&[i64], &mut HandlerMemory) -> Option<EventEmit>),
+  Io(fn(Vec<i64>, HandlerMemory) -> HMFuture),
+}
+
```
depombo, line 55: nice!

depombo (approved):

dfellis (approved):

PR #382, opened 2021-01-18
update node to v14

cdmistman:


dfellis (approved):
This looks good, but can we also update the release yaml file to use Node 14, too? https://github.com/alantech/alan/blob/main/.github/workflows/release.yml#L38-L41 (those are the lines for Linux, similar config further along for OSX and Windows)

dfellis (approved):

PR #383, opened 2021-01-19
Fan out parallel array operations across cores

depombo:
`eachl` is already doing the right thing here, but the rest of the linear opcodes are spinning up multiple tokio tasks when there is no explicit parallelism necessary beyond the IO parallelism provided by tokio.

### DIFF
--- a/avm/src/vm/opcode.rs
+++ b/avm/src/vm/opcode.rs
@@ -1,4 +1,5 @@
 use futures::future::{join_all, poll_fn};
+use futures::stream::{FuturesUnordered, StreamExt};
 use futures::task::{Context, Poll};
 use std::collections::HashMap;
 use std::convert::{Infallible, TryInto};
@@ -2311,7 +2312,7 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
       for i in 0..fractal.len() {
         hand_mem.register_out(args[0], i, CLOSURE_ARG_MEM_START + 1);
         hand_mem.write_fixed(CLOSURE_ARG_MEM_START + 2, i as i64);
-        hand_mem = subhandler.clone().run(hand_mem).await;
+        hand_mem = subhandler.clone().run_local(hand_mem).await;
         hand_mem.push_register(args[2], CLOSURE_ARG_MEM_START);
       }
       hand_mem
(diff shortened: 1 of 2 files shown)


cdmistman (approved):

dfellis:
> `eachl` is already doing the right thing here, but the rest of the linear opcodes are spinning up multiple tokio tasks when there is no explicit parallelism necessary beyond the IO parallelism provided by tokio.
> 
> Also keeping the number of blocking threads equal to the number of worker threads and the number of cores. The default for blocking threads is 512, but we run CPU-bound operations on them which is not their intended use case.

Only concern with the latter part there is that we do use Tokio for the HTTP server, too. I don't remember if it uses a different pool or the same shared pool. If the latter, we need to change that out so we don't create a server that can only handle `n` requests at a time. :)

depombo:
> > `eachl` is already doing the right thing here, but the rest of the linear opcodes are spinning up multiple tokio tasks when there is no explicit parallelism necessary beyond the IO parallelism provided by tokio.
> > Also keeping the number of blocking threads equal to the number of worker threads and the number of cores. The default for blocking threads is 512, but we run CPU-bound operations on them which is not their intended use case.
> 
> Only concern with the latter part there is that we do use Tokio for the HTTP server, too. I don't remember if it uses a different pool or the same shared pool. If the latter, we need to change that out so we don't create a server that can only handle `n` requests at a time. :)

Good point, let me confirm

avm/src/vm/opcode.rs
```diff
@@ -2349,14 +2350,27 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
         hm.register_out(args[0], i, CLOSURE_ARG_MEM_START + 1);
         hm.write_fixed(CLOSURE_ARG_MEM_START + 2, i as i64);
-        runners.push(subhandler.clone().run(hm));
+        let j = i / s;
+        runners[j].push(Box::pin(subhandler.clone().run_local(hm)));
       }
-      let mut hms = join_all(runners).await;
+      let mut hms: Vec<HandlerMemory> = Vec::with_capacity(l);
+      for runner in runners {
+        let mut chuncked_hms = task::spawn(async move { runner.collect().await })
```
depombo, line 2369: this code completes the parallel version of https://github.com/ledwards/advent-2020/blob/main/day01b.ln in:

```
$ time ./avm/target/release/alan run ../out1.agc 

real	1m52.170s
user	4m39.735s
sys	2m18.951s
```
However, when we try to call `join_all` on the tasks  instead of awaiting, it takes forever without the OS killing the process in my laptop

depombo: Perhaps it has to do with the fact that the opcodes are all cpu bound so the `block_in_place` call is not getting properly parallelized? Here is the alternate code calling `join_all` on the tokio `task` https://gist.github.com/depombo/9f2d8bfb53587329fad0a7db28978876

PR #388, opened 2021-01-21
Add `toSubtree` method, and make fixes to the `addChild` functions to properly handle Trees and Nodes attached to other Trees

dfellis:
This was broken out from the JSON work because it became beefy enough as it is, and it could be useful now. It also fixes a bug in the js-runtime that wasn't caught until now.

### DIFF
--- a/std/root.ln
+++ b/std/root.ln
@@ -2590,7 +2590,8 @@ export fn toSubtree(n: Node<any>): Tree<any> {
   let outTree = newTree(val);
   let children = n.tree.children[n.id].getOr(new Array<int64> []);
   let parentIds = [0].repeat(children.length().sub(1).getOr(0));
-  seqdo(newseq(pow(2, 64).getOr(0)), fn (): bool {
+  //seqdo(newseq(pow(2, 64).getOr(0)), fn (): bool {
+  seqdo(newseq(100), fn (): bool {
     const childId = children.pop();
     const parentId = parentIds.pop();
     const childVal = n.tree.vals[childId].getR();
--- a/std/root.ln
+++ b/std/root.ln
@@ -2591,8 +2591,7 @@ export fn toSubtree(n: Node<any>): Tree<any> {
   let children = n.tree.children[n.id].getOr(new Array<int64> []);
   if children.length().gt(0) {
     let parentIds = [0].repeat(children.length());
-    //seqdo(newseq(pow(2, 64).getOr(0)), fn (): bool {
-    seqdo(newseq(100), fn (): bool {
+    seqdo(newseq(pow(2, 64).getOr(0)), fn (): bool {
       const childId = children.pop();
       const parentId = parentIds.pop();
       const childVal = n.tree.vals[childId].getR();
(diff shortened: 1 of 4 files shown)


std/root.ln
```diff
@@ -2583,26 +2583,84 @@ export fn prune(n: Node<any>): Tree<any> {
...
 
+// Takes the given node and copies that node and all of its children into a new tree with the
+// current node as the root. Unfortunately a sequential algorithm right now.
+export fn toSubtree(n: Node<any>): Tree<any> {
+  const val = n.tree.vals[n.id].getR();
+  let outTree = newTree(val);
+  let children = n.tree.children[n.id].getOr(new Array<int64> []);
+  if children.length().gt(0) {
+    let parentIds = [0].repeat(children.length());
+    //seqdo(newseq(pow(2, 64).getOr(0)), fn (): bool {
```
dfellis, line 2594: Oh, I was debugging an infinite loop at one point, but I should restore this instead of limiting subtrees to 100 nodes.

cdmistman (approved):

std/root.ln
```diff
@@ -2583,26 +2583,83 @@ export fn prune(n: Node<any>): Tree<any> {
...
+  n.tree.children[n.id].getOr(new Array<int64> []).push(childId);
   return new Node<any> {
     id: childId,
     tree: n.tree,
   };
 }
 
+// Attaches the specified Tree to the specified Node. Can be done with simple maps and concats. This
+// returns the new child node corresponding to the root node of the tree.
+export fn addChild(n: Node<any>, val: Tree<any>): Node<any> {
```
depombo, line 2629: what happens if `val` and `n` are already in the same tree?

dfellis: Since we map the nodes of the tree into new arrays before concatenating them with the original arrays and replacing them, it *should* work, actually.

dfellis: Do I need to write a test for that? :/

depombo: > Since we map the nodes of the tree into new arrays before concatenating them with the original arrays and replacing them, it _should_ work, actually.

Wait, what I meant is that if they are in the same tree already and we create another relationship between them we would create a cycle in the tree

dfellis: Oh, no, when a tree is attached to another tree, we're basically copying everything over from the one tree to the other and restitching the parent/child relationships, so there's no cycle here.

We effectively take the first tree, sorta clone it, then attach that clone as a child of whatever node we attached it to. Eg:

```
    A          A                                                                                       
   / \   =>   / \                                                                                      
  B   C      B   C                                                                                     
                 |                                                                                     
                 A'                                                                                    
                / \                                                                                    
               B'  C'
```

depombo: > Oh, no, when a tree is attached to another tree, we're basically copying everything over from the one tree to the other and restitching the parent/child relationships, so there's no cycle here.
> 
> We effectively take the first tree, sorta clone it, then attach that clone as a child of whatever node we attached it to. Eg:
> 
> ```
>     A          A                                                                                       
>    / \   =>   / \                                                                                      
>   B   C      B   C                                                                                     
>                  |                                                                                     
>                  A'                                                                                    
>                 / \                                                                                    
>                B'  C'
> ```

ohh got it, can we include your ASCII Art prowess in the method description comment? :)

dfellis: You want this diagram added to the comments?

depombo: Yes!

depombo (approved):

PR #391, opened 2021-01-22
Arc-ifying HandlerMemory

cdmistman:
In the current implementation, every `HandlerMemory` might get heap-allocated and `memcpy`'d multiple times, which can balloon runtime when many `io` opcodes are called. Instead, we should try to allocate every `HandlerMemory` in an `Arc`.


avm/err.txt
```diff
@@ -0,0 +1,266 @@
+warning: unused import: `futures::future::join_all`
```
depombo, line 1: I'm guessing this one sneaked in :)

avm/src/vm/opcode.rs
```diff
@@ -2682,12 +2677,12 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
         hm.register_out(args[0], i, CLOSURE_ARG_MEM_START + 1);
         filters.push(subhandler.clone().run(hm));
       }
-      let mut hms = join_all(filters).await;
+      let hms = join_all(filters).await;
       let mut idxs = vec![];
-      for i in 0..len {
-        let hm = &mut hms[i];
+      for (i, hm) in hms.into_iter().enumerate() {
```
depombo, line 2680: nice

avm/src/vm/memory.rs
```diff
@@ -154,11 +153,16 @@ impl HandlerMemory {
...
-  pub fn drop_parent(self: &mut HandlerMemory) {
-    if self.parent.is_some() {
-      let arc = self.parent.take();
-      drop(arc);
+  pub fn drop_parent(mut self: Arc<HandlerMemory>) -> Arc<HandlerMemory> {
+    let hm = Arc::get_mut(&mut self).expect("unable to drop parent HM: dangling pointer");
+    if let Some(parent) = hm.parent.take() {
+      drop(parent);
     }
+    self
```
depombo, line 160: I don't think we need to return `self` here, only in `drop_parent_async`

avm/src/vm/memory.rs
```diff
@@ -154,11 +153,16 @@ impl HandlerMemory {
...
+  }
+
+  pub(crate) async fn _drop_parent(self: Arc<HandlerMemory>) -> Arc<HandlerMemory> {
```
depombo, line 163: can we call this `drop_parent_async` like I had mentioned during the pair program?

avm/src/vm/memory.rs
```diff
@@ -190,11 +192,7 @@ impl HandlerMemory {
     return if addr >= 0 {
       *self.addr.0.get(addr as usize).unwrap_or(&None)
     } else if addr <= CLOSURE_ARG_MEM_END {
-      *self
-        .addr
-        .1
-        .get((addr - CLOSURE_ARG_MEM_START) as usize)
-        .unwrap_or(&None)
+      *self.addr.1.get((addr - CLOSURE_ARG_MEM_START) as usize).unwrap_or(&None)
```
depombo, line 194: maybe we need to run the rust fmt command?

avm/src/vm/memory.rs
```diff
@@ -701,20 +675,21 @@ impl HandlerMemory {
...
   /// Returns a new HandlerMemory with a read-only reference to HandlerMemory as parent
-  pub fn fork(parent: Arc<HandlerMemory>) -> HandlerMemory {
+  pub fn fork(parent: Arc<HandlerMemory>) -> Arc<HandlerMemory> {
     let s = parent.mems.len();
     let mut hm = HandlerMemory::new(None, 1);
-    hm.parent = Some(parent);
-    hm.mems.resize(s + 1, Vec::new());
-    hm.mem_addr = s;
-    hm.args_addr = s;
+    let handlermem = Arc::get_mut(&mut hm).expect("somehow a dangling pointer for a brand new HM?");
```
depombo, line 687: nit just `handmem`?

avm/src/vm/opcode.rs
```diff
@@ -2682,12 +2677,12 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+        // let hm = &mut hms[i];
         let val = hm.read_fixed(CLOSURE_ARG_MEM_START);
-        hm.drop_parent();
+        hm.drop_parent(); // this drops `hm`
```
depombo, line 2683: you mean `hm` parrent which seems self explanatory?

avm/src/vm/run.rs
```diff
@@ -38,14 +38,14 @@ impl VM {
...
-  fn sched_event(self: &mut VM, event: EventEmit) {
+  fn sched_event(self: &mut VM, mut event: EventEmit) {
     // schedule 1st fragment of each handler of this event
     let handlers = Program::global().event_handlers.get(&event.id).unwrap();
     for (i, hand) in handlers.iter().enumerate() {
       // first fragment of this handler
       let frag = HandlerFragment::new(event.id, i);
       // memory frag representing the memory for each handler call
-      let hand_mem = HandlerMemory::new(event.payload.clone(), hand.mem_req);
+      let hand_mem = HandlerMemory::new(event.payload.take(), hand.mem_req);
```
depombo, line 48: I'm confused because [take](https://doc.rust-lang.org/std/option/enum.Option.html#method.take) clears the event payload and sets it to `None` so the next handler that comes along

avm/temp.aga
```diff
@@ -0,0 +1,108 @@
+Alan Graphcode Assembler v0.0.1
```
depombo, line 1: another one that sneaks through

avm/src/vm/memory.rs
```diff
@@ -204,174 +202,143 @@ impl HandlerMemory {
...
+        // The indexes point to fixed data
+        (vec![hm.mems[a][b_usize].clone()], false)
+      } else { // b_usize > std::usize::MAX
+        // The indexes point at nested data
+        (hm.mems[a].clone(), true)
+      };
+      (FractalMemory {
+        hm_addr: None,
+        block,
+        hm_id: Arc::as_ptr(hm) as usize,
```
depombo, line 333: I remember you pointed out this doesn't actually work like I thought it did when I implemented. Does it work as a unique id with Arcs? Maybe we should remove all together

avm/err.txt
cdmistman: heh, i forgot to get rid of the cargo warnings and the temp files. fixed!

avm/src/vm/opcode.rs
cdmistman: with the change I made to `drop_parent`'s `self` parameter, it also drops `hm`. I can revert `drop_parent` back though

avm/src/vm/memory.rs
cdmistman: I'm doing the same thing as you did before, but the difference is that the references you were using before were references to the object on the stack. I'm not 100% sure how often Rust allows for moving values on the stack, but it's entirely possible for the `HandlerMemory` to not be at the same memory address between the time the `FractalMemory` is created and the time its `hm_id` field is used

avm/src/vm/run.rs
cdmistman: ah... this will take a bit more to figure out then, since using `clone` increments the refcount of the payload...

depombo: oof that sucks. I think subsequent events are payload-less. However we can verify with a bdd test that has a custom event with a payload and it register 2 handlers to it and checks that the payload is present

bdd/spec/001_event_spec.sh
```diff
@@ -82,4 +82,43 @@ Describe "Events"
       The status should eq "0"
     End
   End
+
+  Describe "converting integers to strings when received as an event"
```
depombo, line 86: nit `integers from global memory` since we verified the following works:
```
from @std/app import start, print, exit

event aNumber: int64

on aNumber fn(num: int64) {
  print(num.toString());
}

on start {
  const a = 5 + 5;
  emit aNumber a.getOr(0);
}
```

depombo (approved):

bdd/spec/001_event_spec.sh
cdmistman: oh right! fixing

avm/src/vm/opcode.rs
cdmistman: We decided not to revert `drop_parent` so that it matches the signature of the new `drop_parent_async`

avm/src/vm/memory.rs
cdmistman: see https://github.com/alantech/alan/pull/391#discussion_r564812823

avm/src/vm/memory.rs
```diff
@@ -204,174 +204,157 @@ impl HandlerMemory {
...
-        hm_id: hm as *const HandlerMemory as usize,
-      }
-    } else {
-      FractalMemory {
-        hm_addr: Some(addr),
-        block: hm.mems[a][..].to_vec(),
-        hm_id: hm as *const HandlerMemory as usize,
-      }
+    // Special behavior to read strings out of global memory
+    let start = if addr_type(addr) == GMEM_ADDR { b } else { 0 };
```
dfellis, line 251: Clever

avm/src/vm/memory.rs
```diff
@@ -204,174 +204,157 @@ impl HandlerMemory {
...
-        },
-        false,
-      )
     } else {
-      // The indexes point at nested data
+      let (block, is_fractal) = if b_usize < std::usize::MAX {
+        // The indexes point to fixed data
+        (vec![hm.mems[a][b_usize].clone()], false)
+      } else {
+        // b_usize > std::usize::MAX
```
dfellis, line 340: Wouldn't this be `b_usize == std::usize::MAX`? You can't exactly exceed it.

avm/src/vm/memory.rs
```diff
@@ -640,15 +638,32 @@ impl HandlerMemory {
           }
         }
       }
-      dest.mems.append(&mut orig_arr_copies);
+      if cfg!(debug_assertions) {
```
dfellis, line 641: What is this?

avm/src/vm/opcode.rs
```diff
@@ -2592,58 +2583,60 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+  //     // let hms = join_all(reducers).await;
+  //     // hand_mem.init_fractal(args[2]);
+  //     // for i in 0..n {
+  //     //   let hm = hms[i].as_ref().unwrap();
+  //     //   HandlerMemory::transfer(&hm, 0, &mut hand_mem, CLOSURE_ARG_MEM_START);
+  //     //   hand_mem.push_register(args[2], CLOSURE_ARG_MEM_START);
+  //     // }
+  //     // hand_mem
+  //   // })
+  // });
```
dfellis, line 2639: Why the double-commenting here?

avm/src/vm/opcode.rs
```diff
@@ -2682,12 +2675,11 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
       }
-      let mut hms = join_all(filters).await;
+      let hms = join_all(filters).await;
       let mut idxs = vec![];
-      for i in 0..len {
-        let hm = &mut hms[i];
+      for (i, hm) in hms.into_iter().enumerate() {
         let val = hm.read_fixed(CLOSURE_ARG_MEM_START);
-        hm.drop_parent();
+        hm.drop_parent(); // this drops `hm`
```
dfellis, line 2682: Wait, what? Why does that happen? Why can't we just call `drop(hm);` if that's the case?

bdd/spec/001_event_spec.sh
```diff
@@ -82,4 +82,43 @@ Describe "Events"
...
+    BeforeAll before
+
+    after() {
+      cleanTemp
+    }
+    AfterAll after
+    INTOUTPUT="I got a number! 5"
+
+    It "runs js"
+      Pending integer-event-arg-from-global-memory
```
dfellis, line 113: This could also be avoided by not having the number in constant memory, which is probably why `looper` works. I bet `emit aNumber 5 | 0;` would succeed.

depombo: https://github.com/alantech/alan/pull/391#discussion_r564804326

avm/src/vm/memory.rs
cdmistman: evaluates to `true` on debug builds and `false` on release builds

avm/src/vm/opcode.rs
cdmistman: heh, that was from when i was trying to fix it... fixing

avm/src/vm/opcode.rs
cdmistman: https://github.com/alantech/alan/pull/391#discussion_r564812823

cdmistman: oh also, `hm` is dropped on line 2686 anyways, so it doesn't actually impact the loop, I just left this here so people don't get confused if they can't access `hm` on the next line or something

dfellis: Is that because it now returns `hm`, so if you don't grab its return value its effectively dropped?

It may make sense to rename these two functions, since it doesn't just drop the parent reference anymore?

dfellis (approved):

avm/src/vm/opcode.rs
cdmistman: > Is that because it now returns `hm`, so if you don't grab its return value its effectively dropped?

Yes

> It may make sense to rename these two functions, since it doesn't just drop the parent reference anymore?

I mean, it still does. Effectively, the only difference between `drop_parent` and `Drop::<HandlerMemory>::drop` is that `drop_parent` *only* drops the parent, and still allows you to access the `HandlerMemory` afterwards.

PR #393, opened 2021-01-26
alan deploy frontend

depombo:


avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,159 @@
...
+
+pub fn kill(app_id: &str) {
+  let config = get_config().unwrap();
+  let body = json!({
+    "aws": {
+      "accessKeyId": config.access_key_id,
+      "secretAccessKey": config.secret_access_key,
+    },
+    "appId": app_id,
+  });
```
dfellis, line 73: Why the flip-flopping between camelCase and snake_case? While Rust uses snake_case, Alan uses camelCase and so does AWS, it seems, so I'd normalize it all to camelCase and do something to silence the warnings from Rust?

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,159 @@
...
+  if res.is_ok() {
+    println!("Killing {} ...", app_id);
+  } else {
+    println!("Killing {} failed", app_id);
+  }
+}
+
+pub fn new() {
+  // let dest_file = "temp.agz";
+  // let status_code = compile(&ln_file, &dest_file, true);
```
dfellis, line 84: I understand that this is commended out code, but it would probably be a good idea to check if the user provided an `agz` file themselves before trying to compile it?

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+#[derive(Deserialize, Debug)]
+struct VM {
+  id: String,
+  url: String,
+  status: String,
+}
+
+const URL: &str = "http://localhost:8000";
```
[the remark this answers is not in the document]
depombo: Yes

avm/src/main.rs
```diff
@@ -33,13 +34,25 @@ fn main() {
...
     .subcommand(SubCommand::with_name("compile")
-      .about("Compiles the given source file (.ln, .amm, .aga) to a new output file (.amm, .aga, .agc, .js)")
+      .about("Compiles the given source file (.ln, .amm, .aga) to a new output file (.amm, .aga, .agz, .agc, .js)")
       .arg_from_usage("<INPUT> 'Specifies the input file to load'")
       .arg_from_usage("<OUTPUT> 'Specifies the output file to generate'"))
     .subcommand(SubCommand::with_name("install")
       .about("Install '/dependencies' from '.dependencies.ln'")
       .version(crate_version!()))
-    .arg_from_usage("[SOURCE] 'Specifies a source ln file to compile and run'")
+    .subcommand(SubCommand::with_name("deploy-new")
```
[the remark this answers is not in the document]
depombo: Yeah, I ended up just going with `alan deploy-new` in the spirit of the MVP because setting up nested subcommands for `alan deploy new` was a huge pain

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,159 @@
...
+      "{} does not exist. Please define one.",
+      file_name
+    );
+    std::process::exit(1);
+  }
+  let file = File::open(path)?;
+  let reader = BufReader::new(file);
+  let c = from_reader(reader)?;
+  Ok(c)
+}
```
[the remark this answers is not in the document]
depombo: Good point

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,159 @@
...
+
+fn post(url: String, body: Value) -> Result<String, Box<dyn Error>> {
+  let client = Client::builder().build::<_, Body>(hyper_tls::HttpsConnector::new());
+  let rt  = Runtime::new().unwrap();
+  rt.block_on(async move {
+    let req = Request::post(url)
+      .header("Content-Type", "application/json")
+      .body(body.to_string().into())?;
+    let mut resp = client.request(req).await?;
+    let data = hyper::body::to_bytes(resp.body_mut()).await?;
```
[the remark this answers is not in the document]
depombo: Unwrap an `Option` or `Result` and return `Error` if it fails so it is only allowed if used within a function that returns a `Result` itself

avm/src/vm/deploy.rs
depombo: yes, I changed this locally but hadn't pushed it

avm/src/vm/deploy.rs
depombo: Good point!

avm/src/main.rs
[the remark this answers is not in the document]
depombo: I mostly just prefer `alan deploy-new` over `alan deploy new` after actually using it, but I could use [this](https://docs.rs/clap/2.33.3/clap/struct.SubCommand.html#method.with_name) as @cdmistman pointed out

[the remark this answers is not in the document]
dfellis: So having them be separate names with this sort of grouping is kinda conventional across most CLI apps, and it's also nicer if/when we add autocomplete functionality for alan in bash and other shells so you can do `alan deploy<tab><tab>` and see the deploy commands instead of needing to do `alan<tab><tab>` and seeing *everything* including stuff you don't care about. Less of an issue now, but when we sprout package management commands on the alan command it'll start getting hard to follow.

avm/src/main.rs
```diff
@@ -26,19 +27,31 @@ fn main() {
...
+      .arg_from_usage("<AGZ_FILE> 'Specifies the .agz file to deploy'"))
+    .subcommand(SubCommand::with_name("deploy-kill")
+      .about("Kills an app with the provided id in the cloud provider described in the deploy config at ~/.alan/deploy.json")
+      .arg_from_usage("<APP_ID> 'Specifies the alan app to kill'"))
+    .subcommand(SubCommand::with_name("deploy-status")
+      .about("Displays all the apps deployed in the cloud provider described in the deploy config at ~/.alan/deploy.json"))
+    .subcommand(SubCommand::with_name("deploy-upgrade")
+      .about("Deploys an .agz file to an existing app in the cloud provider described in the deploy config at ~/.alan/deploy.json")
+      .arg_from_usage("<APP_ID> 'Specifies the alan app to upgrade'")
+      .arg_from_usage("<AGZ_FILE> 'Specifies a source ln file to deploy'"))
```
dfellis, line 54: Should it be a source ln file that it then compiles, or an already compiled agz file as implied by the argument name?

depombo: ah good catch

depombo: It should match the rest

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,185 @@
...
+use std::fmt::Display;
+use std::fs::{File, read};
+use std::io::BufReader;
+use std::path::Path;
+
+use ascii_table::{AsciiTable, Column, Align};
+use base64;
+use tokio::runtime::Runtime;
+
+const URL: &str = "https://alan-deploy-prod.herokuapp.com";
```
dfellis, line 15: Should we CNAME this to `deploy.alantechnologies.com`?

dfellis: Another thing; should we shove a `/v1/` into the URL paths so we can change the APIs in the future without breaking older versions of `alan`. If we get upstreamed into something like Debian there's going to be significant usage of an earlier version of alan just like we ourselves were using an earlier version of Rust for a while since 1.41 is what Debian still ships.

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,185 @@
...
+  \"aws\": {
+      \"region\": \"string\",
+      \"credentials\": {
+        \"accessKeyId\": \"string\",
+        \"secretAccessKey\": \"string\",
+      }
+  }
+}";
+const HOW_TO_AWS: &str = "
+To create an AWS access key follow this tutorial:\nhttps://aws.amazon.com/premiumsupport/knowledge-center/create-access-key/\n
```
dfellis, line 53: Nit: double `\n` before and after the URL so it's got some whitespace surrounding it and therefore a bit easier to copy-paste or cmd/ctrl-click in the terminal?

avm/src/vm/deploy.rs
depombo: > Another thing; should we shove a `/v1/` into the URL paths so we can change the APIs in the future without breaking older versions of `alan`. If we get upstreamed into something like Debian there's going to be significant usage of an earlier version of alan just like we ourselves were using an earlier version of Rust for a while since 1.41 is what Debian still ships.

Great point! I decouple the CLI from the rest api internals as much as possible, but let's do this!

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,185 @@
...
+use std::fmt::Display;
+use std::fs::{File, read};
+use std::io::BufReader;
+use std::path::Path;
+
+use ascii_table::{AsciiTable, Column, Align};
+use base64;
+use tokio::runtime::Runtime;
+
+const URL: &str = "http://deploy.alantechnologies.com";
```
dfellis, line 15: Next comment; we need to make sure this is https if we're transmitting AWS credentials through this. Getting a cert shouldn't be too hard with Let's Encrypt or whatever Heroku has built-in.

depombo: Oh it works with https. That's a typo. Will update tomorrow.

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,185 @@
...
+Then enable programmatic access for the IAM user, and attach the built-in 'AdministratorAccess' policy to your IAM user.
+";
+
+fn get_config() -> Config {
+  let home = std::env::var("HOME").unwrap();
+  let file_name = &format!("{}/{}", home, CONFIG_NAME);
+  let path = Path::new(file_name);
+  let file = File::open(path);
+  if file.is_err() {
+    println!("No deploy config at {}", file_name);
```
cdmistman, line 63: Technically, the config could still be there, but we don't have permissions or there could be some other reason the file failed to open. Maybe print the `Error` instead?

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,185 @@
...
+    println!("{}", CONFIG_SCHEMA);
+    println!("{}", HOW_TO_AWS);
+    std::process::exit(1);
+  }
+  let reader = BufReader::new(file.unwrap());
+  let config = from_reader(reader);
+  if config.is_err() {
+    println!("Invalid deployment config. Error: {}", config.unwrap_err());
```
cdmistman, line 71: maybe grab the `Error` from `if let` instead?

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,185 @@
...
+    "deployConfig": get_config(),
+    "appId": app_id,
+  });
+  let res = post(format!("{}/v1/kill", URL), body);
+  if res.is_ok() {
+    println!("Killing app with id {} if it exists...\n", app_id);
+    status();
+  } else {
+    println!("Killing app with id {} failed.", app_id);
+  }
```
cdmistman, line 114: maybe `match` the result of the `post` so we can better report the error?

avm/src/vm/deploy.rs
```diff
@@ -0,0 +1,185 @@
...
+    "deployConfig": get_config(),
+    "agzB64": app_str,
+  });
+  let body = json!(body);
+  let res = post(format!("{}/v1/new", URL), body);
+  if res.is_ok() {
+    println!("Creating new app with id {}...\n", res.unwrap());
+    status();
+  } else {
+    println!("Failed to create a new app.");
```
cdmistman, line 129: nit: some messages end with a period and some don't

avm/src/main.rs
```diff
@@ -1,71 +1,127 @@
...
-      compile_and_run(source_file);
-    } else {
-      println!(
-        "{} does not exist. Dependencies can only be installed for {}",
-        source_file, source_file
+  match matches.subcommand() {
+    ("run",  Some(matches)) => {
+      let agc_file = matches.value_of("FILE").unwrap();
+      let fp = &format!(
+        "{:}/{:}",
```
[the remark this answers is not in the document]
depombo: tbh I am not sure. This shows up as green because I restructured things to use a match statement, but I am not changing it

avm/src/vm/deploy.rs
depombo: Good point

avm/src/vm/deploy.rs
depombo: the correct way would be to print the error from the REST API, but AWS never fails the terminate instance request so this branch should never fail/run. Down the line, the endpoint could do another request to the AWS sdk and check if it worked, but this is okay as-is for now

avm/src/main.rs
```diff
@@ -1,71 +1,127 @@
...
-    std::process::exit(compile(&source_file, &dest_file, false));
-  } else if let Some(_matches) = matches.subcommand_matches("install") {
-    let source_file = ".dependencies.ln";
-    if Path::new(source_file).exists() {
-      compile_and_run(source_file);
-    } else {
-      println!(
-        "{} does not exist. Dependencies can only be installed for {}",
-        source_file, source_file
+  match matches.subcommand() {
```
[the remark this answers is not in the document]
depombo: Yeah, I was inspired by you on this one :)

avm/src/main.rs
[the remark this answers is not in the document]
dfellis: It uses a different formatter internally. Rust has `{}`, `{:}` and `{:?}` as different kinds of formatters by default, as I recall. I don't remember why I went with this one, though.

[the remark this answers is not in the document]
dfellis: I can't find the bare `:` formatter in the documentation, though. Perhaps it's just a silly alias for `{}` that I didn't quite understand when we were first starting to use Rust.

dfellis (approved):

cdmistman (approved):

avm/src/vm/deploy.rs
```diff
@@ -107,13 +107,8 @@ pub fn kill(app_id: &str) {
   });
   let resp = post(format!("{}/v1/kill", URL), body);
   match resp {
-    Ok(_) => {
-      println!("Killing app with id {} if it exists...\n", app_id);
-      status();
```
dfellis: You don't need these status calls anymore?

depombo: They should have never been there in part because AWS doesn't pick up the change kickly enough so it wasn't showing the proper status of the kill-ed / new one

PR #395, opened 2021-01-27
@std/json representation and serialization on js-runtime

dfellis:
This is not the final way I'd want to have to use `@std/json` -- you shouldn't have to import the nested types like it currently requires if you're only pulling in `toJSON`. I also hope to get the `json` interface done so you can automatically pull in a lot of related functions without a massive import line, but those can get fixed later.

### DIFF
--- a/bdd/spec/032_json_spec.sh
+++ b/bdd/spec/032_json_spec.sh
@@ -5,7 +5,7 @@ Describe "JSON"
     before() {
       sourceToAll "
         from @std/app import start, print, exit
-        from @std/json import JSON, toJSON, toString, JSONBase, JSONNode, IsObject
+        from @std/json import JSON, toJSON, toString, JSONBase, JSONNode, IsObject, Null
 
         on start {
           1.0.toJSON().print();
--- a/bdd/spec/032_json_spec.sh
+++ b/bdd/spec/032_json_spec.sh
@@ -25,10 +25,10 @@ Describe "JSON"
     }
     AfterAll after
 
-    BASICOUTPUT="1.0
+    BASICOUTPUT="1
 true
 \"Hello, JSON!\"
-[1.0, 2.0, 5.0]
+[1, 2, 5]
 null"
 
     It "runs js"
@@ -37,8 +37,9 @@ null"
     End
 
     It "runs agc"
-      When run test_agc
-      The output should eq "$BASICOUTPUT"
+      Pending debugging-whats-wrong-in-compiler-or-avm
+      #When run test_agc
+      #The output should eq "$BASICOUTPUT"
     End
   End
 End
--- a/bdd/spec/032_json_spec.sh
+++ b/bdd/spec/032_json_spec.sh
@@ -1,0 +1,44 @@
+Include build_tools.sh
+
+Describe "JSON"
+  Describe "basic construction and printing"
+    before() {
+      sourceToAll "
+        from @std/app import start, print, exit
+        from @std/json import json
+
+        on start {
+          1.0.toJSON().print();
+          true.toJSON().print();
+          'Hello, JSON!'.toJSON().print();
+          [1.0, 2.0, 5.0].toJSON().print();
+          toJSON().print();
+
+          emit exit 0;
+        }
+      "
+    }
+    BeforeAll before
+
+    after() {
+      cleanTemp
+    }
+    AfterAll after
+
+    BASICOUTPUT="1.0
+true
+\"Hello, JSON!\"
+[1.0, 2.0, 5.0]
+null"
+
+    It "runs js"
+      When run test_js
+      The output should eq "$BASICOUTPUT"
+    End
+
+    It "runs agc"
+      When run test_agc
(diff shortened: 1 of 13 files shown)


compiler/src/index.ts
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env node
 
+// Make sure stack traces are *full*
```
dfellis, line 3: Oh, I need to remove this.

compiler/src/lntoamm/opcodes.ts
```diff
@@ -240,9 +241,24 @@ const addopcodes = (opcodes: object) => {
...
+                      const generics = [
+                        fulltypenameAst.get('opttypegenerics').get('generics').get('fulltypename').t
+                      ]
+                      fulltypenameAst.get('opttypegenerics').get('generics').get('cdr').getAll()
+                        .forEach(r => {
+                          generics.push(r.get('fulltypename').t)
+                        })
+                      return baseType.solidify(generics, scope)
+                    })() : scope.deepGet(arrayInnerTypeStr) as Type
+                  /*const arrayInnerType = scope.deepGet(
```
dfellis, line 259: Need to delete this line too

compiler/src/lntoamm/opcodes.ts
```diff
@@ -421,9 +437,24 @@ const addopcodes = (opcodes: object) => {
...
+                        const generics = [
+                          fulltypenameAst.get('opttypegenerics').get('generics').get('fulltypename').t
+                        ]
+                        fulltypenameAst.get('opttypegenerics').get('generics').get('cdr').getAll()
+                          .forEach(r => {
+                            generics.push(r.get('fulltypename').t)
+                          })
+                        return baseType.solidify(generics, scope)
+                      })() : scope.deepGet(arrayInnerTypeStr) as Type
+                      /*const arrayInnerType = scope.deepGet(
```
dfellis, line 455: Same, note to delete this

std/json.ln
```diff
@@ -0,0 +1,633 @@
+/* @std/json - A library for parsing, representing, and serializing to and from JSON
+ * This representation can handle every JSON type, and if you squint at it, it cannot represent an
```
depombo, line 2: 😂

std/json.ln
```diff
@@ -0,0 +1,633 @@
...
+ * invalid value. This is not entirely true without a small liberty being taken. If a JSONNode is an
+ * object, its children *should* be `KeyVal<string, JSONBase>` type, but if they're just `JSONBase`
+ * you could also just use the `toString` representation of the value as its key and the value as
+ * the value. Similarly, you should not ever have children attached to a `JSONBase`, but you can
+ * just ignore those children.
+ */
+export type IsObject {
+  isObject: bool,
+}
+// TODO: Fix issues with using `void` as a value in the compiler
```
depombo, line 12: we should track this with a task if one doesn't exist

bdd/spec/032_json_spec.sh
```diff
@@ -0,0 +1,45 @@
...
+
+Describe "JSON"
+  Describe "basic construction and printing"
+    before() {
+      sourceToAll "
+        from @std/app import start, print, exit
+        from @std/json import JSON, toJSON, toString, JSONBase, JSONNode, IsObject, Null
+
+        on start {
+          1.0.toJSON().print();
```
depombo, line 11: This reads a bit odd, like a tuple in rust.

bdd/spec/032_json_spec.sh
```diff
@@ -0,0 +1,45 @@
...
+          true.toJSON().print();
+          'Hello, JSON!'.toJSON().print();
+          [1.0, 2.0, 5.0].toJSON().print();
+          toJSON().print();
```
depombo, line 15: I really like that all base types implement this interface

depombo (approved):

bdd/spec/032_json_spec.sh
dfellis: *Almost* all base types. I haven't implemented it for the `int` types yet, though it should be trivial.

bdd/spec/032_json_spec.sh
dfellis: Oh, but it's the constant `1.0` having the `toJSON` method called on it. :)

That you've got it a bit confused with tuples is part why I haven't implemented tuples; keeping it unambiguous in the grammar is complex for multiple reasons (tuple definition, tuple type definition, and even tuple accessor syntax could be confusing).

std/json.ln
dfellis: I plan to make two tasks after this, at minimum. Not just `void` as value not working, but also the variable aliasing bug I ran across.

std/json.ln
dfellis: squint

bdd/spec/032_json_spec.sh
depombo: Yep, I agree. No tuples!

cdmistman (approved):

PR #397, opened 2021-02-02
allow multiple aws configurations

depombo:


avm/src/main.rs
```diff
@@ -44,8 +44,9 @@ fn main() {
       .about("Deploy .agz files to the cloud provider described in the deploy config at ~/.alan/deploy.json")
       .setting(AppSettings::SubcommandRequiredElseHelp)
       .subcommand(SubCommand::with_name("new")
-        .about("Deploys an .agz file to a new app in the cloud provider described in the deploy config at ~/.alan/deploy.json")
+        .about("Deploys an .agz file to a new app in one the cloud providers described in the deploy config at ~/.alan/deploy.json")
```
cdmistman, line 47:
```suggestion
        .about("Deploys an .agz file to a new app in one of the cloud providers described in the deploy config at ~/.alan/deploy.json")
```

cdmistman (approved):

avm/src/vm/deploy.rs
```diff
@@ -21,40 +21,46 @@ struct AWSCredentials {
...
 
 const CONFIG_NAME: &str = ".alan/deploy.json";
 const CONFIG_SCHEMA: &str = "Please define a deploy config with the following schema: \n{
-  \"aws\": {
-      \"region\": \"string\",
-      \"credentials\": {
-        \"accessKeyId\": \"string\",
-        \"secretAccessKey\": \"string\",
-      }
+  \"aws\": [{
```
dfellis, line 49: Oh, I was thinking of something like:

```json
{
  "myCompany": {
    "provider": "aws",
   ...
  }, 
  "myPersonalProject": {
    "provider": "azure", 
   ... 
  } 
}
```

depombo: So I did not go that route because this config should not really be application/project specific, but just for different cloud credentials / providers. My thought was there would be a one-to-many relationship between a cloud credential and an app/project. The application specific configuration as you describe it would come eventually, be checked into the repo and reference the cloud credential alias from this credentials configuration which really ought to be called credentials.json instead of deploy.json. Region and things like whether we use spot instances would eventually belong in that application specific configuration.

dfellis: So that's why I called the first one `"myCompany"` implying many projects under those credentials. The only difference here between your's and this one is pulling the alias out as the key in a hashmap instead of being another field within an array of JSON values?

depombo: Ah I see, your point is just to key by the `cloudAlias` instead of arrays of cloud `provider`. That makes sense. I understood something else since I would probably use the same `cloudAlias` for all my personal projects.

depombo: The other reason I had not done that is because the schema for `GCP` or `AZURE` might look very different from the ones `AWS` ones so it will be more complex or awkward to enforce a schema on this object, but it makes more sense to the user anyways

dfellis (approved):

PR #399, opened 2021-02-05
Get the JSON test working on the AVM

dfellis:
Need to:

1. Decide if we wait for @cdmistman's change to fix the dependency declarations or
2. Adjust the AVM unit tests to this slower reality for now.


avm/src/vm/event.rs
```diff
@@ -77,6 +77,13 @@ impl EventHandler {
         self.fragments.push(frag);
       }
     } else {
+      // Temporarily disable all IO opcode reordering to test a theory
```
depombo, line 80: I don't think we need this rn? The code that runs IO opcodes within a fragment (with IO parallelization turned off) should already run things sequentially

dfellis: No, that's critical to it working. If the AVM decides to move the group of IO opcodes around (as it does in the JSON test) then it can break execution if the compiler is not setting the dependency graph correctly (which it isn't right now in certain cases, which the JSON test triggers).

It *could* be a mirror of the `cpu` branch, though, and group the IO opcodes together into the same fragment without reordering them, but it's not critical for it to work. I did it this simple way as a test to confirm my theory on why what I was seeing was happening.

depombo: aaaah got it, it is re-ordering them even if it executes them sequentially

depombo:
> Need to:
> 
> 1. Decide if we wait for @cdmistman's change to fix the dependency declarations or
> 2. Adjust the AVM unit tests to this slower reality for now.

Because we are already running io opcodes sequentially this should not slow things down. We are just not reordering the IO opcodes.

cdmistman (approved):

depombo (approved):

PR #400, opened 2021-02-07
Get complex JSON API creation and serialization working

dfellis:
Works for both the js-runtime and AVM now!

### DIFF
--- a/bdd/spec/032_json_spec.sh
+++ b/bdd/spec/032_json_spec.sh
@@ -76,9 +76,8 @@ null"
     End
 
     It "runs agc"
-      Pending debug-what-goes-wrong-on-avm-again
-      #When run test_agc
-      #The output should eq "$COMPLEXOUTPUT"
+      When run test_agc
+      The output should eq "$COMPLEXOUTPUT"
     End
   End
 End
--- a/bdd/spec/032_json_spec.sh
+++ b/bdd/spec/032_json_spec.sh
@@ -41,4 +41,44 @@ null"
       The output should eq "$BASICOUTPUT"
     End
   End
+
+  Describe "complex JSON type construction and printing"
+    before() {
+      sourceToAll "
+        from @std/app import start, print, exit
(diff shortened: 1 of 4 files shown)


bdd/spec/032_json_spec.sh
```diff
@@ -41,4 +41,44 @@ null"
...
+
+  Describe "complex JSON type construction and printing"
+    before() {
+      sourceToAll "
+        from @std/app import start, print, exit
+        from @std/json import JSON, toString, JSONBase, JSONNode, IsObject, Null, newJSONObject, newJSONArray, addKeyVal, push
+
+        on start {
+          newJSONObject()
+            .addKeyVal('mixed', 'values')
```
depombo, line 53: huge fan of this API

std/json.ln
```diff
@@ -532,16 +535,29 @@ export fn toString(j: JSON): string {
...
+        missingChildren = true;
+      }
+    });
+    return missingChildren;
+  }).eachLin(fn (kv: KeyVal<int64, bool>) {
+    nodeList.set(kv.key, false);
+  });
   // Execute the following block over and over again until there are no more nodes to process
-  seqdo(newseq(pow(2, 64).getOr(0)), fn (): bool {
+  seqdo(newseq(pow(2, 63).getOr(0)), fn (): bool {
```
depombo, line 560: what's this about?

depombo (approved):

std/json.ln
dfellis: Oh, the actual 2^64 doesn't fit in 64-bits, so it was hitting the `.getOr(0)` and then never running more than once for me.

depombo: wow, that is a fun bug. we might have other occurrences of `pow(2, 64)` in `root.ln` for the tree implementation no?

dfellis: You're right, there is one other place this exists, I will patch it.

dfellis:
Hmm. I seem to have broken something in this PR and the test suite is running forever. Digging in now, but this isn't going to be as cut-and-dried as I thought :(

cdmistman (approved):

PR #402, opened 2021-02-10
upgrade tokio, basic daemon module and use async everywhere

depombo:


avm/src/main.rs
```diff
@@ -61,69 +64,90 @@ fn main() {
...
-    },
-    ("install",  _) => {
-      let source_file = ".dependencies.ln";
-      if Path::new(source_file).exists() {
-        std::process::exit(compile_and_run(source_file));
-      } else {
-        println!(
-          "{} does not exist. Dependencies can only be installed for {}",
-          source_file, source_file
+  let rt = Builder::new_multi_thread()
```
depombo, line 77: Creating the tokio runtime at the beginning of the CLI now so we can use async everywhere

avm/src/daemon/daemon.rs
```diff
@@ -0,0 +1,28 @@
...
+  let set_app_id = APP_ID.set(app_id.to_string());
+  let set_deploy_key = DEPLOY_KEY.set(deploy_key.to_string());
+  if set_app_id.is_err() || set_deploy_key.is_err() {
+    eprintln!("Failed to globally set app id or deploy key");
+    std::process::exit(1);
+  }
+  let _app_id_global = APP_ID.get().unwrap();
+  let _deploy_key_global = DEPLOY_KEY.get().unwrap();
+  task::spawn(async move {
+    let ten_min_period = Duration::from_secs(60*10);
```
dfellis, line 20: Ten minutes?

avm/src/daemon/daemon.rs
```diff
@@ -0,0 +1,28 @@
+use once_cell::sync::OnceCell;
+use tokio::task;
+use tokio::time::{Duration, sleep};
+
+use crate::vm::run::run;
+
+pub static APP_ID: OnceCell<String> = OnceCell::new();
+pub static DEPLOY_KEY: OnceCell<String> = OnceCell::new();
```
dfellis, line 8: This is fine for now, but presumably we'll be able to replace this in the future during upgrades.

avm/src/daemon/daemon.rs
```diff
@@ -0,0 +1,28 @@
+use once_cell::sync::OnceCell;
+use tokio::task;
+use tokio::time::{Duration, sleep};
+
+use crate::vm::run::run;
```
dfellis, line 5: run run

depombo: lol do you want to leave `exec`? I actually changed the name

avm/src/daemon/daemon.rs
depombo: How much time did you have in mind?

depombo: That was a placeholder and I agree it is too high, but I was still thinking what it should actually be

avm/src/main.rs
```diff
@@ -61,69 +64,90 @@ fn main() {
...
+    .enable_time()
+    .enable_io()
+    .build()
+    .unwrap();
```
dfellis, line 81: Weird that Tokio lives in here, now, but it makes sense.

avm/src/daemon/daemon.rs
depombo: Right, once we have VM ssh access

avm/src/daemon/daemon.rs
dfellis: So, if we go with a uniform sampling frequency, then the frequency of the updates should be ~2x the speed we want to target potential upgrades/downgrades, because of [Nyquist sampling theory](https://en.wikipedia.org/wiki/Nyquist_frequency), so if we want deploys to happen in ~1 minute, we should update the status of the cluster once every 30 seconds.

However, the real scaling needs are based on the actual load the servers are under, and that's going to vary from cluster to cluster, and some are going to have "spikier" demand than others. Internally checking the stats and updating when some threshold is reached is probably better. I'd imagine there's 3 thresholds we care about:

1. max update interval -- if we've reached it, force an update and reset the internal counter; this is good for keeping an eye on low-utilization or stable-utilization clusters
2. min update interval -- if we've triggered an update very recently, we should not fire another update for some time; this should be at least the expected time to upgrade/downgrade a cluster, ideally 2x.
3. last update stats ratio threshold -- for all of the stats sent to the server, hold onto them as the most-recently sent stats and compare the current stats versus them. If the ratio between the two has significantly changed (say, 50% up or down) then push the update assuming we're in between the two intervals

This approach is more complicated, but should reduce bandwidth usage while also being more responsive at the same time.

avm/src/daemon/daemon.rs
dfellis: Doesn't bother me either way. :)

avm/src/daemon/daemon.rs
dfellis: So something like `max = 30, min = 2, ratio = 0.5` would be what I'd propose as our first pass for the more complicated update logic.

dfellis: That doesn't change my note about making sure that the actual upgrade/downgrade rules make sure to avoid "resonance" causing constant creates/deletes of VMs; this is just about the relative change of the stats, not the absolute values of them.

dfellis: though that does imply a fourth metric to push an update on: absolute threshold. In case the load balancing logic somehow failed, the AVM needs to put out an SOS kind of message, since 50% more than 90% utilization is over 100% utilization, and nonsensical.

depombo: > So something like `max = 30, min = 2, ratio = 0.5` would be what I'd propose as our first pass for the more complicated update logic.

What are these units?

depombo: All that sounds good and we can go really deep here, but we have to be careful not to over-engineer for the very first pass. As you put it, perfect is the enemy of good.

dfellis: That's why I put the simple sampling approach first ;)

But the units for the first two are minutes and ratio is unitless.

depombo: > > So something like `max = 30, min = 2, ratio = 0.5` would be what I'd propose as our first pass for the more complicated update logic.
> 
> What are these units?

Minutes, I believe. Hmm

depombo: > All that sounds good and we can go really deep here, but we have to be careful not to over-engineer for the very first pass. As you put it, perfect is the enemy of good.

Actually, this might not be too bad to implement initially. I thought you meant storing stats and looking at the derivative, but you just mean keeping the previous value around.

depombo: > That's why I put the simple sampling approach first ;)
> 
> But the units for the first two are minutes and ratio is unitless.

Oh I think I'm tired. I missed the fact that you were proposing 2 different ones 😴

dfellis: No problem :)

I think a ~1 minute tolerance to increased latencies will be fine for most users, so a 30 second update interval will work, but you can see how you get 60 hits per cluster per minute that way (or 3600 hits per cluster per hour), while the more advanced one *should* have the same sort of reaction speed but potentially only 2 hits per cluster per *hour* and a maximum of only 30 hits per cluster per hour, so literally 120x - 1800x less traffic for the same number of managed clusters.

avm/src/daemon/daemon.rs
cdmistman: It is possible to replace the value <https://doc.rust-lang.org/std/lazy/struct.OnceCell.html#method.take> but we'd have to have a mutable reference.

avm/src/main.rs
cdmistman: i usually use [this](https://docs.rs/tokio/1.2.0/tokio/attr.main.html) ;)

avm/src/daemon/daemon.rs
depombo: Changing to 30 seconds and created https://github.com/alantech/deploy/issues/14 as a followup 🚀

avm/src/main.rs
depombo: I also prefer that, usually, but in this case the macro does not make it obvious that `time` and `io` are on so I prefer the specificity of the builder (I once wasted like 30min because I was trying to do `io` within the runtime and the error was not very helpful then

avm/src/daemon/daemon.rs
depombo: > It is possible to replace the value https://doc.rust-lang.org/std/lazy/struct.OnceCell.html#method.take but we'd have to have a mutable reference.

This made me realize that I can just pass an immutable reference :)

dfellis (approved):

avm/src/main.rs
cdmistman: er, it calls `enable_all` so it *should* enable `io` (maybe you didn't have the feature enabled when you were doing that?)

It's nbd to manually build the runtime and we may need to do it anyways since it provides more flexibility, I was just throwing out there that it's an option that makes it so that the entire program can be `async`

cdmistman (approved):

avm/src/main.rs
depombo: Got it, that makes sense. Maybe I was using a builder and it was a me issue and not a tokio issue. Will leave this as-in case we need to configure something down the line

PR #403, opened 2021-02-10
terminate > kill, display spinner while waiting for VMs to spin up/down

depombo:


avm/src/vm/deploy.rs
```diff
@@ -114,16 +114,20 @@ pub async fn vm_health(app_id: &str, deploy_key: &str) {
...
-    Ok(_) => println!("Killing app with id {} if it exists...\n", app_id),
-    Err(err) => println!("Killing app with id {} failed. Error: {}", app_id, err)
-  }
+  let sp = SpinnerBuilder::new(format!("Terminating app {} if it exists", app_id)).start();
+  let resp = post(format!("{}/v1/terminate", URL), body).await;
+  let res = match resp {
+    Ok(_) => format!("Terminated app {} succesfully!", app_id),
+    Err(err) => format!("Failed to terminate app {}. Error {}", app_id, err),
+  };
+  sp.update(format!(""));
```
cdmistman, line 128: is this needed?

avm/src/vm/deploy.rs
```diff
@@ -201,7 +206,7 @@ pub async fn status() {
 
   let mut data: Vec<Vec<&dyn Display>> = vec![];
   for app in &mut apps {
-    data.push(vec![&app.id, &app.status, &app.url, &app.cloudProvider, &app.cloudAlias]);
+    data.push(vec![&app.id, &app.url, &app.cloudProvider, &app.cloudAlias]);
```
dfellis, line 209: Why does the `status` subcommand no longer have a status field?

depombo: "info" is probably more accurate. Now that we wait for vm instances to spin up and down, the status is always "running". When we have SSH access or store the status of the cluster we will probably want to do something more elaborate to provide an actual "status". We could require the user to provide a `health` endpoint that we check through the main HTTPS url though, but if the user doesn't have that endpoint the app would show up as `unavailable`

depombo: Discussed offline we decided to go with calling the command `info` for now

avm/src/vm/deploy.rs
depombo: removed!

dfellis (approved):

cdmistman (approved):

avm/src/vm/deploy.rs
```diff
@@ -88,7 +88,11 @@ async fn post(url: String, body: Value) -> Result<String, Box<dyn Error>> {
   let mut resp = client.request(req).await?;
   let data = hyper::body::to_bytes(resp.body_mut()).await?;
   let str = String::from_utf8(data.to_vec())?;
```
cdmistman, line 90: I *just* noticed that you're getting a name clash since `str` is a primitive type. Maybe use `data_str`?

depombo: Sure, will address that in the next PR which I already started working on.

PR #404, opened 2021-02-11
JSON Parsing, partial implementation for some of the most basic JSON types so far

dfellis:
Putting this up for testing and being able to swap between laptops for me. Definitely not ready, yet.


dfellis:
Very likely to take a vastly different approach to parsing in the future.

PR #405, opened 2021-02-12
Get peers from DNS and determine leader

depombo:
Missing actually sending system stats to finish CLI part of autoscalable alan deploy

### DIFF
--- a/avm/src/daemon/dns.rs
+++ b/avm/src/daemon/dns.rs
@@ -1,0 +1,44 @@
+use std::str;
+
+use trust_dns_resolver::TokioAsyncResolver;
+use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
+
+pub struct DNS {
+  resolver: TokioAsyncResolver,
+  domain: String
+}
+
+impl DNS {
+  pub fn new(domain: &str) -> DNS {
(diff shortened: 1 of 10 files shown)


avm/src/daemon/dns.rs
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  fn ip_from_txt_data(data: &[u8]) -> String {
+    let txt = str::from_utf8(&*data).expect("Data in TXT record is not a valid string");
+    let err = format!("VM metadata in DNS TXT record has invalid schema version: `{}`", &txt);
+    let parts: Vec<&str> = txt.split("|").collect();
+    if parts.len() != 4 || parts[0] != "v1" {
+      panic!(err);
+    }
+    parts[3].to_string()
```
dfellis, line 42: Feels weird to do all of that work and toss 3 of the four fields. I assume eventually this will be refactored to provide them all in a struct?

cdmistman (approved):

dfellis (approved):

avm/src/daemon/dns.rs
depombo: Yes, eventually we will use surely use other elements of the metadata and we will have a proper struct!

PR #398, opened 2021-02-04
AMM to AGA Mutation Graph

cdmistman:
fixes #381
fixes #331 

~~This should also fix the regression that broke IO opcode parallelization, which was worked-around in #346~~ There's an issue with the AVM opcode batching that results in programs breaking, will fix in future PR

PR #406, opened 2021-02-12
Enable IO opcode parallelization

cdmistman:
#346 introduced a workaround in the AVM that disabled IO opcode parallelization, since the compiler wasn't properly tracking deeply-nested statements for the dependency graph. #398 fixed the dependency tracking in the compiler, but attempting to re-enable IO opcode parallelization resulted in the discovery of numerous bugs. This PR seeks to re-enable parallelization by fixing all of these bugs.

- [x] AVM panics after calling `filter` due to `HandlerMemory::join` not inheriting pointer address spaces.
- [x] only the first io opcode in a batch's return value is used
- [x] 3 panics @ <https://github.com/alantech/alan/blob/io_parallelization/avm/src/vm/memory.rs#L758>
- [x] 3 panics @ <https://github.com/alantech/alan/blob/io_parallelization/avm/src/vm/memory.rs#L494> (`len is 0 but index is 0`)
- [x] wrong output for <https://github.com/alantech/alan/blob/io_parallelization/bdd/spec/012_hash_spec.sh#L46-L60>
- [x] recursion test broken (wrong output)
- [x] JSON test broken (panic)


avm/src/vm/memory.rs
```diff
@@ -793,22 +794,25 @@ impl HandlerMemory {
       }
     }
     // Finally pull any addresses added by the old object into the new with a similar stitching
-    /*if hm.addr.0.len() > self.addr.0.len() {
```
depombo, line 796: Nice, I think @dfellis had commented this code out for a particular reason but I'm not sure what it was

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
depombo, line 1: can we .gitignore this? but also this is mine
```
    {
      "type": "lldb",
      "request": "launch",
      "name": "run agc",
      "program": "${workspaceFolder}/alan/avm/target/debug/alan",
      "args": ["run", "out1.agc"],
      "cwd": "${workspaceFolder}",
      "sourceLanguages": ["rust"],
      "preLaunchTask": "build avm"
    },
```

adding the source language made it the debugger more amicable

avm/src/vm/memory.rs
```diff
@@ -356,6 +368,8 @@ impl HandlerMemory {
     fractal.block[idx].1 = val;
     if fractal.belongs(self) && fractal.hm_addr.is_some() {
       self.write_fractal(fractal.hm_addr.unwrap(), fractal);
+    } else {
+      panic!("Attempting to write fixed data to a fractal owned by a different HandlerMemory");
```
cdmistman, line 372: 👍

avm/src/vm/memory.rs
```diff
@@ -380,9 +394,9 @@ impl HandlerMemory {
       }
       // copy fractal from ancestor
       let addr = fractal.hm_addr.as_ref().unwrap().clone();
-      let (_, hm_opt) = self.addr_to_idxs(addr);
+      let ((a, _), hm_opt) = self.addr_to_idxs(addr);
       let hm = hm_opt.expect("couldn't write fractal to HandlerMemory: address doesn't exist");
-      HandlerMemory::transfer(&hm, addr, self, addr);
+      Arc::get_mut(self).expect("how").mems[a] = hm.mems[a].clone();
       drop(hm);
       todo!()
```
cdmistman, line 401: forgot about this `todo!()` - since we haven't hit this path but we still need to cover it I think it's fair to remove it? I inserted it when I was changing everything to use `Arc<HandlerMemory>` since I thought I'd have to do more here...

avm/src/vm/event.rs
```diff
@@ -192,6 +195,80 @@ impl HandlerFragment {
     };
   }
 
+  #[inline(always)]
```
depombo, line 198: what does this do?

avm/src/vm/event.rs
```diff
@@ -192,6 +195,80 @@ impl HandlerFragment {
...
+
+  #[inline(always)]
+  async fn run_unpred_cpu(
+    &mut self,
+    hand_mem: Arc<HandlerMemory>,
+    instrs: &Vec<Instruction>
+  ) -> Arc<HandlerMemory> {
+    // These instructions are always in groups by themselves
+    let op = &instrs[0];
+    if let OpcodeFn::UnpredCpu(func) = op.opcode.fun {
```
depombo, line 230: so an `unpred_cpu` can never emit an event?

depombo (commented):
Can we revive the unit tests that were commented out in `event.rs`?

avm/src/vm/event.rs
cdmistman: it ensures that the function gets inlined - originally, we were handling each instruction within `HandlerFragment::run` and I wanted the performance to otherwise be the same since I was pulling the handling into other functions

avm/src/vm/event.rs
cdmistman: `emit` is a `cpu` opcode so it gets handled by `run_cpu`

dfellis: I've honestly never liked how we handled event emission in the opcodes. There's really only one opcode that ever emits an event, but *all* of the `cpu!` opcodes have to deal with that. `unpred_cpu` opcodes by themselves are unpredictable because they internally run another handler (and we want all of their mutations to stick around). One or more of the opcodes within that handler can `emit` just fine?

avm/src/vm/memory.rs
dfellis: So drop the entire branch, or just the `todo!()` call?

cdmistman: just the `todo!()` call - although we can keep it and add a message asking the user to report to us their use case so we can reproduce and it's correct, which would make debugging their probably-wrong program easier?

dfellis: Let's go with the latter; I don't think our test suite is comprehensive enough to even consider an end user's program "probably wrong." More likely "AVM is probably wrong", I think?

avm/src/vm/memory.rs
```diff
@@ -104,10 +104,8 @@ pub struct HandlerMemory {
   /// The first is the "normal" memory space, and the second is the args memory space.
   /// Global addresses are fixed for the application and do not need a mutable vector to parse.
   addr: (Vec<Option<(usize, usize)>>, Vec<Option<(usize, usize)>>),
-  /// Specifies which memory block to push "normal" values into.
+  /// Specifies which memory block to push values into.
   mem_addr: usize,
-  /// Specifies which memory block to push "args" into.
-  args_addr: usize,
```
depombo, line 110: nice

avm/src/vm/memory.rs
```diff
@@ -166,7 +163,7 @@ impl HandlerMemory {
   /// Returns true if the idxs are valid in self
   fn is_idx_defined(self: &HandlerMemory, a: usize, b: usize) -> bool {
     let is_raw = a == std::usize::MAX;
-    let safe_mem_space = self.mem_addr == 1 || a >= self.mem_addr;
+    let safe_mem_space = self.mem_addr == 1 || a >= self.mem_addr || self.mems[a].len() > 0;
```
depombo, line 166: 💯

avm/src/vm/event.rs
depombo: That makes sense. I agree with this:
```
I've honestly never liked how we handled event emission in the opcodes. There's really only one opcode that ever emits an event, but all of the cpu! opcodes have to deal with that.
```

depombo (approved):

PR #415, opened 2021-02-16
use anycloud deploy module

depombo:


avm/src/daemon/daemon.rs
```diff
@@ -18,7 +17,7 @@ pub async fn get_ip() -> String {
 
 pub async fn start(agz_file: &str, app_id: &str, deploy_token: &str) {
   let app_id = app_id.to_string();
-  let deploy_token = deploy_token.to_string();
+  let _deploy_token = deploy_token.to_string();
```
cdmistman, line 20: do we even need this?

avm/src/daemon/lrh.rs
```diff
@@ -46,7 +46,7 @@ impl LogRendezvousHash {
 
   // Runs a binary search for the record whose hash is closest to the key hash without
   // going over. If none are found, the *last* record in the list is returned as it wraps around.
-  pub fn get_id_for_key(&self, key: &str) -> &str {
+  pub fn _get_id_for_key(&self, key: &str) -> &str {
```
cdmistman, line 49: do we even need this?

depombo: it will be used later so suppressing the warning for now

avm/src/daemon/daemon.rs
depombo: it will be used later so suppressing the warning for now

avm/Cargo.toml
```diff
@@ -12,7 +12,10 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-ascii_table = "3.0"
+# New commits to a git dependency are locked in place and only replaced by a `cargo update`.
+anycloud = { git = "https://github.com/alantech/anycloud", branch = "main" }
+# For local development, it is possible to specify a relative path location instead of a git repository
+# anycloud = { path = "../../anycloud/cli" }
```
dfellis, line 18: Why does this one have `/cli` at the end but not the git repo above?

depombo: The one above is the github repository and finds the cargo lib recursively, no need (or way) to specify the folder. That one below however specifies a specific folder path.

cdmistman (approved):

avm/Cargo.toml
dfellis: That's bonkers. What if there are two cargo libs in the repo?

dfellis (approved):
Approved, though still perplexed at the Cargo.toml file.

avm/Cargo.toml
depombo: I agree and had the same thought

PR #416, opened 2021-02-17
Daemon sends system stats for cluster leader

depombo:


cdmistman (approved):

avm/src/main.rs
```diff
@@ -67,7 +67,7 @@ fn main() {
     .subcommand(SubCommand::with_name("daemon")
       .about("Run an .agz file in daemon mode. Used on deploy within cloud provider VMs.")
       .arg_from_usage("<APP_ID> 'Specifies the alan app to upgrade'")
-      .arg_from_usage("<AGZ_FILE> 'Specifies the .agz file to deploy'")
+      .arg_from_usage("<AGZ_B64> 'Specifies the .agz program as a base64 encoded string'")
       .arg_from_usage("<DEPLOY_TOKEN> 'Specifies the deploy token'")
```
dfellis, line 71: I'll probably be ripping this out in the near future and replacing the daemon mode with an actual control port approach, btw.

dfellis (approved):

avm/src/daemon/mod.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg_attr(linux, path = "linux.rs")]
+#[cfg_attr(not(linux), path = "rest.rs")]
```
dfellis, line 2: "rest" is a really weird name here. I'd immediately jump to REST and think it's related to the control port that will soon exist. :)

dfellis (approved):

PR #419, opened 2021-02-20
properly get private ip address and alan version in dns vm metadata

depombo:


avm/src/daemon/linux.rs
```diff
@@ -39,71 +39,99 @@ struct VMStatsV1 {
...
+    .arg("-I")
+    .output()
+    .await;
+  let err = "Failed to execute `hostname`";
+  let stdout = res.expect(err).stdout;
+  String::from_utf8(stdout)
+    .expect(err)
+    .trim()
+    .to_string()
+}
```
dfellis, line 53: Are you using this function?

dfellis: You are, you not only want to trim it, but split on `\n` and take just the first value, just in case.

avm/src/daemon/linux.rs
```diff
@@ -39,71 +39,99 @@ struct VMStatsV1 {
...
-          let scale_body = json!({
-            "appId": app_id,
-            "agzB64": agzb64,
-            "deployToken": deploy_token,
-            "clusterDelta": delta,
-          });
-          post("scale", scale_body).await;
+          post_v1_scale(&cluster_id, &agzb64, &deploy_token, alan_version, delta).await;
         }
+        // TODO maybe ask deploy service to /upgrade if delta = 0 and not on the latest alan version
```
dfellis, line 134: I don't think so. There's still the regression potentiality. We should only upgrade the `alan` version on user-triggered upgrades.

avm/src/daemon/linux.rs
dfellis: Wait, sorry, split on space ` `, not `\n`

dfellis (approved):

dfellis (approved):

PR #420, opened 2021-02-21
Update to the latest anycloud deploy module, which un-DRYs some code but since it's only two places, it's okay

dfellis:


depombo (approved):

depombo:
@dfellis looks good, but seems to be complaining about `cargo:rerun-if-changed=alan/anycloud.ln` https://github.com/alantech/alan/pull/420/checks?check_run_id=1944448313#step:3:435

dfellis:
> @dfellis looks good, but seems to be complaining about `cargo:rerun-if-changed=alan/anycloud.ln` https://github.com/alantech/alan/pull/420/checks?check_run_id=1944448313#step:3:435

It worked locally because I have an `alan` application available for that build to work. :/ We'll have to do something bonkers and circular to get the build process to pass. :(

depombo (approved):

PR #421, opened 2021-02-23
Use cluster factor instead of delta and update anycloud

depombo:


avm/src/daemon/linux.rs
```diff
@@ -65,13 +65,13 @@ async fn post_v1(endpoint: &str, body: Value) -> String {
   }
 }
 
-async fn post_v1_scale(cluster_id: &str, agz_b64: &str, deploy_token: &str, alan_version: &str, delta: i32) -> String {
+async fn post_v1_scale(cluster_id: &str, agz_b64: &str, deploy_token: &str, alan_version: &str, factor: &str) -> String {
```
dfellis, line 68: Why is this a string instead of a float64?

dfellis (approved):
It should work, just weirdly typed.

avm/src/daemon/linux.rs
depombo: I realized that it doesn't make a sense to parse it back and forth since the deploy service emits it and handles it so I'm just transmitting it as-is now from the daemon

dfellis: Oh, but doesn't it change the type when it is re-serialized back into JSON over the wire into a string type? This may only be working because of automatic type coercion in Javascript.

depombo: the [endpoint](https://github.com/alantech/deploy/blob/1b3a83cad7cb0f1fa52de4c39f4b1d8ea3ef809c/src/routesV1.ts#L80) returns a string, `1` and not json :)

dfellis: wtf

cdmistman (approved):

PR #424, opened 2021-02-24
Listen on cluster size changes to determine cluster leader

depombo:
- pass in cluster id when posting stats 
- remove alan version which is readily available via DNS given the cluster id
- pick a new leader when the cluster size changes

### DIFF
--- a/avm/src/daemon/linux.rs
+++ b/avm/src/daemon/linux.rs
@@ -103,12 +103,12 @@ async fn get_v1_stats() -> VMStatsV1 {
 }
 
 // returns cluster delta
-async fn post_v1_stats(deploy_token: &str, cluster_size: usize) -> String {
+async fn post_v1_stats(cluster_id: &str, deploy_token: &str) -> String {
   let vm_stats = get_v1_stats().await;
   let stats_body = json!({
     "deployToken": deploy_token,
     "vmStats": vm_stats,
-    "clusterSize": cluster_size,
+    "clusterId": cluster_id,
   });
   post_v1("stats", stats_body).await
 }
@@ -120,7 +120,6 @@ pub async fn start(cluster_id: &str, agz_b64: &str, deploy_token: &str) {
   task::spawn(async move {
     let dns = DNS::new(DOMAIN);
     let vms = dns.get_vms(&cluster_id).await;
-    let cluster_size = vms.len();
     let ips = vms.iter().map(|vm| vm.private_ip_addr.to_string()).collect();
     let alan_version = &vms[0].alan_version;
     let lrh = LogRendezvousHash::new(ips);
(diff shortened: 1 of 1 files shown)


dfellis (dismissed):

avm/src/daemon/linux.rs
```diff
@@ -65,9 +65,8 @@ async fn post_v1(endpoint: &str, body: Value) -> String {
   }
 }
 
-async fn post_v1_scale(cluster_id: &str, agz_b64: &str, deploy_token: &str, alan_version: &str, factor: &str) -> String {
+async fn post_v1_scale(cluster_id: &str, agz_b64: &str, deploy_token: &str, factor: &str) -> String {
   let scale_body = json!({
-    "alanVersion": alan_version,
```
depombo, line 70: this is available in the deploy service so dropping it here

cdmistman, line 70: why are we removing this part? I feel like it's still important information, even if it's basically just a sanity check between all nodes

depombo: Oh this was never really necessary since it is readily available via DNS for the recipient and all nodes in the cluster

depombo: Updated the PR description too

cdmistman (approved):

avm/src/daemon/linux.rs
```diff
@@ -118,25 +117,30 @@ pub async fn start(cluster_id: &str, agz_b64: &str, deploy_token: &str) {
   let deploy_token = deploy_token.to_string();
   let agzb64 = agz_b64.to_string();
   task::spawn(async move {
+    // TODO better period determination
+    let period = Duration::from_secs(30);
```
dfellis, line 121: So, with respect to re-evaluating which node is in charge, this is fine, and potentially we crank that up to, say, every 1-2 seconds?

avm/src/daemon/linux.rs
```diff
@@ -118,25 +117,30 @@ pub async fn start(cluster_id: &str, agz_b64: &str, deploy_token: &str) {
...
+      }
+      if leader_ip == self_ip {
+        let factor = post_v1_stats(&cluster_id, &deploy_token).await;
+        println!("VM stats sent for cluster {} of size {}. Cluster factor: {}.", cluster_id, vms.len(), factor);
         if factor != "1" {
-          post_v1_scale(&cluster_id, &agzb64, &deploy_token, alan_version, &factor).await;
+          post_v1_scale(&cluster_id, &agzb64, &deploy_token, &factor).await;
         }
-        sleep(period).await
       }
```
dfellis, line 142: Here is where we'd potentially have some back-off logic. Ideally the `deploy` service can send a "do not contact until X timestamp" flag that this just obeys, and the logic of how often to actually call is adjustable and doesn't have to be in here.

depombo: Yes :)
https://github.com/alantech/deploy/issues/14

avm/src/daemon/linux.rs
dfellis: Based on our conversation, it may be best to crank this from 30 to 180 to be on the safe side for now?

dfellis (approved):

dfellis (approved):

PR #426, opened 2021-02-26
HTTP Client Rewrite

dfellis:
Creates a new `fetch` function backed by a singular opcode and uses the same `InternalRequest` and `InternalResponse` data structures as the HTTP Server, so it's fully possible to proxy through from one to the other, now. :)

There's some bugfixes to the compiler and also some hackery going on as I realized the "builtinTypes" don't actually line up with reality sometimes and that was causing some weird type matching bugs. I also added more scope dual-inheritance in there that I hope doesn't break anything else.


compiler/src/lntoamm/Type.ts
```diff
@@ -502,25 +502,23 @@ export class Type {
     const otherTypeAst = fulltypenameAstFromString(otherType.typename)
     let generics = []
     if (typeAst.has('opttypegenerics')) {
-      const genericsAst = typeAst.get('opttypegenerics').get('generics')
-      generics.push(genericsAst.get('fulltypename').t)
-      genericsAst.get('cdr').getAll().forEach(r => {
+      generics.push(typeAst.get('opttypegenerics').get('generics').get('fulltypename').t)
+      typeAst.get('opttypegenerics').get('generics').get('cdr').getAll().forEach(r => {
         generics.push(r.get('fulltypename').t)
       })
```
dfellis, line 508: Somehow this code and the similar block below weren't being run *at all* until now, shockingly. It was translated incorrectly back when I switched from ANTLR and never noticed.

compiler/src/lntoamm/Type.ts
```diff
@@ -502,25 +502,23 @@ export class Type {
...
     }
     generics = generics.map(
-      g => scope.deepGet(g) || Type.fromStringWithMap(g, interfaceMap, scope) as Type
+      g => scope.deepGet(g) || Type.fromStringWithMap(g, interfaceMap, scope) as Type || new Type('-bogus-', false, true)
```
dfellis, line 511: Instead of the `typeApplies` call crashing, I realized that it should always return a totally bogus type that shouldn't match with anything if it can't find the type at all. Then I got much saner errrors instead of stacktraces.

compiler/src/lntoamm/UserFunction.ts
```diff
@@ -642,9 +642,11 @@ ${statements[i].statementAst.t.trim()} on line ${statements[i].statementAst.line
 
   microstatementInlining(
     realArgNames: Array<string>,
-    scope: Scope,
+    s: Scope,
     microstatements: Array<Microstatement>,
   ) {
+    const scope = new Scope(s)
+    scope.secondaryPar = this.scope
```
dfellis, line 649: I changed it so function inlining and dispatching get access to both the current scope and the scope they were originally created in to make sure unexported types are available for use. Hopefully this doesn't break something else, this is the piece I'm most worried about.

compiler/src/lntoamm/opcodes.ts
```diff
@@ -29,6 +29,22 @@ opcodeScope.put(
...
+  headers: opcodeScope.get('Array<KeyVal<string, string>>') as Type,
+  body: opcodeScope.get('string') as Type,
+  connId: opcodeScope.get('int64') as Type,
+}
+Type.builtinTypes.InternalResponse.properties = {
+  status: opcodeScope.get('int64') as Type,
+  headers: opcodeScope.get('Array<KeyVal<string, string>>') as Type,
+  body: opcodeScope.get('string') as Type,
+  connId: opcodeScope.get('int64') as Type,
+}
```
dfellis, line 47: I despise this hackery here, but you can't define a `static` element of a class that is self-referential, so constructing the `builtinTypes` field on `Type` will need to be reworked completely to avoid this, I realized, because it is itself referring to a built-in Type that is generic in its definition (`Array<KeyVal<string, string>>`) and wasn't getting defined properly before.

std/http.ln
```diff
@@ -46,6 +35,109 @@ export type Response {
   connId: int64,
 }
 
+/**
+ * HTTP Client
+ */
```
dfellis, line 40: The new client code became a *lot* bigger than I had hoped for since I couldn't implement `Result`'s `map` function. Really want function types! :)

std/root.ln
```diff
@@ -173,7 +173,7 @@ export fn toString(n: Result<Stringifiable>): string {
 }
 export fn getOrExit(result: Result<any>): any {
   if result.isErr() {
-    stderrp(result.toString());
+    stderrp(result.getErr(noerr()).toString());
```
dfellis, line 176: This works around the fact that I can't `toString` directly all `Result` types, but I can the `Error` type.

std/root.ln
```diff
@@ -2411,7 +2411,7 @@ export fn newHashMap(firstKey: Hashable, firstVal: any): HashMap<Hashable, any>
   return hm.set(firstKey, firstVal);
 }
 
-export fn toHashMap(kva: Array<KeyVal<Hashable, any>>) {
+export fn toHashMap(kva: Array<KeyVal<Hashable, any>>): HashMap<Hashable, any> {
```
dfellis, line 2414: The compiler at one point was complaining that the output type of this was `void`. I don't know how the type inference failed there, but I decided to just add it in if it was going to fail in edge cases.

avm/src/vm/opcode.rs
```diff
@@ -2758,48 +2758,77 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
     })
   });
 
-  async fn __httpget(uri: String) -> std::result::Result<String, Box<dyn std::error::Error>> {
-    let url = TryInto::<Uri>::try_into(uri)?;
-    let mut response = Client::builder()
+  fn __httpreq(
+    method: String,
```
depombo, line 2762: nice!

bdd/spec/023_http_spec.sh
```diff
@@ -61,6 +61,52 @@ Describe "@std/http"
...
+      sourceToAll "
+        from @std/app import start, print, exit
+        from @std/http import fetch, Request, Response
+
+        on start {
+          const res = fetch(new Request {
+            method: 'GET',
+            url: 'https://raw.githubusercontent.com/alantech/hellodep/aea1ce817a423d00107577a430a046993e4e6cad/index.ln',
+            headers: newHashMap('Content-Length', '0'),
+            body: '',
```
depombo, line 75: `GET`s with bodies always weird me out even if the content length is set to nil

depombo (approved):

bdd/spec/023_http_spec.sh
dfellis: That's why you should just use `get(url)` for GET calls :)

avm/src/vm/opcode.rs
```diff
@@ -2758,48 +2758,77 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    body: Option<String>,
+  ) -> ResponseFuture {
+    let mut req = Request::builder()
+      .method(method.as_str())
+      .uri(uri.as_str());
+    for header in headers {
+      req = req.header(header.0.as_str(), header.1.as_str());
+    }
+    let req_obj = if body.is_some() {
+      req.body(Body::from(body.unwrap())).unwrap()
```
cdmistman, line 2774: you can get rid of `body.unwrap()` with `if let Some(body) = body`

avm/src/vm/opcode.rs
```diff
@@ -2758,48 +2758,77 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    } else {
+      req.body(Body::empty()).unwrap()
+    };
```
cdmistman, line 2777: assuming our code is perfect, if `req.body(...)` returns an `Err` we should probably give that error to the user, since this is an issue with input and can probably be resolved.

avm/src/vm/opcode.rs
```diff
@@ -2758,48 +2758,77 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      let method = HandlerMemory::fractal_to_string(hand_mem.read_from_fractal(&req, 0).0);
+      let url = HandlerMemory::fractal_to_string(hand_mem.read_from_fractal(&req, 1).0);
+      let headers = hand_mem.read_from_fractal(&req, 2).0;
+      let mut out_headers = Vec::new();
+      for i in 0..out_headers.len() {
+        let header = hand_mem.read_from_fractal(&headers, i).0;
+        let key = HandlerMemory::fractal_to_string(hand_mem.read_from_fractal(&header, 0).0);
+        let val = HandlerMemory::fractal_to_string(hand_mem.read_from_fractal(&header, 1).0);
+        out_headers.push((key, val));
       }
```
cdmistman, line 2794: this for loop doesn't do anything

avm/src/vm/opcode.rs
```diff
@@ -2758,48 +2758,77 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-    let res = String::from_utf8(body.to_vec())?;
-    Ok(res)
-  }
-  io!(httppost => fn(args, mut hand_mem) {
-    Box::pin(async move {
-      let uri = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[0]));
-      let payload = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[1]));
-      let res = __httppost(uri, payload).await;
+      let body = HandlerMemory::fractal_to_string(hand_mem.read_from_fractal(&req, 3).0);
+      let out_body = if body.len() > 0 { Some(body) /* once told me... */ } else { None };
```
cdmistman, line 2796: 😂

avm/src/vm/opcode.rs
```diff
@@ -2758,48 +2758,77 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+          let mut headers_hm = HandlerMemory::new(None, headers.len() as i64);
+          let mut i = 0;
+          for (key, val) in headers.iter() {
+            let key_str = key.as_str();
+            let val_str = val.to_str().unwrap();
+            headers_hm.init_fractal(i);
+            headers_hm.push_fractal(i, HandlerMemory::str_to_fractal(key_str));
+            headers_hm.push_fractal(i, HandlerMemory::str_to_fractal(val_str));
+            i = i + 1;
+          }
```
cdmistman, line 2815: shortcut:
```rust
for ((key, val), i) in headers.iter().enumerate() {
  ...
}
```

avm/src/vm/opcode.rs
```diff
@@ -2758,48 +2758,77 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+        Ok(mut res) => {
+          let mut res_hm = HandlerMemory::new(None, 3);
+          res_hm.init_fractal(0);
+          res_hm.push_fixed(0, res.status().as_u16() as i64);
+          let headers = res.headers();
+          let mut headers_hm = HandlerMemory::new(None, headers.len() as i64);
+          let mut i = 0;
+          for (key, val) in headers.iter() {
+            let key_str = key.as_str();
+            let val_str = val.to_str().unwrap();
```
cdmistman, line 2810: should we let the user handle invalid response headers?

avm/src/vm/opcode.rs
```diff
@@ -2758,48 +2758,77 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+            i = i + 1;
+          }
+          HandlerMemory::transfer(&headers_hm, 0, &mut res_hm, CLOSURE_ARG_MEM_START);
+          res_hm.push_register(0, CLOSURE_ARG_MEM_START);
+          let body = hyper::body::to_bytes(res.body_mut()).await;
+          if body.is_ok() {
+            let body_str = String::from_utf8(body.unwrap().to_vec()).unwrap_or("".to_string());
+            res_hm.push_fractal(0, HandlerMemory::str_to_fractal(&body_str));
+          } else {
+            res_hm.push_fractal(0, HandlerMemory::str_to_fractal(&""));
```
cdmistman, line 2823: same as above - if the body is invalid shouldn't we let the user handle it?

avm/src/vm/opcode.rs
```diff
@@ -2758,48 +2758,77 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+            let val_str = val.to_str().unwrap();
+            headers_hm.init_fractal(i);
+            headers_hm.push_fractal(i, HandlerMemory::str_to_fractal(key_str));
+            headers_hm.push_fractal(i, HandlerMemory::str_to_fractal(val_str));
+            i = i + 1;
+          }
+          HandlerMemory::transfer(&headers_hm, 0, &mut res_hm, CLOSURE_ARG_MEM_START);
+          res_hm.push_register(0, CLOSURE_ARG_MEM_START);
+          let body = hyper::body::to_bytes(res.body_mut()).await;
+          if body.is_ok() {
```
cdmistman, line 2819: also here you can use `if let Ok(body) = body` to avoid the extra `unwrap`

js-runtime/index.js
```diff
@@ -924,20 +924,18 @@ module.exports = {
   seqrec: (seq, recurseFn) => [seq, recurseFn],
 
   // IO opcodes
-  httpget:  async url => {
+  httpreq:  async req => {
```
cdmistman, line 927: much simpler impl haha

avm/src/vm/opcode.rs
dfellis: Oops! And it worked because I wasn't sending anything with (real) headers.

js-runtime/index.js
dfellis: Well, the opcode was based in part on this API, but this API is also very close to just generating the entire request string yourself and splatting onto a TCP socket.

avm/src/vm/opcode.rs
dfellis: We should definitely guard against maliciously crafted headers instead of crashing here, I agree with you on that. The code is getting very branchy and nested as I tackle this, though.

avm/src/vm/opcode.rs
dfellis: Same as above as well, I will, but the code is starting to look not great with all of these branches.

cdmistman (approved):

PR #428, opened 2021-02-26
Bump the dependencies of the AVM

dfellis:


avm/Cargo.lock
```diff
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
```
cdmistman, line 2: i still want to review though!

cdmistman (approved):

avm/Cargo.lock
depombo: lol

depombo (approved):

PR #429, opened 2021-02-27
Fix http client and server headers in the AVM and fix a type inference bug in the compiler

dfellis:


std/http.ln
```diff
@@ -48,16 +48,16 @@ export fn fetch(req: Request): Result<Response> {
     connId: 0, // Unused by the client
   };
   const ro = httpreq(i);
-  if ro.isErr() {
-    return ro;
-  } else {
+  if ro.isOk() {
```
dfellis, line 51: So what's happening here is, when there's early return, it figures out the return type by checking branch-by-branch until it has a fully-realized type. In this case, since I didn't want to unwrap the error object, the variable it was returning in the error path was `Result<InternalResponse>` instead of `Result<Response>` like I wanted. Simply flipping the conditionals around fixed the inferred type here.

avm/src/vm/opcode.rs
```diff
@@ -2866,7 +2868,7 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
       let mut res_hm = HandlerMemory::new(None, 3);
       res_hm.init_fractal(0);
       res_hm.push_fixed(0, res.status().as_u16() as i64);
-      HandlerMemory::transfer(&headers_hm, 0, &mut res_hm, CLOSURE_ARG_MEM_START);
+      HandlerMemory::transfer(&headers_hm, CLOSURE_ARG_MEM_START, &mut res_hm, CLOSURE_ARG_MEM_START);
```
dfellis, line 2871: What I've done here is add another bit of nesting; before we were accidentally grabbing the first record of the array and returning it as if it was the whole array, eg instead of getting `Array<KeyVal<string, string>>` it was getting `KeyVal<string, string>`.

That also explains why the test of the `length()` was `2` in the test suite, because `KeyVal<string, string>` being treated as an array would show up with 2 elements.

js-runtime/index.js
```diff
@@ -933,7 +933,7 @@ module.exports = {
         body: body.length > 0 ? body : undefined,
       });
       const rstatus = response.status
-      const rheaders = Object.entries(response.headers)
+      const rheaders = [...response.headers.entries()].map(kv => [kv[0] + '', kv[1] + ''])
```
dfellis, line 936: `response.headers` wasn't actually an object, it was a `Headers` type with some weirdness. Of course JS being JS it just succeeded at doing effectively nothing before.

depombo (approved):

PR #430, opened 2021-03-02
Rework conditional rewrites

### LINKED ISSUE #307: Simplifying the conditional rewrite logic
Simplifying/unifying the conditional rewrite logic to hopefully solve the issue with return statements within conditionals that are within an array method closure function.

cdmistman:
fixes #307 
- [x] update compiler's understanding of new conditional opcodes
- [ ] update compiler's rewrite phase for conditional execution
- [x] update JS runtime's conditional opcodes
- [x] update AVM's conditional opcodes


compiler/src/lntoamm/UserFunction.ts
```diff
@@ -8,6 +8,8 @@ import StatementType from './StatementType'
 import Type from './Type'
 import { Args, Fn, } from './Function'
 import { LPNode, } from '../lp'
+import { stat } from 'fs'
+import { passCommandToAction } from 'commander'
```
dfellis, line 12: What are these imports for? They don't seem to be used.

cdmistman: autocomplete :/

compiler/src/lntoamm/UserFunction.ts
```diff
@@ -654,12 +654,50 @@ ${statements[i].statementAst.t.trim()} on line ${statements[i].statementAst.line
...
     for (let i = originalStatementLength; i < microstatements.length; i++) {
       if (microstatements[i].statementType === StatementType.ENTERFN) {
         microstatements.splice(i, 1)
         i--
+      } else if (microstatements[i].statementType === StatementType.TAIL) {
+        tailed = [
+          // TAIL mStmt
+          microstatements.splice(i, 1)[0],
+          // everything that goes in the tail function. even if there are no more elements,
+          // assume we still need to provide an empty closure.
```
dfellis, line 667: Currently an empty closure causes AGA to AGC to fail, and the AVM has no idea how to handle a handler with no instructions. Is there something that can be done instead in that situation?

cdmistman: per an offline conversation, we agreed to create an AVM `nop` closure (cc #433)

dfellis:
Closing this (but not deleting it @cdmistman if you wanted to grab your handiwork still) because the AVM is going away and the compiler is being rewritten in Rust to target Rust instead. So that means we can just let conditionals be conditionals and have LLVM rewrite them into opcodes for us :)

PR #432, opened 2021-03-04
Switch to rustls for hyper in the AVM

dfellis:
This doesn't add HTTPS support, but sets us up for it, since [hyper-rustls](https://github.com/ctz/hyper-rustls/blob/master/examples/server.rs) has server support, but [hyper-tls](https://github.com/hyperium/hyper-tls/tree/master/examples) apparently does not.

### DIFF
--- a/avm/Cargo.lock
+++ b/avm/Cargo.lock
@@ -37,7 +37,7 @@ dependencies = [
  "futures",
  "heim",
  "hyper",
- "hyper-tls",
+ "hyper-rustls",
  "num_cpus",
  "once_cell",
  "rand 0.7.3",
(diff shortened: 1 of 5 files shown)


depombo (approved):
The daemon's dns client uses rust-tls but I don't think that should be a problem or interfere at all with this

dfellis:
> The daemon's dns client uses rust-tls but I don't think that should be a problem or interfere at all with this

I was wondering why `hyper-tls` didn't disappear in my PR. But yeah that should be fine.

cdmistman (approved):

cdmistman:
should be fine, although we should try and prefer `rustls` as it's being used by more and more projects (eg `curl`)

PR #433, opened 2021-03-04
nop closure for avm

cdmistman:


avm/src/vm/event.rs
```diff
@@ -20,13 +20,15 @@ pub enum BuiltInEvents {
...
   HTTPCONN,
+  NOP,
 }
 
 impl From<BuiltInEvents> for i64 {
   fn from(ev: BuiltInEvents) -> Self {
     match ev {
       BuiltInEvents::START => -9213673853036498142,
       BuiltInEvents::HTTPCONN => -9214243417005793441,
+      BuiltInEvents::NOP => i64::MAX,
```
dfellis, line 31: I do wonder if we should make it the negative max so it's negative like the rest of the events? That would also reduce the amount of edge case handling in the compiler since it should never emit a negative number for a custom event?

dfellis (approved):

compiler/src/ammtoaga/index.ts
```diff
@@ -193,23 +196,26 @@ const closuresFromDeclaration = (
       [ name, ...scope, ], // Newest scope gets highest priority
       graph,
     )
-  ).reduce((obj, rec) => ({
+  ).filter((clos) => clos != null).reduce((obj, rec) => ({
```
dfellis, line 199: Oh, just noticed, this should probably be `clos !== null`

avm/src/vm/program.rs
```diff
@@ -84,6 +84,11 @@ impl Program {
     let httpconn = i64::from(BuiltInEvents::HTTPCONN);
     self.event_pls.insert(httpconn, 0);
     self.event_handlers.insert(httpconn, Vec::new());
+    // NOP
+    // the compiler should prevent programs from handling this event.
```
depombo, line 88: but sometimes it does? this comment is confusing to me, can you elaborate within it?

avm/src/vm/opcode.rs
```diff
@@ -3205,9 +3253,13 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
       while current < limit && hand_mem.read_fixed(CLOSURE_ARG_MEM_START) > 0 {
-        hand_mem = body_handler.clone().run(hand_mem).await;
+        if args[2] != NOP_ID {
+          hand_mem = body_handler.clone().run(hand_mem).await;
+        }
         current = current + 1;
-        hand_mem = cond_handler.clone().run(hand_mem).await;
+        if args[2] != NOP_ID {
+          hand_mem = cond_handler.clone().run(hand_mem).await;
+        }
```
dfellis, line 3262: Instead of this, maybe just detect that the body is a no-op at the top like the `cond_handler` and just short-circuit by consuming the `seq` object? (set the iterator to the same as its limit value)

avm/src/vm/program.rs
cdmistman: i just mean that there shouldn't be a way to say `on NOP { ... }`

depombo: ooooh gotcha. maybe `the compiler does not allow explicitly defining handlers for this event and it's used internally by the compiler for closures in unused conditional branches`?

depombo (approved):

dfellis (approved):
Just realized that I forgot to approve this.

PR #434, opened 2021-03-04
Remove the http listen opcode (from users) and auto-listen on 80 or 8000 depending on context

dfellis:
Had to split the http client from the server because the `__conn` event handler is auto-attached when importing and that handler was being used to determine whether or not the http server should be started or not, but it now works as intended, and we can make it switch to HTTPS in the daemon with the appropriate PEM file being handed to it in a follow-up PR. :)

### DIFF
--- a/avm/src/vm/program.rs
+++ b/avm/src/vm/program.rs
@@ -1,10 +1,10 @@
 use std::collections::HashMap;
 
+use once_cell::sync::OnceCell;
+
 use crate::vm::event::{BuiltInEvents, EventHandler};
 use crate::vm::instruction::Instruction;
-use crate::vm::opcode::ByteOpcode;
-
-use once_cell::sync::OnceCell;
+use crate::vm::opcode::{ByteOpcode, opcode_id, OPCODES};
 
 // Facilitates parsing the alan graph code program
 struct BytecodeParser {
@@ -65,6 +65,8 @@ pub struct Program {
   pub(crate) event_pls: HashMap<i64, i64>,
   /// Memory of the program for global variables and string literals
   pub(crate) gmem: Vec<(usize, i64)>,
+  /// The port the http server should use
+  pub(crate) http_port: u16,
 }
 
 pub static PROGRAM: OnceCell<Program> = OnceCell::new();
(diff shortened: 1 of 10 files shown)


depombo (approved):

avm/src/vm/program.rs
```diff
@@ -164,6 +167,33 @@ impl Program {
...
+        let start_handler = &mut start_handlers[0];
+        let start_handler_fragment_idx = start_handler.fragments.len() - 1;
+        let last_frag = &mut start_handler.fragments[start_handler_fragment_idx];
+        let last_id = last_frag[last_frag.len() - 1].id;
+        start_handler.add_instruction(Instruction {
+          id: i64::MAX, // That shouldn't collide with anything
+          opcode: OPCODES.get(&opcode_id("httplsn")).unwrap(),
+          args: vec![0, 0, 0],
+          dep_ids: vec![last_id], // To make sure this is the last instruction run (usually)
+        });
```
depombo, line 194: actually why don't we just have an `append_instruction` method that handles this and that way we don't have to do this here or make fragments public?

dfellis: I need the ID of the last instruction in order to add it to the dependency list and make sure it doesn't accidentally run the httplsn opcode before the `_start` handler is finished, though. (Or we start listening to events before the docker container is built and running for anycloud, for instance.)

depombo: wouldn't that id be available in the within the handler object the same way you access it here?

dfellis: No. I mean the instruction ID, which will vary from handler to handler and I see no way to get at it after the processing of the handler is complete.

depombo (approved):

avm/src/vm/program.rs
depombo: Discussed offline, this logic is effectively modifying the agz binary, but it doesn't belong in the compiler and there is no clean way to do it so better to leave it contained in one place. If we find another use case for an internal "event completion" event that we can hang internal code on for something other than this use case (when _start is done, start a web server if some condition is met) we should generalize this.

aguillenv (approved):

PR #435, opened 2021-03-05
make execop an unpredcpu opcode

depombo:


dfellis:
To summarize some discussion: This change un-parallelizes usage of `exec` and brings to question whether or not automatic parallelization of external IO is something that we even want, if that is the case? The same argument can be made for the `http.get` method here.

I think we need to actually think more on this before we toggle stuff back and forth in behavior/functionality.

PR #436, opened 2021-03-06
Rework IO Parallelization RFC

dfellis:


rfcs/014 - Rework IO Parallelization.md
```diff
@@ -0,0 +1,151 @@
...
+```js
+const googleHomepage = await get('https://google.com');
+const yahooHomepage = await get('https://yahoo.com');
+```
+
+```js
+const homepages = await Promise.all(['https://google.com', 'https://yahoo.com'].map(get));
+```
+
+The third opcode type: `unpred_cpu!` gives us an out; if we had a version of every relevant IO opcode where you could opt in to the current semantics and another version that behaved like `unpred_cpu!` does that is the default, then we get cleaner code in both paths. A few considered suffix names for these: `Par`, like `execPar` based on `reducePar`, but where parallelization makes sense with reduce since there's guaranteed to be many calls, the potential advantages involved with calling `execPar` with only one `exec` usage (but perhaps other IO opcode uses that can batch with it) are unclear. `Async` was considered to potentially confuse people with Rust/C#/JS async/await and people would expect Promises/Futures they need to attach a handler or await on. So while not great, the current suffix under consideration is `Eager`. Eg,
```
depombo, line 113: For the sake of completion/posterity, after we talked I thought of `Disorderly` or `Whenever` as alternatives to `Eager`. However, `Eager` is both shorter and makes much more sense semantically

depombo (approved):
Well put!

cdmistman (approved):

PR #437, opened 2021-03-06
Implement RFC 014, adding a `syncop` opcode and using it to force-linearize exec and fetch

dfellis:


dfellis:
This also fixes `execop` which apparently had broken semantics in the AVM versus the js transpile.

std/http.ln
```diff
@@ -61,6 +61,46 @@ export fn fetch(req: Request): Result<Response> {
   }
 }
 
+export fn fetch(req: Request): Result<Response> {
```
depombo, line 64: not very DRY yet, but soon!

depombo (approved):

std/http.ln
dfellis: soon

avm/src/vm/opcode.rs
```diff
@@ -2837,15 +2840,25 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
       let ms = hand_mem.read_fixed(args[0]) as u64;
       sleep(Duration::from_millis(ms)).await;
       hand_mem
     })
   });
 
+  unpred_cpu!(syncop => fn(args, mut hand_mem) {
+    Box::pin(async move {
+      let closure = HandlerFragment::new(args[0], 0);
+      hand_mem.register(CLOSURE_ARG_MEM_START + 1, args[1], true);
```
cdmistman, line 2854: This only works for `execop` and `fetchop` since they only have 1 param right?

dfellis: Mostly correct (`httpsend` also works). This is part of why it's a temporary solution. However, of the `io!` opcodes that aren't the array opcodes that we wouldn't try this on, only the datastore opcodes are incompatible with this, and since we both don't document datastore in docs.alan-lang.org and because we intend to make some changes to it, I figured it was okay to leave that as is for now.

PR #438, opened 2021-03-08
Add daemon domain argument

aguillenv:
The `alan daemon` stop using the hardcoded `anycloudapp.com` value and use the provided domain with its dns client.


cdmistman (approved):

depombo (approved):
Good to land after doing alpha release testing!

PR #441, opened 2021-03-09
Add deploy --app-id option

aguillenv:
Fixes: alantech/meta#157

### DIFF
--- a/avm/src/main.rs
+++ b/avm/src/main.rs
@@ -58,7 +58,7 @@ fn main() {
         .about("Deploys an .agz file to a new app with one of the deploy configs at ~/.anycloud/deploy.json")
         .arg_from_usage("<AGZ_FILE> 'Specifies the .agz file to deploy'")
         .arg_from_usage("[DEPLOY_NAME] 'Specifies the name of the deploy config to use, or the first definition if not specified'")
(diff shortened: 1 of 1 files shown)


avm/src/main.rs
```diff
@@ -58,6 +58,7 @@ fn main() {
         .about("Deploys an .agz file to a new app with one of the deploy configs at ~/.anycloud/deploy.json")
         .arg_from_usage("<AGZ_FILE> 'Specifies the .agz file to deploy'")
         .arg_from_usage("[DEPLOY_NAME] 'Specifies the name of the deploy config to use, or the first definition if not specified'")
+        .arg_from_usage("--app-id=[APP_ID] '[Optional] Specifies the application identifier'")
```
cdmistman, line 61: I don't think it's required to put `[Optional]` here

depombo: same here https://github.com/alantech/anycloud/pull/24/files#r590443000

aguillenv: Yes, thats only for the message when we execute `--help`. I've added to make it explicit but I know it is redundant.

dfellis (approved):

dfellis (approved):

avm/src/main.rs
aguillenv: Updated

cdmistman (approved):

Issue #443, opened 2021-03-10
HTTPS Server crashes if TLS handshake fails


dfellis:
The process continues running, but it no longer accepts requests.

Steps to reproduce: spin up an alan daemon locally with a self-signed cert, then run:

```
damocles@Talyn:/usr/local/alancorp/alan(main)$ curl -k https://localhost
Hello, HTTPS!damocles@Talyn:/usr/local/alancorp/alan(main)$ 
damocles@Talyn:/usr/local/alancorp/alan(main)$ curl https://localhost
curl: (60) SSL certificate problem: self signed certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
damocles@Talyn:/usr/local/alancorp/alan(main)$ curl -k https://localhost
curl: (7) Failed to connect to localhost port 443: Connection refused
damocles@Talyn:/usr/local/alancorp/alan(main)$ 
```

The first `curl -k ...` tells curl to ignore the self-signed cert and make the connection anyways, and it works. The second one doesn't have this flag, so it fails and tells you that the server has an invalid certificate. The third one repeats the first one, but shows that now it accepts no connections at all, the server has crashed (but in the other terminal you can see it still running as a zombie).

PR #444, opened 2021-03-10
Don't hang HTTPS server when TLS handshake fails

dfellis:
This resolves \#443 and big thanks to @cdmistman for the help on getting this done so quickly!


dfellis:
Proof it works: rapidly spamming "normal" and "early hangup" connections at the server and it continues working.

Screenshot from 2021-03-10 12-15-12

depombo (approved):
Nice!

PR #446, opened 2021-03-11
Update pinned alan version

aguillenv:


cdmistman (approved):

dfellis (approved):

depombo (approved):

aguillenv:
We have added the string version format to avoid manage this logic in different places.

Since for `alan deploy` anycloud cli is only used as a library, meaning no anycloud.ln need to be compiled, we just set the current alan version for the daemon to run. In other words, the alan version of the anycloud cli doesn't matter for `alan deploy` because in that case it is used as a library.

PR #449, opened 2021-03-14
Fix memory leak and remove http request opcode bottleneck

depombo:
Solves https://github.com/alantech/meta/issues/168 and https://github.com/alantech/meta/issues/171
- adding stats per process in the daemon is how I bumped into this
- we created an https hyper client on every request which created a [~100x slowdown on high throughput scenarios](https://github.com/alantech/meta/issues/168#issuecomment-798851975). Simply making a global one fixed it!
- doing async polling on http connection event which doesn't help performance, but we should do it anyways.
- restoring `block_in_place` for CPU fragments which we had dropped in https://github.com/alantech/alan/pull/406 unintentionally?
- allow the daemon to work in OS other than linux


avm/src/vm/opcode.rs
```diff
@@ -3072,29 +3065,16 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-          thread::sleep(Duration::from_millis(10));
-          waker.wake();
-        });
-        Poll::Pending
+        sleep(Duration::from_millis(10)).await;
       }
-    })
-    .await;
+    }
+    let response_hm = responses.get(&conn_id).unwrap();
```
depombo, line 3077: we should probably clear this field in `HTTP_RESPONSES` or a long running server could hit an OOM after a very long time?

depombo: confirmed this is a memory leak and created https://github.com/alantech/meta/issues/171 to track it

dfellis: It's not a memory leak, though? If you have a huge buffer of requests, it will take memory regardless of the transit mechanism.

aguillenv (approved):

cdmistman (approved):

avm/src/vm/opcode.rs
dfellis: What? We trigger a delete on send, don't we?

avm/src/vm/opcode.rs
```diff
@@ -3167,22 +3159,24 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+        let tx_raw_ptr = fractal.read_fixed(3) as *const Sender<Arc<HandlerMemory>>;
+        let tx = Arc::from_raw(tx_raw_ptr);
+        tx.send(hm).unwrap();
+        // TODO: Add a second synchronization tool to return a valid Result status, for now, just
+        // return success
+        hand_mem.init_fractal(args[2]);
+        hand_mem.push_fixed(args[2], 0i64);
+        hand_mem.push_fractal(args[2], HandlerMemory::str_to_fractal("ok"));
+        hand_mem
+      }
```
cdmistman, line 3179: which part of this is unsafe? `unsafe` should only be used for the statement(s) that require it.

avm/src/vm/opcode.rs
```diff
@@ -3167,22 +3159,24 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      drop(responses);
-      // TODO: Add a second synchronization tool to return a valid Result status, for now, just
-      // return success
-      hand_mem.init_fractal(args[2]);
-      hand_mem.push_fixed(args[2], 0i64);
-      hand_mem.push_fractal(args[2], HandlerMemory::str_to_fractal("ok"));
-      hand_mem
+      unsafe {
+        // get watch tx from the raw ptr
+        let tx_raw_ptr = fractal.read_fixed(3) as *const Sender<Arc<HandlerMemory>>;
```
cdmistman, line 3170: is this always guaranteed to be at block 3?

depombo: unless we change https://github.com/alantech/alan/pull/449/commits/2fbb1ce0c77c79fb2bd49b199a70cc6cd12ec59f#diff-aa0e5a81bac19bdacb5f06f8d0cefe2516b53e02227ac03e058f2230d02f7ba2R3055, yes

avm/Cargo.toml
```diff
@@ -5,7 +5,7 @@ license = "AGPL-3.0"
 homepage = "https://alan-lang.org"
 documentation = "https://docs.alan-lang.org"
 repository = "https://github.com/alantech/alan"
-version = "0.1.34"
+version = "0.1.34-5"
```
dfellis, line 8: Can this be reverted?

avm/src/daemon/mod.rs
```diff
@@ -1,5 +1,3 @@
-#[cfg_attr(target_os = "linux", path = "linux.rs")]
-#[cfg_attr(not(target_os = "linux"), path = "nonlinux.rs")]
```
dfellis, line 2: # :partying_face:

dfellis (approved):

avm/Cargo.toml
depombo: It can, but should it? That is actual prerelease that can be downloaded and it makes sense to know what prerelease main is on particularly for Daemon testing?

avm/src/daemon/daemon.rs
```diff
@@ -93,6 +97,24 @@ async fn post_v1_scale(cluster_id: &str, agz_b64: &str, deploy_token: &str, fact
...
 
+async fn get_procs_cpu_usage() -> Vec<f32> {
+  let mut usages = Vec::new();
+  let mut processes = processes();
+  while let Some(process) = processes.next().await {
+    if let Ok(proc) = process {
+      let measurement_1 = proc.cpu_usage().await.expect("Failed to get CPU usage for process");
+      // CpuUsage struct represents instantaneous CPU usage and
+      // does not represent any reasonable value by itself
+      sleep(Duration::from_secs(1)).await;
```
depombo, line 108: @dfellis thoughts on how long we should wait here? They recommend an absolute minimum of 100ms in the documentation.

avm/Cargo.toml
depombo: Otherwise we actually have to delete tags and preleases and push those deletions to `main` which is somewhat weird to me

avm/src/daemon/daemon.rs
dfellis: Well, a 1 sec wait is alright, but doing it sequentially like this doesn't seem good, since the time it takes to run this function will vary depending on how many processes the OS is running. Can you put the body of this loop inside of an `async move` instead and then execute them in parallel?

avm/Cargo.toml
dfellis: Well, I'm not sure if the `.version.sh` script will barf on this when we try to bump the version next. Can you confirm it won't?

I personally would prefer that the versions are just not hand edited ever to prevent potential mistakes; this still has to be tagged appropriately for it to work right, for instance, so maybe there should be a mechanism to support pre-release versioning there as an explicit thing (that bypasses the `npm publish`, amongst other things).

dfellis: > Otherwise we actually have to delete tags and preleases and push those deletions to `main` which is somewhat weird to me

Those tags will already break the moment you merge the PR since the exact commit its tagging disappears?

depombo: > > Otherwise we actually have to delete tags and preleases and push those deletions to `main` which is somewhat weird to me
> 
> Those tags will already break the moment you merge the PR since the exact commit its tagging disappears?

Ah that is true!

avm/src/daemon/daemon.rs
depombo: I know. It seems like we can't :'( Streams of futures can't be iterated over and the only concurrent method they have doesn't allow mutations. Hmm unless we put it in an Arc, let me see. https://docs.rs/futures-preview/0.2.2/futures/stream/trait.StreamExt.html#method.for_each_concurrent

depombo: That can also be in a followup PR though so that I unblock you

avm/Cargo.toml
depombo: Will put it back

avm/src/daemon/daemon.rs
dfellis: Followup PR sounds good to me.

PR #452, opened 2021-03-16
AVM Http refactoring WIP

dfellis:
Almost done with this, but I don't know how to make Rust happy with the types that were automatically resolved before.

### DIFF
--- a/avm/src/vm/opcode.rs
+++ b/avm/src/vm/opcode.rs
@@ -1,42 +1,33 @@
 use futures::future::{join_all, FutureExt};
-use futures::task::{Context, Poll};
 use std::collections::HashMap;
 use std::convert::Infallible;
 use std::fmt::Debug;
 use std::future::Future;
 use std::hash::Hasher;
-use std::io::{self, Write, BufReader};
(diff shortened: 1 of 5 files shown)


avm/src/vm/opcode.rs
```diff
@@ -1,79 +1,34 @@
...
-use rustls::internal::pemfile;
-use tokio::net::{TcpListener, TcpStream};
 use tokio::process::Command;
 use tokio::sync::watch::{self, Sender, Receiver};
 use tokio::time::sleep;
-use tokio_rustls::TlsAcceptor;
-use tokio_rustls::server::TlsStream;
 use twox_hash::XxHash64;
 
+use crate::make_server;
```
dfellis, line 25: It's weird that the path to this macro is not `crate::vm::http::make_server` @cdmistman any ideas how to make it make sense, or is this the Rust Way?

dfellis: Thanks @cdmistman :)

cdmistman (approved):

depombo (approved):
nice!

PR #453, opened 2021-03-16
get proc usage concurrently

depombo:


avm/src/daemon/daemon.rs
```diff
@@ -100,20 +101,27 @@ async fn post_v1_scale(cluster_id: &str, agz_b64: &str, deploy_token: &str, fact
...
+        },
+        Err(_) => 0.0
+      }
+    });
+  let mut futures = Vec::new();
+  while let Some(fut) = stream.next().await {
+    futures.push(fut);
   }
-  usages
+  join_all(futures).await
```
cdmistman, line 124: I believe this can all be done with `.collect().await`

depombo: What would the type of `collect` be?

cdmistman: `Vec<f32>`

cdmistman: turns out it'd be `Vec<impl Future<Output = f32>>`, so you do still need the `join_all`

depombo: yes, thanks for the help!

cdmistman (approved):

avm/src/daemon/daemon.rs
```diff
@@ -100,20 +101,25 @@ async fn post_v1_scale(cluster_id: &str, agz_b64: &str, deploy_token: &str, fact
...
-      // does not represent any reasonable value by itself
-      sleep(Duration::from_secs(1)).await;
-      let measurement_2 = proc.cpu_usage().await.expect("Failed to get CPU usage for process");
-      let usage = (measurement_2 - measurement_1).get::<ratio>();
-      usages.push(usage);
-    }
-  }
-  usages
+  let processes = processes();
+  let futures = processes
```
dfellis, line 105: Nit: can it just be `let futures = processes()...`?

dfellis (approved):
Looks great!

PR #454, opened 2021-03-16
Add the control port v0 implementation

dfellis:
Breaking this up into multiple PRs. The next one will be in the deploy service to make 4142 public and start using it for pinging, then a PR to add accepting an optional secret token to the AVM, then to get the deploy service to use that, then making it non-optional as the last PR for the initial version of the control port.

### DIFF
--- a/avm/src/vm/run.rs
+++ b/avm/src/vm/run.rs
@@ -1,3 +1,4 @@
+use std::convert::Infallible;
 use std::fs::File;
 use std::io::Read;
 use std::path::Path;
@@ -6,6 +7,7 @@ use std::sync::Arc;
 use base64;
 use byteorder::{LittleEndian, ReadBytesExt};
 use flate2::read::GzDecoder;
+use hyper::{Body, Request, Response};
 use once_cell::sync::OnceCell;
 use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
 
@@ -97,6 +99,11 @@ pub async fn run_file(fp: &str, delete_after_load: bool) {
   run(bytecode, HttpType::HTTP(HttpConfig { port: 8000, })).await;
 }
 
+async fn control_port(_req: Request<Body>) -> Result<Response<Body>, Infallible> {
+  // TODO: Add secret validation once the deploy service starts creating one for the cluster
+  Ok(Response::builder().status(200).body("ok".into()).unwrap())
+}
+
 // Used by the `daemon` mode only
 pub async fn run_agz_b64(agz_b64: &str, priv_key_b64: Option<&str>, cert_b64: Option<&str>) {
   let bytes = base64::decode(agz_b64).unwrap();
@@ -106,6 +113,12 @@ pub async fn run_agz_b64(agz_b64: &str, priv_key_b64: Option<&str>, cert_b64: Op
   let mut gz = GzDecoder::new(bytes.as_slice());
   gz.read_i64_into::<LittleEndian>(&mut bytecode).unwrap();
   if priv_key_b64.is_some() && cert_b64.is_some() {
+    // Spin up a control port if we can start a secure connection
+    make_server!(HttpType::HTTPS(HttpsConfig {
+      port: 4142, // 4 = A, 1 = L, 2 = N (sideways) => ALAN
+      priv_key_b64: priv_key_b64.unwrap().to_string(),
+      cert_b64: cert_b64.unwrap().to_string(),
+    }), control_port);
     run(bytecode, HttpType::HTTPS(HttpsConfig{
       port: 443,
       priv_key_b64: priv_key_b64.unwrap().to_string(),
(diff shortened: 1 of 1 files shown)


dfellis:
New control port working:

Screenshot from 2021-03-16 14-04-01

Screenshot from 2021-03-16 14-04-24

cdmistman (approved):

avm/src/vm/run.rs
```diff
@@ -97,6 +99,11 @@ pub async fn run_file(fp: &str, delete_after_load: bool) {
   run(bytecode, HttpType::HTTP(HttpConfig { port: 8000, })).await;
 }
 
+async fn control_port(_req: Request<Body>) -> Result<Response<Body>, Infallible> {
+  // TODO: Add secret validation once the deploy service starts creating one for the cluster
+  Ok(Response::builder().status(200).body("ok".into()).unwrap())
```
depombo, line 104: Let's make the health check smarter by checking if there are servers running in port `443` and/or port `8088`?

avm/src/vm/run.rs
```diff
@@ -106,6 +113,12 @@ pub async fn run_agz_b64(agz_b64: &str, priv_key_b64: Option<&str>, cert_b64: Op
   let mut gz = GzDecoder::new(bytes.as_slice());
   gz.read_i64_into::<LittleEndian>(&mut bytecode).unwrap();
   if priv_key_b64.is_some() && cert_b64.is_some() {
+    // Spin up a control port if we can start a secure connection
+    make_server!(HttpType::HTTPS(HttpsConfig {
+      port: 4142, // 4 = A, 1 = L, 2 = N (sideways) => ALAN
+      priv_key_b64: priv_key_b64.unwrap().to_string(),
+      cert_b64: cert_b64.unwrap().to_string(),
+    }), control_port);
```
depombo, line 121: this is very daemon specific now. let's call this function `run_daemon` or this specific logic should probably be in `daemon.rs` or we should move the entire function to `daemon.rs`

avm/src/vm/run.rs
dfellis: Since this control port is only spun up in the situation where `443` is gonna be used, that part is easy, but how do we determine at the AVM level that an Anycloud app is running and that `8088` is also relevant?

avm/src/vm/run.rs
dfellis: I do agree that this stuff feels like it should be moved into the daemon. I'm not yet sure how I want to do it, though? Can that be a refactor in a follow-up PR, since I want to fast-follow multiple PRs this afternoon on this? :)

depombo: I am okay with it happening in a follow up PR

avm/src/vm/run.rs
depombo: Currently, an alan app is an anycloud app to the daemon if the `Dockerfile` and `app.tar.gz` are present in `pwd` https://github.com/alantech/alan/blob/7245b348e9720b90921d45c6c3e623f174747502/avm/src/daemon/daemon.rs#L89

dfellis: Hmm... I can't use that particular logic since that's reading the file and then checking if that data was loaded correctly or not, but I can do something like that.

dfellis: As for testing the port, I could either just try to connect to `https://localhost` and `https://localhost:4142` and see what happens, or I can execute a shell command like this:

Screenshot from 2021-03-16 14-36-17

depombo: > Hmm... I can't use that particular logic since that's reading the file and then checking if that data was loaded correctly or not, but I can do something like that.

I know. We could generalize the "health" endpoint to provide a list of ports that we return status for all of them. Or we could just return the status for those 2 and transmit that in a similar manner to how we check for the files and transmit them.

depombo: > As for testing the port, I could either just try to connect to `https://localhost` and `http://localhost:8088` and see what happens, or I can execute a shell command like this:
> 
> Screenshot from 2021-03-16 14-36-17

I was thinking a shell command? I'm not sure how we could connect to the port if we don't know the endpoint? You mean like a ping?

dfellis: After lots of discussion, decided re-using Rust's own TCP connection logic to test is cleaner and likely faster than shell scripts.

Here's a result of the new logic:

Screenshot from 2021-03-16 15-30-21

avm/src/vm/run.rs
```diff
@@ -97,6 +100,20 @@ pub async fn run_file(fp: &str, delete_after_load: bool) {
   run(bytecode, HttpType::HTTP(HttpConfig { port: 8000, })).await;
 }
 
+async fn control_port(_req: Request<Body>) -> Result<Response<Body>, Infallible> {
+  // TODO: Add secret validation once the deploy service starts creating one for the cluster
+  if TcpStream::connect("127.0.0.1:443").is_err() {
+    Ok(Response::builder().status(500).body("fail".into()).unwrap())
+  } else if 
+    Path::new("./Dockerfile").exists() &&
+    Path::new("./app.tar.gz").exists() &&
```
depombo, line 109: not probably worth it to centralize this logic in one place, but akin to the other conversation on this PR it can happen in a follow up PR

depombo (approved):

PR #455, opened 2021-03-17
Switch to oneshot for http responses

cdmistman:
#449 introduced a potential use-after-free bug that would get triggered on the 2nd call to the `httpsend` opcode. This resolves that by instead using a `Box`.

The bug was caused by using an `Arc` to store a channel that gets used for sending `HandlerMemory`s to the http server that's running in the AVM. We store the `Arc` pointer in global memory, ensuring that it's accessible throughout the program's lifetime, but because `Arc`, `Box`, and `Rc` all use their `Drop` implementations to determine when to free their heap allocation, we have to ensure that those values don't get dropped. We could avoid this by calling [`Arc::incr_strong_count` ](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.incr_strong_count), but it's still an experimental API and we don't need to use `Arc` in the first place since its benefits over `Box` are non-existent in this use case. This PR changes that `Arc` usage to a `Box` instead to avoid the unnecessary overhead associated with `Arc`, and to avoid the use-after-free there's a call to `Box::leak` with the return value ignored to prevent the implementation of `Drop` for `Box` from getting called.


dfellis:
One quick question on all of this: How do we eventually clean it up once the HandlerMemory for the connection handler is freed?

cdmistman:
I was under the assumption that the channel would be used many times across the runtime of the program, but in discussions yesterday realized that the channel is intended to be used exactly once, so this use-after-free bug only exists when a user calls the `httpsend` opcode the 2nd time in response to a single http request. To avoid this, I'll allow the `Box` to drop instead of leaking it, and instead remove the value from the `HandlerMemory` so that this bug won't exist.

avm/src/vm/opcode.rs
```diff
@@ -3019,8 +3019,8 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
       std::process::exit(3);
     }
     // Await HTTP response from the user code
-    let response_hm = match rx.changed().await {
-      Ok(_) => rx.borrow(),
+    let response_hm = match rx.await {
+      Ok(hm) => hm,
       Err(_) => panic!("Failed to receive a response for an HTTP request"),
     };
```
dfellis, line 3025: Can the `rx` channel decide to close itself out? We may want to do something other than a `panic!` in the error path if that is the case.

dfellis (approved):
Definitely an improvement. If the answer to my question is "no" then I see no problems merging this.

avm/src/vm/opcode.rs
cdmistman: in an offline conversation we said we would return an error response (I added a TODO for that right after you left this comment).

depombo (approved):

PR #457, opened 2021-03-18
calculate time window instead of total cpu time

depombo:


avm/src/daemon/daemon.rs
```diff
@@ -123,105 +93,6 @@ async fn post_v1_scale(
...
-          // CpuUsage struct represents instantaneous CPU usage and
-          // does not represent any reasonable value by itself
-          sleep(Duration::from_secs(1)).await;
-          let measurement_2 = proc.cpu_usage().await;
-          if measurement_1.is_err() || measurement_2.is_err() {
-            return 0.0;
-          }
-          let usage = measurement_2.unwrap() - measurement_1.unwrap();
-          usage.get::<ratio>()
-        }
```
depombo, line 141: this cpu usage per process calculation was bogus

avm/src/daemon/stats.rs
```diff
@@ -0,0 +1,176 @@
...
+          if cpu_1.is_err() || cpu_2.is_err() {
+            return 0.0;
+          }
+          let times_1 = cpu_1.unwrap();
+          let times_2 = cpu_2.unwrap();
+          let system = times_2.system().get::<second>() - times_1.system().get::<second>();
+          let user = times_2.user().get::<second>() - times_1.user().get::<second>();
+          (user + system) / duration
+        }
+        Err(_) => 0.0,
```
depombo, line 59: this cpu usage per process calculation does the right thing and replaces the one in https://github.com/alantech/alan/pull/457/files#r597315536

dfellis: It definitely looks good, but why divide by the duration if it is 1.0?

depombo: oh true. I did that just in case we changed this value to a duration other than one second

dfellis (approved):

aguillenv (approved):

cdmistman (approved):

PR #458, opened 2021-03-18
Refactor the daemon-related code to be more self-contained in the daemon module, and add a secret string for use in guarding access to the control port

dfellis:


dfellis:
Spinning up a daemon with a secret:

Screenshot from 2021-03-18 09-45-04

The control port now blocking requests that don't provide that secret in the headers:

Screenshot from 2021-03-18 09-45-23

avm/src/daemon/daemon.rs
```diff
@@ -221,6 +233,60 @@ async fn post_v1_stats(cluster_id: &str, deploy_token: &str) -> String {
...
+  let agz = GzDecoder::new(bytes.as_slice());
+  let count = agz.bytes().count();
+  let mut bytecode = vec![0; count / 8];
+  let mut gz = GzDecoder::new(bytes.as_slice());
+  gz.read_i64_into::<LittleEndian>(&mut bytecode).unwrap();
+  if priv_key_b64.is_some() && cert_b64.is_some() {
+    // Spin up a control port if we can start a secure connection
+    make_server!(
+      HttpType::HTTPS(HttpsConfig {
+        port: 4142, // 4 = A, 1 = L, 2 = N (sideways) => ALAN
```
cdmistman, line 270: image

aguillenv (approved):

cdmistman (approved):

avm/src/daemon/daemon.rs
```diff
@@ -213,14 +225,75 @@ async fn get_v1_stats() -> VMStatsV1 {
...
+    // TODO: Any way to generalize this so we don't have special logic for Anycloud?
+    Ok(Response::builder().status(500).body("fail".into()).unwrap())
+  } else {
+    // Everything passed, send an ok
+    Ok(Response::builder().status(200).body("ok".into()).unwrap())
+  }
+}
+
+// Used by the `daemon` mode only
+pub async fn run_agz_b64(agz_b64: &str, priv_key_b64: Option<&str>, cert_b64: Option<&str>) {
```
depombo, line 266: well we can just remove the comment and the `pub`? ;)

dfellis: Oh, right. Will do.

avm/src/daemon/daemon.rs
```diff
@@ -19,7 +27,11 @@ use tokio::time::{sleep, Duration};
 
 use crate::daemon::dns::DNS;
 use crate::daemon::lrh::LogRendezvousHash;
-use crate::vm::run::run_agz_b64;
+use crate::make_server;
+use crate::vm::http::{HttpConfig, HttpType, HttpsConfig};
+use crate::vm::run::run;
+
+pub static SECRET_STRING: OnceCell<Option<String>> = OnceCell::new();
```
depombo, line 34: nit should we maybe call this `CLUSTER_TOKEN`? `SECRET_STRING` is accurate, but it also doesn't say much about the actual use case of the secret?

dfellis: So `cluster_token` is very confusing with `deploy_token`, which is why I changed that. It also does have multiple uses; it's not just for the inter-cluster-communication but verifying `deploy` to cluster communication, and potentially could be used in other ways, too?

dfellis: Actually, I do have a question on that front. What are we using the `deploy_token` for that couldn't be handled here? I wouldn't mind calling it the `cluster_token` if we can just eliminate the `deploy_token`?

depombo: I don't think haven two tokens is confusing. They have different use cases, but more of a nit. `deploy` to cluster communication still falls under correct usage or proper naming for something called cluster token to me. The deploy token is given from `deploy` to the `daemon` and then back (since `deploy` is stateless) to the provide the deployment configuration (credentials, region, vmType) to `deploy` when it needs to add or remove more VMs to the cluster

depombo: And the deploy token was deemed unusable for authentication since it can change during an upgrade

depombo: Maybe `CLUSTER_SECRET`?

dfellis: I like `CLUSTER_SECRET`

dfellis: > I don't think having two tokens is confusing? They have different use cases, but more of a nit. `deploy` to cluster communication still falls under correct usage or proper naming for something called cluster token to me. The deploy token is given from `deploy` to the `daemon` and then back (since `deploy` is stateless) to the provide the deployment configuration (credentials, region, vmType) to `deploy` when it needs to add or remove VMs from the cluster

I forgot that the `deploy_token` was a signed, encrypted payload of stuff to make upgrades possible.

And in that case keeping `secret` in the name of this variable makes sense because it's not encrypting anything.

avm/src/daemon/daemon.rs
```diff
@@ -213,14 +225,74 @@ async fn get_v1_stats() -> VMStatsV1 {
...
+      .insert("clusterSecret".to_string(), json!(cluster_secret));
+  }
   post_v1("stats", stats_body).await
 }
 
+async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible> {
+  let cluster_secret = CLUSTER_SECRET.get().unwrap();
+  if cluster_secret.is_some() && !req.headers().contains_key(cluster_secret.as_ref().unwrap()) {
+    // If this control port is guarded by a secret string, make sure there's a header with that
+    // secret as the key (we don't care about the value) and abort otherwise
```
depombo, line 247: so the key to the header will be the `CLUSTER_SECRET` and the value empty?

avm/src/main.rs
```diff
@@ -81,6 +81,7 @@ fn main() {
       .arg_from_usage("<DOMAIN> 'Specifies the application domain'")
       .arg_from_usage("-k, --private-key=[PRIV_KEY_B64] 'An optional base64 encoded private key for HTTPS mode'")
       .arg_from_usage("-c, --certificate=[CERT_B64] 'An optional base64 encoded certificate for HTTPS mode'")
+      .arg_from_usage("-s, --cluster-secret=[SECRET_STRING] 'An optional secret string to constrain access to the control port'")
```
depombo, line 84: `CLUSTER_SECRET`?

depombo (approved):

avm/src/daemon/daemon.rs
depombo: Yeah, I also prefer `CLUSTER_SECRET` to `CLUSTER_TOKEN` for that reason

avm/src/daemon/daemon.rs
dfellis: Sorta. I think you have to set it to *something* so I set it to `true` in my test above. We could eventually rotate the secret, for instance? We could have it be `CURRENT_SECRET: CURRENT_SECRET` normally and `CURRENT_SECRET: REPLACEMENT_SECRET` if/when we want to rotate the secret out.

That will require replacing `OnceCell` with something else, but something to think about on the security front. The shorter-lived a secret, the harder it is to do anything with it.

avm/src/main.rs
dfellis: Missed one!

avm/src/daemon/daemon.rs
cdmistman: it still works with `OnceCell` <https://docs.rs/once_cell/1.7.2/once_cell/sync/struct.OnceCell.html#method.take>

dfellis: > Sorta. I think you have to set it to _something_ so I set it to `true` in my test above. We could eventually rotate the secret, for instance? We could have it be `CURRENT_SECRET: CURRENT_SECRET` normally and `CURRENT_SECRET: REPLACEMENT_SECRET` if/when we want to rotate the secret out.
> 
> That will require replacing `OnceCell` with something else, but something to think about on the security front. The shorter-lived a secret, the harder it is to do anything with it.

Though when we want to dig into secret rotation, we should think hard on the implications of the API. It would be useful to be able to obfuscate *how* the rotation works, too, if that is somehow possible.

dfellis: > it still works with `OnceCell` https://docs.rs/once_cell/1.7.2/once_cell/sync/struct.OnceCell.html#method.take

For certain definitions of "once," it seems. 🤨

depombo (approved):

PR #459, opened 2021-03-19
Logger

aguillenv:
Tests should pass once we merge https://github.com/alantech/anycloud/pull/32


dfellis:
@aguillenv you'll need to update the `Cargo.lock` to the latest anycloud once the other PR is landed before any of these tests can pass: https://github.com/alantech/alan/pull/459/checks?check_run_id=2185186570

PR #462, opened 2021-03-19
error handling for http server synchronization

depombo:


avm/src/vm/opcode.rs
```diff
@@ -3063,9 +3065,8 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
         hm.register_from_fractal(i as i64, &res_out, i);
       }
       // Get the watch channel tx from the raw ptr previously generated in http_listener
-      let fractal = hand_mem.read_mut_fractal(args[0]);
-      // TODO: avoid the panic if the value doesn't exist and return an error
-      let tx_raw_ptr = fractal.remove(3).1 as *mut Sender<Arc<HandlerMemory>>;
+      let fractal = hand_mem.read_fractal(args[0]);
+      let tx_raw_ptr = fractal.read_fixed(3) as *mut Sender<Arc<HandlerMemory>>;
```
cdmistman, line 3069: this reverts the use-after-free bug, but this can be solved in other ways and will be fixed in a future PR (as well as removing the `unsafe` on the next line of code)

cdmistman (approved):

dfellis (approved):
Let's try to get the double-use bug fixed ASAP, too.

avm/src/vm/opcode.rs
depombo: Yes, tracked it here https://github.com/alantech/alan/issues/463

Issue #463, opened 2021-03-19
Investigate potential memory issues around http server responses that never calls `send`


depombo:
The current http server implementation will probably give rise to potential memory or security issues if an alan program ~calls `send` more than once or if it~ never calls `send` on response for the same HTTP server connection. We should write a few test alan servers and see how the HTTP opcodes fail and properly handle this.

PR #464, opened 2021-03-19
Fix double-calling `send` on an http server from crashing the AVM and js-runtime

dfellis:
This fixes half of \#463. The other half is a memory leak if `send` is *never* called, but that will require a much more clever solution to tackle.

### DIFF
--- a/avm/src/vm/opcode.rs
+++ b/avm/src/vm/opcode.rs
@@ -6,6 +6,7 @@ use std::future::Future;
 use std::hash::Hasher;
 use std::io::{self, Write};
 use std::pin::Pin;
+use std::ptr::NonNull;
 use std::str;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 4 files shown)


avm/src/vm/opcode.rs
```diff
@@ -3055,36 +3056,35 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      let tx: Box<Sender<Arc<HandlerMemory>>> = unsafe { Box::from_raw(tx_raw_ptr) };
+  cpu!(httpsend => fn(args, hand_mem) {
+    hand_mem.dupe(args[0], args[0]); // Make sure there's no pointers involved
+    let mut hm = HandlerMemory::new(None, 1);
+    HandlerMemory::transfer(&hand_mem, args[0], &mut hm, CLOSURE_ARG_MEM_START);
+    let res_out = hm.read_fractal(CLOSURE_ARG_MEM_START);
+    for i in 0..res_out.len() {
+      hm.register_from_fractal(i as i64, &res_out, i);
+    }
+    // Get the watch channel tx from the raw ptr previously generated in http_listener
```
cdmistman, line 3067: oops!
```suggestion
    // Get the oneshot channel tx from the raw ptr previously generated in http_listener
```

cdmistman (approved):

bdd/spec/023_http_spec.sh
```diff
@@ -201,4 +201,59 @@ correct"
...
+        from @std/app import print, exit
+        from @std/httpserver import connection, Connection, body, send
+
+        on connection fn (conn: Connection) {
+          const res = conn.res;
+          const firstMessage = res.body('First Message').send();
+          print(firstMessage);
+          const secondMessage = res.body('Second Message').send();
+          print(secondMessage);
+          emit exit 0;
```
depombo, line 217: 🥳

depombo (approved):

PR #465, opened 2021-03-23
More Stringifiable Definitions

### LINKED ISSUE #311: Make `Array<Stringifiable>`, `Maybe<Stringifiable>`, etc into `Stringifiable`, too
Make it easier to print out arrays, optional values, etc by implementing `toString` for these types, similar to how `Result<Stringifiable>` has been handled.

cdmistman:
fixes #311

### DIFF
--- a/avm/src/vm/opcode.rs
+++ b/avm/src/vm/opcode.rs
@@ -3563,7 +3563,7 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
     }
     None
   });
-  cpu!(getM => fn(args, hand_mem) {
+  cpu!(getMaybe => fn(args, hand_mem) {
     let fractal = hand_mem.read_fractal(args[0]);
     let variant = fractal.read_fixed(0);
     if variant == 1 {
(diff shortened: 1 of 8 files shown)


avm/src/vm/opcode.rs
```diff
@@ -3563,6 +3563,16 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
     }
     None
   });
+  cpu!(getM => fn(args, hand_mem) {
```
depombo, line 3566: why not `getMaybe`? Outside the context of this PR, it would be hard to grasp the difference between `getM` and `getMain` or what the `M` in `getM` stands for

dfellis (approved):

bdd/spec/015_maybe_result_either_spec.sh
```diff
@@ -140,21 +152,33 @@ Divide by zero error!
...
             print('uhhh');
           }
+
+          strOrNum.toString().print();
+          strOrNum2.toString().print();
+
           emit exit 0;
         }
+
+        fn getMainOrAlt(isMain: bool) {
```
dfellis, line 175: I'm surprised declaring this function after it is used works, tbh.

dfellis (approved):

depombo (approved):
LGTM once tests pass!

PR #466, opened 2021-03-24
RFC for lnn (new compiler front-end)

cdmistman:


rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,179 @@
...
+
+Also, "constraint" is used in literature related to Hindley-Milner type checking to signify that a narrowing-down on the type of a value, expression, or function.
+A type can only be selected if it matches constraints, and if there are constraints that conflict with each other, that results in an invalid program.
+
+The initial motivating example is as follows:
+```ln
+from @std/app import print, start, exit
+
+on start {
+  const myMaybe: Maybe<int64> = none();
```
dfellis, line 57: Correct me if I'm wrong, but this particular example would not compile without the type hint by the user, correct?

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,179 @@
...
+  return add(num, 4);
+}
+```
+
+At the time of writing this RFC, this function fails to parse due to a missing type annotation on `num`, but it's intended to support gradual typing even for function parameters in Alan.
+The current front-end will require substantial work to support this, requiring one of:
+- An interface (internal or in `root.ln`) that ensures there's an `add` operation for the type of `num` (for example, `Addable`).
+- Type-checking can only happen once the function is inlined.
+
+However, the new type resolution system would simply add a constraint that there be a function named `add` that satisfies the signature `(T, 4 as T) -> T` (ignoring the fact that arithmetic operations in Alan actually use `Result` values).
```
dfellis, line 112: I don't see how the `4 as T` is determined by the compiler? Since there is an add with the signature: `fn add(Result<int64>, int64): Result<int64>` already in the current codebase being able to assume that `num` and `4` are the same type does not seem correct?

But I don't think that affects the possibility of this particular example working or not.

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,179 @@
...
+condfn(conditionalTable, condition1Boolean, condition1ThenFn);
+const condition2Boolean = true;
+const condition2ThenFn = fn(): U { return 'none'; };
+condfn(conditionalTable, condition2Boolean, condition2ThenFn);
+const conditionTailFn = fn(): U {};
+const conditionResult: Maybe<U> = evalcond(conditionalTable, conditionTailFn);
+const conditionResultUnwrapped: U = conditionResult.getMaybe();
+return conditionResultUnwrapped;
+```
+3. Add the constraint that `val` has type `A` where there's a function called `isSome` with the signature `(A) -> bool`.
```
dfellis, line 143: I also don't see how the `(A) -> bool` is determined here without some deep knowledge of how conditionals work. That could be fine, but wouldn't it also function it just determined a signature `(A) -> X` where X is just some other type, and it just happens to be `bool` since there is no other implementation of `isSome` in the standard library.

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,179 @@
...
+    return 'none';
+  }
+}
+```
+
+In order to share more nuanced details about the new front-end, here are the steps the compiler might take upon having to use this function:
+1. Assign the variable `val` to type `T`. There are no constraints on the type `T`.
+2. Desugar the if/else. How this is done is unimportant, but the body of the `toString` function will effectively be desugared into (ignore the invalid `conditionTailFn` function body - it's required for `evalcond` to work but its contents won't be checked by the type checker, since this desugaring will actually happen just before generating the Microstatements):
+```ln
+const conditionalTable = newarr(2);
```
dfellis, line 131: Nit: while `const` does work alright now because of other bugs in the existing compiler, this should probably be a `let` here. :)

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+At the time of writing this RFC, this function fails to parse due to a missing type annotation on `num`, but it's intended to support gradual typing even for function parameters in Alan.
+The current front-end will require substantial work to support this, requiring one of:
+- An interface (internal or in `root.ln`) that ensures there's an `add` operation for the type of `num` (for example, `Addable`).
+- Type-checking can only happen once the function is inlined.
+
+However, the new type resolution system would simply add a constraint that there be a function named `add` that satisfies the signature `(T, 4 as T) -> T` (ignoring the fact that arithmetic operations in Alan actually use `Result` values).
+This wouldn't require the `Addable` interface to be defined at all, and instead would allow for an anonymous interface that's only visible in the compiler to be created.
+The constraints would then not need to be computed for each time the function is called - instead, we must only ensure that the value passed in to `add4` satisfies the so-called "Addable" interface.
```
dfellis, line 114: I do have to say, these constraints seem very similar to the [automatic argument interfaces](https://github.com/alantech/alan/blob/main/rfcs/006%20-%20Automatic%20Argument%20Interfaces%20RFC.md) proposed earlier. It would be nice to:

1. Compare/contrast with that already existing RFC. Does this revoke/supersede it?
2. Can interfaces themselves been treated similarly to constraints such that we can have the compiler generate interfaces for the user based on function usage and inject them into source code through a code formatting tool?

I actually haven't seen anything in this RFC that implies that constraints and interfaces are actually different, except one being created by the compiler and the other the developer.

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,179 @@
...
+10. Since `A` and `B` both are in reference to the same variable, add the constraint that they satisfy each other.
+11. Reduce type `Maybe<R>` to be equivalent to the type `Maybe<Stringifiable>`
+12. `Maybe<Stringifiable>` is the minimal interface value for `val`, so the signature of the `toString` fn is now rewritten as `(Maybe<Stringifiable>) -> U`.
+13. `string` is the concrete type representation for `U`, so the signature of the `toString` fn is now rewritten as `(Maybe<Stringifiable>) -> string`.
+
+(side note: the `getMaybe()` call will also eventually be syntactic sugar, according to [RFC 12][RFC-12]).
+
+### Alternatives Considered
+
+One alternative is to keep the current type system and compiler front-end and adding changes as-needed.
```
dfellis, line 159: There are a lot of parts of the current compiler that may still be worth keeping around, though. How module resolution and import/export works is quite a bit of logic to recreate.

I would also argue that the RFC as-is is incomplete. There's a brief statement about opting into the new frontend by naming the file `.lnn` but how does it fit in with the rest of the compiler, how does it do AST parsing, the above-mentioned module resolution, represent scoping, what the intermediate types are (equivalents to Interface/Type/UserFunction/etc) and then whether or not the microstatement generation that is then serialized to AMM is performed or if the Handlers convert themselves to AMM which triggers the UserFunction within to do so which triggers the Statement within, etc, and then if/how constant hoisting and deduplication is handled.

All of that is *a lot* that the current first stage of the compiler does, and throwing it all away for just the type system is a hard sell.

I think another option to consider is "forking" the current `lntoamm` directory by copying it into an `lnntoamm` directory, ripping out the current type system and replacing it.

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,179 @@
...
+However, as much of the current compiler already performs well, that would not be very productive and would just be reinventing the wheel, for the most part.
+
+[cond-table-PR]: https://github.com/alantech/alan/pull/430
+[emlti]: http://pauillac.inria.fr/~fpottier/publis/emlti-final.pdf
+[haskell-type-system-revered]: https://softwareengineering.stackexchange.com/questions/279316/what-exactly-makes-the-haskell-type-system-so-revered-vs-say-java
+[tech-debt-issues]: https://github.com/alantech/alan/labels/tech%20debt
+
+## Affected Components
+
+A brief listing of what part(s) of the language ecosystem will be impacted should be written here.
```
dfellis, line 174: Missed this part.

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Expected Timeline
+
+An RFC proposal should define the set of work that needs to be done, in what order, and with an expected level of effort and turnaround time necessary. *No* multi-stage work proposal should leave the language in a non-functioning state.
```
dfellis, line 178: Same here.

rfcs/015 - Reworked Compiler Front End.md
cdmistman: That's correct, I only used `none` instead of `some` to avoid walking through argument passing since I wanted a simpler example first.

rfcs/015 - Reworked Compiler Front End.md
cdmistman: Ah, you're right - I'll fix that

rfcs/015 - Reworked Compiler Front End.md
cdmistman: I guess it would be `(A) -> X` where `X` is constrained to be a `bool`, I left that part out since it's part of the same step but I'll clarify.

rfcs/015 - Reworked Compiler Front End.md
cdmistman: Gotcha - will add that it will be effectively a fork of the current `lntoamm` directory with parsing and module resolution kept intact, but with a new Microstatement generation system that follows this new type system.

rfcs/015 - Reworked Compiler Front End.md
cdmistman: I've intentionally kept the user-facing part of the type system the same, to prevent code breakages as we migrate the compiler to the new front-end. It's meant to be compatible with the current intended design of the language, including RFC 6.

dfellis: Ahh, can you elaborate on that? What part of the type system is "submerged"?

cdmistman: I'm effectively proposing a middle stage in the compiler front-end that does type resolution and function selection *before* emitting Microstatements in the output. I've already laid out some of the advantages in the RFC, but another notable advantage is that inferring argument types is separate from the inlining stage - with the current system, it'd only be possible for arguments to be type-checked once the function is being inlined, resulting in the compiler having to type-check the same code `n` times for each time the function is called. In the proposed system, the function's type signature need only be generated once, all of its statements need be type-checked once, all of its function selection possibilities need be gathered once, and Microstatement generation need only happen once per function use.

To clarify, this is not a change to the type system of Alan itself, it's a change to the type resolution and inference system in the front end of the compiler. Users shouldn't experience any changes - instead, there will be benefits in other things, such as compilation speed, code maintainability, code modularity, and a familiar algorithm for many. Since this is a new "phase" in the compiler, that means that the Microstatement generation will also have to be rewritten, but it will be much simpler since most of it will already be generated, it'll simply be a matter of inlining and outputting the correct AMM once that happens.

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,183 @@
...
+13. `string` is the concrete type representation for `U`, so the signature of the `toString` fn is now rewritten as `(Maybe<Stringifiable>) -> string`.
+
+(side note: the `getMaybe()` call will also eventually be syntactic sugar, according to [RFC 12][RFC-12]).
+
+### Alternatives Considered
+
+One alternative is to keep the current type system and compiler front-end and adding changes as-needed.
+However, the current front-end's design means that it takes relatively substantial work to add each feature, with [new features potentially exposing "load-bearing" bugs][cond-table-PR].
+Additionally, there is [a lot of documented tech-debt][tech-debt-issues] that would already result in substantial parts of the compiler being reworked or rewritten.
+Lastly, the Hindley-Milner type system has been well-documented and proven both [over time][haskell-type-system-revered] and [in theory][emlti].
```
depombo, line 164: I'm not familiar with them, but are there other type systems that should be considered here?

rfcs/015 - Reworked Compiler Front End.md
depombo: The fork of the current `lntoamm` makes a lot of sense

rfcs/015 - Reworked Compiler Front End.md
cdmistman: There are alternatives, but only the HM scheme has been proven to not require type annotations anywhere in order to produce a valid program.
- C++/Java-style type inference is rather weak and fails in many cases - they're unable to adapt more complex algorithms due to subtyping in their system (class extensions)
  - Dart has subtyping as well and its type inference is more extensive, but [still has restrictions](https://github.com/dart-lang/sdk/issues/37247)
  - Swift has [similar issues](https://cs.stackexchange.com/questions/26389/why-isnt-the-swift-programming-language-type-inference-more-aggressive)
- Rust is [a modified version of Hindley-Milner](https://rustc-dev-guide.rust-lang.org/type-inference.html)
- Kotlin also uses [a partial version Hindley-Milner](https://kotlinlang.org/spec/type-inference.html), but again subtyping means that type annotations are needed in certain parts of the language
- TypeScript uses a [candidate system similar to the current front-end, but requires multiple passes to correctly determine the type of an object](https://github.com/microsoft/TypeScript/issues/2182) - and even then, it can't determine the concrete types of values all the time, [defaulting to `any` in many scenarios where it can't determine the type of an object](https://basarat.gitbook.io/typescript/type-system/type-inference)
- Golang [uses the C/C++/Java-style type inference (despite not having sub-typing)](https://stackoverflow.com/questions/12589255/gos-type-inference-algorithm), but its otherwise simple syntax and obligatory type declarations in function syntax or for interfaces/structs, etc means that it's still quite effective
- Typed Racket [is Hindley-Milner](https://docs.racket-lang.org/plai-typed/Type_Checking_and_Inference.html)

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,183 @@
+# 000 - RFC Template
```
dfellis, line 1: Can this be fixed and switched to `016 - Reworked Compiler Front End RFC` and the file moved to `016 - Reworked Compiler Front End RFC.md`

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,183 @@
...
+
+2020-03-24
+
+### Accepted
+
+YYYY-MM-DD
+
+#### Approvers
+
+- Full Name <email@example.com>
```
dfellis, line 15: And add @depombo and I here :)

rfcs/015 - Reworked Compiler Front End.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Current Status
+
+### Proposed
+
+2020-03-24
+
+### Accepted
+
+YYYY-MM-DD
```
dfellis, line 11: And set this to today's date. We won't let this go on another day. :)

dfellis (approved):

depombo (approved):

PR #467, opened 2021-03-25
TCP RFC

dfellis:


rfcs/015 - TCP RFC.md
```diff
@@ -0,0 +1,196 @@
...
+}
+
+on connError fn (ctx: Context<Connection>) {
+  ctx.context.close();
+}
+
+on connTimeout fn (ctx: Context<Connection>) {
+  ctx.context.close();
+}
+```
```
cdmistman, line 71: What about listening on multiple ports, or closing down specific ports? It seems that this implementation still requires the remote to only connect to 1 port, at which point the Alan code will move the connection to a different port, which doesn't seem very flexible.

For example, say Alice wanted to implement a web server in Alan that allowed for both HTTP and SSH connections (just using them because I'm familiar with the packet specifications of both). To do this, she can only write the above, except for this change:
```ln
on connection fn (conn: Connection) {
  const c: Chunk = conn.read();
  const tunnel = c.getTunnel();
  if tunnel.isSome() {
    conn.addContext(tunnel);
    tunnel.addContext(conn);
  } else {
    conn.close();
  }
}
fn getTunnel(conn: Connection) {
  const chunk = conn.read();
  if chunk.startsWith('HTTP') {
    return some(connect('localhost', 8088));
  } else if chunk.startsWith('SSH') {
    return some(connect('localhost', 2222));
  } else {
    return none();
  }
}
```

dfellis: First, I attempted to address that here: https://github.com/alantech/alan/pull/467/files#diff-d20ad24b805e5d94f7929b3e2aed48bdad3f4613a80aea17d0b41b4346710665R73

> Whether or not that autoselection (of the listen port) sticks around long-term is still up in the air.

I was explicitly not addressing the situation you bring up as out-of-scope here. It *should* be possible to do something like telling the AVM what `tcp` ports (plural) to listen to and then the `conn` object has a `port` field that you can key on (or maybe a `serverNum` so the port itself is opaque but "first server" vs "second server" is clear in your code) that can then be tackled. And this doesn't seem like it would require ripping everything of this design out, just something that wouldn't be tackled at first.

For the exact situation where the same port handles both HTTP and SSH connections, since `addContext` *can* be called in the `chunk` events as described further down: https://github.com/alantech/alan/pull/467/files#diff-d20ad24b805e5d94f7929b3e2aed48bdad3f4613a80aea17d0b41b4346710665R89

That means that the user could create a type where one of the fields is `Maybe<Connection>` and the other is `Maybe<Array<Chunk>>` (or something) and spinning up the correct child connection for an HTTP server or an SSH server can be deferred until later, and the `chunk` handler function just checks whether the context's `Maybe<Connection>` is defined or not for whether or not it will write the chunk to an established connection, or attempt to probe which kind of connection it is. If it's streaming in, it can also check if `Maybe<Array<Chunk>>` is defined or not to first stream the already buffered array of chunks in before streaming the current chunk, or if it's just going to write the current chunk only.

This also makes it invulnerable to the possible but unlikely situation that the client is super slow and the first chunk you get is just "HT", for instance, and you can't identify it by the first chunk of data alone.

So it wouldn't need to have access to the `read` function during establishing the connection (nor would that be expected to already have a chunk). The current design could work for that scenario, too.

cdmistman: Gotcha, makes sense. Another question I've thought of, but is rather out-of-scope for this RFC: how will `@std/http` know which port to listen to depending on usage in `anycloud` or not? We could define the default port to 8088, but what if users want to listen to port 80 instead?

dfellis: So an Alan application itself is not deployed through Anycloud. It deploys directly with `alan deploy` and this problem is moot. The only HTTP server is actually listening on port 443 as an HTTPS server with a self-signed cert that is connected to Cloudflare where the real cert lives.

dfellis: If you just run an Alan application with `alan run myapp.agz` it will listen on port 8000, intentionally wrong for Anycloud to discourage trying to needlessly wrap your Alan app in Docker and another Alan app.

rfcs/015 - TCP RFC.md
```diff
@@ -0,0 +1,196 @@
...
+
+While changes to the language are on the table (and some of the alternatives will cover a few of them), the best overall solution (considering implementation complexity and cognitive complexity) doesn't need it. :)
+
+The clearest way to introduce the new syntax is to simply use it in an example similar to how it will be used by Anycloud:
+
+```ln
+from @std/tcpserver import connection, Connection
+from @std/tcp import connect, addContext, chunk, Chunk, Context, read, write, connClose, close, connError, connTimeout
```
depombo, line 47: is there a way to consolidate the imports or make them more intuitive / less error-prone? It seems like `connect` returns a `connection` that one can call `addContext` on, but `connection` belongs to `tcpserver` unlike the rest which belong to `tcp`. also it might be more clear to just rename to `tcpclient` and `tcpserver`

rfcs/015 - TCP RFC.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Expected SemBDD Impact
+
+This would be a minor update if Alan was post-1.0.0 as it adds new functionality without regressing on existing functionality.
+
+## Proposal
+
+A TCP connection has four essential states: open, closed-on-purpose, closed-on-error, and closed-by-timeout. While in the open state, data can be read from its output, written to its input, or explicitly closed as triggers by the user, while it can also unexpectedly transition to the other two closed states. The amount of data received at any particular point in time is completely arbitrary and up to both the originating machine and the local kernel, which makes it a good fit for the event system so the users' code is prompted when each chunk is ready to work with that it can continue with.
```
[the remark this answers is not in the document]
dfellis: so, the chunk itself doesn't do any work, but I'll see if I can make that clearer? "the chunk is ready (for you to) work with" is what is implied, and that may be an implication only native speakers pick up?

rfcs/015 - TCP RFC.md
dfellis: The only thing that needs to be imported from `@std/tcpserver` is the actual `connection` event, not even the `Connection` type, which could also be from the regular `@std/tcp` but I thought that looked a bit weird? Would it be less weird to you if everything is imported from `@std/tcp` and only the `connection` event is isolated in `@std/tcpserver`?

depombo: That would be seem less weird to me

rfcs/015 - TCP RFC.md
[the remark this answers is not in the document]
depombo: ah got it. yeah, that might be something I didn't pick up as ESL. definitely a nit so no worries

rfcs/015 - TCP RFC.md
```diff
@@ -0,0 +1,196 @@
...
+  ctx.context.close();
+}
+
+on connError fn (ctx: Context<Connection>) {
+  ctx.context.close();
+}
+
+on connTimeout fn (ctx: Context<Connection>) {
+  ctx.context.close();
+}
```
dfellis, line 70: I am tempted to merge these together anyways, even if properly checking the reason for closing isn't always done.

Why? Because right now the user could forget to handle one or more of the close events at all right now, anyways, and the compiler won't complain that you aren't listening to one of these events, while having fewer events makes things simpler and shorter.

dfellis: Reduces down to:

```ln

```ln
from @std/tcpserver import connection
from @std/tcp import Connection, connect, addContext, chunk, Chunk, Context, read, write, connClose, close

on connection fn (conn: Connection) {
  const tunnel = connect('localhost', 8088);
  conn.addContext(tunnel);
  tunnel.addContext(conn);
}

on chunk fn (ctx: Context<Connection>) {
  const c: Chunk = ctx.conn.read();
  ctx.context.write(c);
}

on connClose fn (ctx: Context<Connection>) {
  ctx.context.close();
}
```

rfcs/015 - TCP RFC.md
```diff
@@ -0,0 +1,196 @@
...
+
+And that's it. Once constructed, the AVM will keep it resident and emit events involving that connection. Those events get the `Context<T>` type, where the `T` is the type that was added to the context. One known vulnerability with this as-is is that the client and server context types need to match each other and the compiler will not be able to keep this completely safe. This could potentially be avoided by having completely separate events for client and server, but that also nearly doubles the size of the codebase so it is being avoided for now. This can also be worked around with an `Either<ServerType, ClientType>` that is checked by the handlers which is why this is not considered a huge problem.
+
+The `Context<T>` type has two properties, the `conn: Connection` that corresponds to the connection that the triggered the event, and the `context: T` property that is whatever value was attached to the connection, which in this case is another connection.
```
depombo, line 79: This part is what has me thinking/hesitant the most. The context seems like completely generic metadata that can be attached to any struct and not necessarily just a `connection`? I'm not sure if we should take the `context` concept to one of the opposite ends: either allow some sort of or arbitrary metadata to be attached to all structs or make this a special property of connection wherein one can associate two connections with one another.

I also keep wondering if it's possible to achieve this in a simpler way, without contention, through some sort of struct composition but it is not very clear to me how it would work.

dfellis: So we can call it `TcpContext` if that makes you more comfortable? I didn't want to go all the way to a [Go Context type](https://golang.org/src/context/context.go?s=2460:6019#L52) (though I did get some inspiration there) because with that the values attached to the context are stored in a hashmap that requires constant unsafe casting from `interface {}` to work with, so instead I went with a generic type for a singular value, which you could make a `HashMap<string, whatever>` but tried to minimize the "blast radius" of casting here.

The only other way it could potentially work out is with `@std/datastore`, shoving Context-like values in there and then pulling them out where necessary, but since datastore is intended to be replicated across the cluster it would be a massive performance bottleneck for manipulating raw TCP streams to require a full TCP request/response under the hood to get the relevant metadata.

Some sort of "forced-local" datastore model could be a compromise there, but still seems like a lower performance than the connection itself holding onto the context that will be needed.

depombo: > So we can call it `TcpContext` if that makes you more comfortable? I didn't want to go all the way to a [Go Context type](https://golang.org/src/context/context.go?s=2460:6019#L52) (though I did get some inspiration there) because with that the values attached to the context are stored in a hashmap that requires constant unsafe casting from `interface {}` to work with, so instead I went with a generic type for a singular value, which you could make a `HashMap<string, whatever>` but tried to minimize the "blast radius" of casting here.
> 
> The only other way it could potentially work out is with `@std/datastore`, shoving Context-like values in there and then pulling them out where necessary, but since datastore is intended to be replicated across the cluster it would be a massive performance bottleneck for manipulating raw TCP streams to require a full TCP request/response under the hood to get the relevant metadata.
> 
> Some sort of "forced-local" datastore model could be a compromise there, but still seems like a lower performance than the connection itself holding onto the context that will be needed.

ah that makes a lot of sense. particularly why we are not considering with a full on `Context type`. I agree using datastore or backing it up with a HashMap would be too much overhead.

I think calling it `TcpContext` might make it easier to follow. What about having a parent `ConnectionPair` that connects both. Or perhaps getting closer to the semantics of a proper channel where a `TcpChannel` has a `tx` and `rx`?

dfellis: Keeping the `Connection` an opaque type was a goal so we can also treat it as a pointer under the hood like we did for the HTTP server in the AVM. Splitting the object into `tx` and `rx` inner properties would be possible, but would be a fabrication here that would also introduce overheads we don't really need? `.read()`ing and `.write()`ing a `Connection` is still very clear for the end users.

Renaming `Connection` to `TcpChannel` would be alright by me, though, if it makes more sense to you that way? Then `TcpContext` would have `channel` and `context` properties?

depombo: > Keeping the `Connection` an opaque type was a goal so we can also treat it as a pointer under the hood like we did for the HTTP server in the AVM. Splitting the object into `tx` and `rx` inner properties would be possible, but would be a fabrication here that would also introduce overheads we don't really need? `.read()`ing and `.write()`ing a `Connection` is still very clear for the end users.
> 

I see. We definitely don't want to introduce unnecessary overhead.

> Renaming `Connection` to `TcpChannel` would be alright by me, though, if it makes more sense to you that way? Then `TcpContext` would have `channel` and `context` properties?

I think that does make more sense in my head, but I am getting tired so let me sleep on it.

rfcs/015 - TCP RFC.md
cdmistman: we could also provide an argument for the reason? some kind of numeric-based enum maybe? or strings that the user compares to?

rfcs/015 - TCP RFC.md
cdmistman: I prefer the `TcpChannel` name - it feels odd having `Connection` as the name when a Connection in other languages/libraries has a lot more information/properties attached to it

rfcs/015 - TCP RFC.md
dfellis: Yes, but we can expose that with a querying function, I think. When the `connClose` event occurs you can probe `ctx.conn` (or whatever we end up calling it) for the reason.

cdmistman (approved):

rfcs/015 - TCP RFC.md
```diff
@@ -0,0 +1,188 @@
...
+
+The main issue is that Alan's event system is intentionally a static event system determined at compile time, so it is impossible to add new event handlers for each new socket that is opened by the users' code or has been received by the TCP server. The secondary issue is that the AVM's internal memory model would require an expensive conversion of that chunk from packed bytes to an internal `Array<int8>` type which, to simplify the memory model, is actually really an `Array<int64>` under the hood, which is very wasteful for a true byte array, so being able to avoid that conversion when possible would be good. Finally, the ergonomics of the API need to be both easy-to-understand and mesh well with the existing syntax while adding stream processing support at the same time.
+
+While changes to the language are on the table (and some of the alternatives will cover a few of them), the best overall solution (considering implementation complexity and cognitive complexity) doesn't need it. :)
+
+The clearest way to introduce the new syntax is to simply use it in an example similar to how it will be used by Anycloud:
```
depombo, line 43: nit: For someone new coming in, it might help them understand this more quickly by pointing out that this is basically proxying the connection

depombo (approved):

PR #468, opened 2021-03-25
Daemon error handling

aguillenv:


avm/src/daemon/daemon.rs
```diff
@@ -94,21 +106,24 @@ async fn post_v1_scale(
...
+  let vm_stats = get_v1_stats().await?;
   let mut stats_body = json!({
     "deployToken": deploy_token,
     "vmStats": vm_stats,
     "clusterId": cluster_id,
   });
-  let cluster_secret = CLUSTER_SECRET.get().unwrap();
+  let cluster_secret = CLUSTER_SECRET.get().unwrap_or_else(|| {
+    error!("No cluster secret");
+    return &None::<String>;
```
depombo, line 118: shouldn't we return an `Err` here?

avm/src/daemon/stats.rs
```diff
@@ -70,16 +73,29 @@ async fn get_proc_usages() -> Vec<f64> {
...
+          system: cpu.system().get::<second>(),
+          idle: cpu.idle().get::<second>(),
+          irq: 0.0,
+          nice: 0.0,
+          ioWait: 0.0,
+          softIrq: 0.0,
+          steal: 0.0,
+        }
+      } else {
+        error!("Failed to get CPU times");
```
depombo, line 88: we should return this error and only log it in the main daemon loop

avm/src/daemon/daemon.rs
aguillenv: Where it get it's value? If is not possible to get it again for some reason yes, but I thought it was a kind of polling or something that could appear, it does not come from redis?

avm/src/daemon/daemon.rs
```diff
@@ -26,29 +30,40 @@ use crate::vm::run::run;
...
 }
 
 async fn post_v1(endpoint: &str, body: Value) -> String {
   let resp = deploy::post_v1(endpoint, body).await;
   match resp {
     Ok(res) => res,
-    Err(err) => format!("{:?}", err),
+    Err(err) => {
+      let err = format!("{:?}", err);
+      error!(121, "{:?}", err).await;
```
dfellis, line 64: Similar comment as what @cdmistman made in the other PR. These should be in a named enum so the error codes make sense to the reader. No perf impact to doing so.

avm/src/daemon/daemon.rs
```diff
@@ -133,34 +155,45 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
...
+            priv_key_b64: priv_key_b64.to_string(),
+            cert_b64: cert_b64.to_string(),
+          }),
+        )
+        .await;
+      } else {
+        run(bytecode, HttpType::HTTP(HttpConfig { port: 80 })).await;
+      }
+    } else {
+      eprintln!("Error reading signed 64 bit integers from src into dst.");
```
dfellis, line 191: This error message would be very confusing to users. This is also the one indicating that the AGZ/AGC data is corrupt, so I might just change it to `AGZ file appears to be corrupt`?

avm/src/daemon/daemon.rs
```diff
@@ -133,34 +155,45 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
...
+      panic!("Error reading signed 64 bit integers from src into dst.");
+    }
   } else {
-    run(bytecode, HttpType::HTTP(HttpConfig { port: 80 })).await;
+    eprintln!("Unable to decode agz base64.");
```
dfellis, line 195: Minor potential improvement: "AGZ payload not properly base64-encoded"

avm/src/daemon/daemon.rs
```diff
@@ -173,43 +206,112 @@ pub async fn start(
...
+              vms.len(),
+              factor
+            );
+            if factor != "1" {
+              post_v1_scale(&cluster_id, &agzb64, &deploy_token, &factor).await;
+            }
+          }
+        }
+      }
+      (Err(dns_err), Ok(_self_ip)) => {
```
dfellis, line 266: These changes are for the best, but they are going to make it harder to locally test `daemon` mode changes. Need to think about how to handle that.

avm/src/daemon/daemon.rs
```diff
@@ -173,43 +206,112 @@ pub async fn start(
...
+    clone_cert_b64 = None;
+  }
+  let run_agz = task::spawn(async {
+    run_agz_b64(clone_agz_b64, clone_priv_key_b64, clone_cert_b64).await;
+  })
+  .await;
+  match run_agz {
+    Ok(_) => println!("ok"),
+    Err(e) => {
+      println!("{:?}", e);
```
dfellis, line 304: This should be `eprintln!`

avm/src/daemon/daemon.rs
```diff
@@ -173,43 +206,112 @@ pub async fn start(
...
+    clone_priv_key_b64 = Some(String::from(priv_key));
+    clone_cert_b64 = Some(String::from(cert));
+  } else {
+    clone_priv_key_b64 = None;
+    clone_cert_b64 = None;
+  }
+  let run_agz = task::spawn(async {
+    run_agz_b64(clone_agz_b64, clone_priv_key_b64, clone_cert_b64).await;
+  })
+  .await;
```
dfellis, line 300: Why did you wrap this in a new task and then just await it again? Wouldn't:

```rust
let run_agz = run_agz_b64(clone_agz_b64, clone_priv_key_b64, clone_cert_b64).await;
```

work?

avm/src/daemon/daemon.rs
```diff
@@ -173,43 +206,112 @@ pub async fn start(
...
+  let clone_agz_b64 = String::from(agz_b64);
+  let clone_priv_key_b64: Option<String>;
+  let clone_cert_b64: Option<String>;
+  if let (Some(priv_key), Some(cert)) = (priv_key_b64, cert_b64) {
+    clone_priv_key_b64 = Some(String::from(priv_key));
+    clone_cert_b64 = Some(String::from(cert));
+  } else {
+    clone_priv_key_b64 = None;
+    clone_cert_b64 = None;
+  }
```
dfellis, line 296: Why do these clones exist at all? I can't see any purpose to them.

avm/src/daemon/daemon.rs
```diff
@@ -173,43 +206,112 @@ pub async fn start(
...
+      if let Ok(reason) = e.try_into_panic() {
+        match reason.downcast_ref::<&str>() {
+          Some(err) => {
+            eprintln!("{}", err);
+          }
+          None => {
+            eprintln!("{:?}", reason);
+          }
+        }
+      }
```
dfellis, line 314: What is this whole block about?

avm/src/daemon/stats.rs
```diff
@@ -109,68 +115,92 @@ async fn get_cores_total_times() -> Vec<CPUSecsV1> {
...
-    usedSwapKb: swap.used().get::<kilobyte>(),
-    freeSwapKb: swap.free().get::<kilobyte>(),
+pub async fn get_v1_stats() -> Result<VMStatsV1, String> {
+  let memory = heim_memory::memory().await;
+  match memory {
+    Ok(memory) => {
+      let swap = heim_memory::swap().await;
+      match swap {
+        Ok(swap) => {
+          let core_times = get_cores_times().await;
```
dfellis, line 151: Maybe pull this out to something like:

```rs
let memory = heim_memory::memory().await;
let swap = heim_memory::swap().await;
let core_times = get_core_times().await;
match (memory, swap, core_times) {
  (Ok(memory), Ok(swap), Ok(core_times)) => {
    ...
  },
  _ => return Err("Failed to get system stats");
}
```

avm/src/daemon/daemon.rs
aguillenv: It is related with your following two comments. That would work but if something fails or panic inside `run_agz_b64`we will not be able to catch it and log it, so I'm running it in another thread and if that thread panics I'm able to catch it, resume it and log the error before terminating.

aguillenv: Not sure if it's the best or the cleanest way but fought we this a bit and was the solution I came with. Open to suggestions. Also tried with `catch unwind` for futures but docs suggest `It is not recommended to use this function for a general try/catch mechanism.` and was having troubles with boundaries and unsafe code compiling errors.

avm/src/daemon/daemon.rs
aguillenv: As explained on [my previous comment](https://github.com/alantech/alan/pull/468#discussion_r607050426) I'm spawning a task in order to execute the`run_agz_b64` method and by doing this was having the following type of error using the existing variables: `agz_b64 has an anonymous lifetime '_ but it needs to satisfy a 'static lifetime requirement`. That why I created a copy of them.

avm/src/daemon/daemon.rs
aguillenv: Following [this](https://github.com/alantech/alan/pull/468#discussion_r607050426), as I'm spawning a new task to run the agz file and if something panics I'm catching it in this block, if the reason of the `JoinError` returned by the spawned task was a panic, get the ref and be able to print/log the string error message.

depombo:
I think it might be cleaner if we use `?` and return `Result<String, Box<dyn Error>>` within the methods?

avm/src/daemon/daemon.rs
```diff
@@ -133,34 +152,46 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
...
-    run(
-      bytecode,
-      HttpType::HTTPS(HttpsConfig {
-        port: 443,
-        priv_key_b64: priv_key_b64.to_string(),
-        cert_b64: cert_b64.to_string(),
-      }),
-    )
-    .await;
+async fn run_agz_b64(agz_b64: String, priv_key_b64: Option<String>, cert_b64: Option<String>) {
```
depombo, line 155: can we maybe have this return `Result<String, Box<dyn Error>>` and then panic if there is an error?

aguillenv: we could for some cases inside the function return Err but we would not handle the case when `vm::run::run` panics, right?

depombo: we could check the Error value and panic of specific ones?

avm/src/daemon/daemon.rs
cdmistman: in that case, this should work:
```rs
let run_agz = task::spawn(run_agz_b64(clone_agz_b64, clone_priv_key_b64, clone_cert_b64)).await;
```

avm/src/daemon/daemon.rs
cdmistman: I think we do want to clean up the AVM to use errors instead of panics at some point, so I think it's better to use `Result`s

aguillenv: Ok, as agreed, will update to return `Result` here and later on the `vm::run::run` will be updated to return a `Result` as well.

avm/src/daemon/daemon.rs
```diff
@@ -56,12 +54,12 @@ async fn get_private_ip() -> Result<String, String> {
...
 
 async fn post_v1(endpoint: &str, body: Value) -> String {
-  let resp = deploy::post_v1(endpoint, body).await;
+  let resp = deploy_post_v1(endpoint, body).await;
   match resp {
     Ok(res) => res,
     Err(err) => {
       let err = format!("{:?}", err);
-      error!(121, "{:?}", err).await;
+      error!(ErrorKind::PostFailed as u8, "{:?}", err).await;
```
dfellis, line 62: These will be switched to no-casting, too, correct?

aguillenv: Yes

depombo:
@aguillenv @dfellis created https://github.com/alantech/alan/issues/472 to track the refactoring of the error handling for the rest of the AVM

avm/src/daemon/daemon.rs
```diff
@@ -26,29 +29,33 @@ use crate::vm::run::run;
...
+async fn get_private_ip() -> Result<String, Box<dyn Error + Send + Sync>> {
+  let res = Command::new("hostname").arg("-I").output().await?;
+  let stdout = res.stdout;
+  let private_ip = String::from_utf8(stdout);
+  match private_ip {
+    Ok(private_ip) => match private_ip.trim().split_whitespace().next() {
+      Some(private_ip) => Ok(private_ip.to_string()),
+      None => return Err("No ip found".into()),
+    },
+    Err(err) => return Err(err.into()),
```
cdmistman, line 41: nit: you have 2 `return`s here even though all 3 arms return something. Should be consistent between arms.

avm/src/daemon/daemon.rs
```diff
@@ -173,43 +202,86 @@ pub async fn start(
...
   let deploy_token = deploy_token.to_string();
   let agzb64 = agz_b64.to_string();
   let domain = domain.to_string();
   task::spawn(async move {
     // TODO even better period determination
     let period = Duration::from_secs(5 * 60);
-    let dns = DNS::new(&domain);
     let self_ip = get_private_ip().await;
     let mut cluster_size = 0;
     let mut leader_ip = "".to_string();
```
cdmistman, line 214: nit: can you change this to `String::new()`?

avm/src/daemon/dns.rs
```diff
@@ -15,48 +16,70 @@ pub struct VMMetadata {
...
-    DNS {
-      domain: domain.to_string(),
-      resolver: TokioAsyncResolver::tokio(resolver, resolver_opts).unwrap(),
+    let resolver_result = TokioAsyncResolver::tokio(resolver, resolver_opts);
+    match resolver_result {
+      Ok(resolver) => Ok(DNS {
+        domain: domain.to_string(),
+        resolver: resolver,
+      }),
+      Err(e) => Err(e.into()),
```
cdmistman, line 57: can't you use `?` here?

avm/src/daemon/dns.rs
```diff
@@ -15,48 +16,70 @@ pub struct VMMetadata {
...
+    if let Ok(resp) = resp {
+      for rec in resp {
+        let data = &rec.txt_data()[0];
+        let vm = VMMetadata::from_txt_data(data);
+        match vm {
+          Ok(vm) => {
+            vms.push(vm);
+          }
+          Err(err) => return Err(err.into()),
+        }
```
cdmistman, line 78: nit: another potential `?` location

avm/src/daemon/daemon.rs
```diff
@@ -26,29 +29,33 @@ use crate::vm::run::run;
...
-  let err = "Failed to execute `hostname`";
-  let stdout = res.expect(err).stdout;
-  String::from_utf8(stdout)
-    .expect(err)
-    .trim()
-    .split_whitespace()
-    .next()
-    .unwrap()
-    .to_string()
+async fn get_private_ip() -> Result<String, Box<dyn Error + Send + Sync>> {
```
cdmistman, line 32: honestly, I think it makes sense to have a declaration like this that we just `use` everywhere to keep things DRY and reduce bloat
```rs
pub type Result<T> = std::result::Result<T, Box<dyn Error + Send + Sync>>;
```

avm/src/daemon/daemon.rs
```diff
@@ -133,35 +149,48 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
...
+        // Spin up a control port if we can start a secure connection
+        make_server!(
+          HttpType::HTTPS(HttpsConfig {
+            port: 4142, // 4 = A, 1 = L, 2 = N (sideways) => ALAN
+            priv_key_b64: priv_key_b64.to_string(),
+            cert_b64: cert_b64.to_string(),
+          }),
+          control_port
+        );
+        run(
```
depombo, line 175: Comment that this can panic and we won't catch it and/or TODO to have this return a `Result`

avm/src/daemon/daemon.rs
```diff
@@ -173,43 +202,86 @@ pub async fn start(
...
+      }
+      (Err(dns_err), Err(self_ip_err)) => {
+        error!(
+          ErrorType::NoDnsPrivateIp,
+          "DNS error: {} and Private ip error: {}", dns_err, self_ip_err
+        )
+        .await;
+        panic!(
+          "DNS error: {} and Private ip error: {}",
+          dns_err, self_ip_err
```
depombo, line 278: I think we get rid of the `NoDnsPrivateIp` error and the complex match statement and simply do:
```
if let Err(dns_err) = dns {
  error!(ErrorType::NoDns, "DNS error: {}", dns_err).await;
        panic!("DNS error: {}", dns_err);
}
if let Err(ip_err) = _self_ip {
  error!(ErrorType:: NoPrivateIp, "Private ip error: {}", ip_err).await;
        panic!("DNS error: {}", ip_err);
}
```

avm/src/daemon/daemon.rs
depombo: That is cleaner. It might need to be called something other than `Result` to avoid import conflicts though

aguillenv: Nice one. I've called it `DaemonResult` to avoid collision with other results around.

avm/src/daemon/daemon.rs
aguillenv: I tried something like that but was having troubles unwrapping inside the `loop`, errors like `value moved here, in previous iteration of loop`. Also tried with `unwrap_or_else` but had troubles with the returns since we need the closure to be `async` and having errors like `expected struct `DNS`, found opaque type`.

So not know if you have any other idea that I'm missing. One thing that I can do for sure is remove one of the error branches.

depombo: Got it. That sucks but makes sense. Let's remove one branch and `NoDnsPrivateIp` then

aguillenv: Wait I might thought a work around, let me check

avm/src/daemon/daemon.rs
```diff
@@ -23,32 +26,38 @@ use crate::make_server;
...
+async fn get_private_ip() -> DaemonResult<String> {
+  let res = Command::new("hostname").arg("-I").output().await?;
+  let stdout = res.stdout;
+  let private_ip = String::from_utf8(stdout);
+  match private_ip {
+    Ok(private_ip) => match private_ip.trim().split_whitespace().next() {
+      Some(private_ip) => Ok(private_ip.to_string()),
+      None => Err("No ip found".into()),
+    },
+    Err(err) => Err(err.into()),
```
cdmistman, line 43: nit: another `?` candidate

avm/src/daemon/daemon.rs
cdmistman: So this doesn't work?
```rust
if let (Ok(dns), Ok(self_ip)) = (dns, self_ip) {
  ...
} else if let Err(dns_err) = dns {
  ...
} else if let Err(ip_err) = self_ip {
  ...
}
```

aguillenv: no, got 
```
error[E0382]: use of moved value
   --> src/daemon/daemon.rs:264:23
    |
213 |     let self_ip = get_private_ip().await;
    |         ------- move occurs because `self_ip` has type `std::result::Result<std::string::String, Box<dyn StdError + std::marker::Send + std::marker::Sync>>`, which does not implement the `Copy` trait
...
217 |     if let (Ok(dns), Ok(self_ip)) = (dns, self_ip) {
    |                                           ------- value moved here
...
264 |     } else if let Err(self_ip_err) = self_ip {
    |                       ^^^^^^^^^^^ value used here after move
```

aguillenv: Manage to remove the `self_ip` match with an `unwrap_or_else`. Could not do the same for the `DNS` since I need to return a DNS struct and did not know how to initialize an `AsyncResolver`.
https://github.com/alantech/alan/pull/468/files?file-filters%5B%5D=.rs#diff-583f2380a4d780cb04bd93e3492a6f2828052a50e23460f3a2c8d89cf3ade79fR197-R281

avm/src/daemon/daemon.rs
```diff
@@ -173,43 +200,79 @@ pub async fn start(
...
+            let mut factor_err: Option<String> = None;
+            let factor = post_v1_stats(&cluster_id, &deploy_token)
+              .await
+              .unwrap_or_else(|err| {
+                factor_err = Some(format!("{}", err));
+                return "1".to_string();
+              });
+            if let Some(err) = factor_err {
+              error!(ErrorType::PostStats, "{}", err).await;
+            };
```
depombo, line 255: can't this be a `match` statement on `post_v1_stats(&cluster_id, &deploy_token).await`?

aguillenv: Yes, will change with `if let`. Legacy from previous changes

avm/src/daemon/daemon.rs
```diff
@@ -23,32 +26,35 @@ use crate::make_server;
...
+  match private_ip.trim().split_whitespace().next() {
+    Some(private_ip) => Ok(private_ip.to_string()),
+    None => Err("No ip found".into()),
+  }
 }
 
 #[cfg(not(target_os = "linux"))]
-async fn get_private_ip() -> String {
+async fn get_private_ip() -> DaemonResult<String> {
   panic!("`hostname` command does not exist in this OS");
```
dfellis, line 46: Technically this could be ```return Err(new Error("`hostname` command does not exist in this OS"))``` now, right?

avm/src/daemon/daemon.rs
```diff
@@ -212,66 +209,55 @@ pub async fn start(
...
+            .map(|vm| vm.private_ip_addr.to_string())
+            .collect();
+          let lrh = LogRendezvousHash::new(ips);
+          leader_ip = lrh.get_leader_id().to_string();
+        }
+        if leader_ip == self_ip.to_string() {
+          let mut factor = String::from("1");
+          let stats_factor = post_v1_stats(&cluster_id, &deploy_token).await;
+          if let Ok(stats_factor) = stats_factor {
+            factor = stats_factor
```
cdmistman, line 240: you're returning the return value of an assignment (`()`) here...
```suggestion
            factor = stats_factor;
```

cdmistman (approved):

avm/src/daemon/daemon.rs
cdmistman: ah yes - we shouldn't be panicking

aguillenv: Yes!

dfellis (approved):
Just a nit.

avm/src/daemon/daemon.rs
```diff
@@ -133,35 +145,50 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
...
+      if let (Some(priv_key_b64), Some(cert_b64)) = (priv_key_b64, cert_b64) {
+        // Spin up a control port if we can start a secure connection
+        make_server!(
+          HttpType::HTTPS(HttpsConfig {
+            port: 4142, // 4 = A, 1 = L, 2 = N (sideways) => ALAN
+            priv_key_b64: priv_key_b64.to_string(),
+            cert_b64: cert_b64.to_string(),
+          }),
+          control_port
+        );
```
cdmistman, line 170: i think this also has the possibility of panicking or exiting, we should fix that soon

dfellis (approved):

avm/src/daemon/daemon.rs
depombo: yeah, https://github.com/alantech/alan/issues/472 tracks it

depombo (approved):

PR #470, opened 2021-03-31
Implement @std/tcp for the js-runtime

dfellis:
Needed to add a `ready` function to get it working as desired. Glad it was caught by the js-runtime since I think the AVM is even more vulnerable to this problem.

### DIFF
--- a/bdd/spec/033_tcp_spec.sh
+++ b/bdd/spec/033_tcp_spec.sh
@@ -1,0 +1,60 @@
+Include build_tools.sh
+
+Describe "@std/tcp"
+  Describe "webserver tunnel test"
+    before() {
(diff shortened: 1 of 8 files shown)


bdd/spec/033_tcp_spec.sh
```diff
@@ -0,0 +1,60 @@
...
+      "
+    }
+    BeforeAll before
+
+    after() {
+      cleanTemp
+    }
+    AfterAll after
+
+    afterEach() {
```
cdmistman, line 41: does it make sense to `unset` the pid vars?

depombo (approved):

cdmistman (approved):

PR #471, opened 2021-04-01
Cluster Data Storage RFC

dfellis:
Man this took longer than I expected, but the RFC process definitely helped clear out some issues with cluster routing, so very glad to have done it!


rfcs/017 - Cluster Data Storage RFC.md
```diff
@@ -0,0 +1,184 @@
...
+
+![A partial partition diagram of the paragraph above](./partial_partition.jpg)
+
+This all can be contained in a relatively simple `/ping/<public_ip_address>` control port endpoint. Normally the `/ping/<public_ip_address>` would be directed at `<public_ip_address>`, which should always return `ok` (as long as the cluster secret token is also provided as a header, otherwise `fail`, but that's at an earlier part of the control port implementation). But if a request for `/ping/<public_ip_address>` comes in and `<public_ip_address>` is not the node's own public IP address, then it immediately forwards the request on to the public IP address in the request URL and returns its response or `fail` if it times out (should be a short timeout as well, such as 50ms, but exact threshold depends on real world data here).
```
cdmistman, line 71: why do we need this instead of just utilizing ICMP? It's more direct, already exposed, and we don't have to write anything to handle the logic on the receiver end - we can just 1 ICMP echo packet directly to the node and if we get a response, we know the node is online.

rfcs/017 - Cluster Data Storage RFC.md
```diff
@@ -0,0 +1,184 @@
...
+
+To be clear, this vulnerability affects `@std/datastore` as it currently exists even without the state being distributed across a cluster. It's just harder to hit when the latencies are lower when completely local. The traditional approach of locks on shared mutable state, besides being complicated to follow for many developers, also becomes a performance bottleneck when the locks are distributed across a cluster and lock acquisition and release times are several orders of magnitude slower. For this reason, an approach based on Software Transactional Memory (STM) is the choice to handle mutation of one-or-more datastore keys.
+
+Because STM re-runs the code involved in a mutation of memory, side-effects can be triggered multiple times. If one of those "side-effects" is billing a customer through Stripe, for instance, that would be a *very* bad thing to do automatically without any indication to the user! Also because the STM behavior increases the total latency to perform a mutation (as it needs to confirm that all writes are good before committing them, more on that algo later) being able to escape that behavior when using datastore as just a cache is also desireable.
```
cdmistman, line 89: Side note: I wonder if it's possible to fake atomicity in our http implementation to ensure that the http request would only be sent 1x in this case?

cdmistman (approved):

rfcs/017 - Cluster Data Storage RFC.md
dfellis: So our definition of "up" and "down" isn't just "is there a working Linux kernel on the other side" but "is the AVM on the other side actually working" which requires both listening on port 4142 and the Tokio event loop not being "stuck" (eg user's code has either hit a bug or slammed the process such that it can't respond to external requests correctly).

Also, since we want to use this for potential forwarding logic, the IP address in the `/ping/` request isn't guaranteed to be the IP address the request is pointed at, and when that's the case, extra logic within the AVM to also potentially update its own status of the node in question should be triggered (if I am being requested to see if another node is down, it is suspect; if my own attempt to ping it fails, I should immediately trigger asking the other nodes if they have seen it, not including the node that triggered this request of me; and if I already have them marked as down, I should just report them as down immediately since I already have my own background Tokio task to periodically check if that node is back online). An ICMP packet isn't capable of all of this downstream work triggering. :)

rfcs/017 - Cluster Data Storage RFC.md
dfellis: Unfortunately not really possible.

1. If the HTTP request is to a static URL with a static payload, then in only that case *maybe* (see point 2 for why even then sometimes no), but if that URL and/or payload depend on something constructed during the `atomic` operation each one would be different and it would be hard to tell which is right.
2. If you try to delay calling the HTTP request until after the `atomic` execution has succeeded, you could cause the equivalent of a deadlock if that HTTP *response* is part of the data to be put into the datastore, so the STM mode is waiting for data that it doesn't have because its waiting for STM mode to finish, therefore it can never finish.

So because of those two points, silently running STM all the time isn't something that could work "correctly" from the perspective of a developer used to single-threaded imperative programming. That's why explicitly calling it out in a function is worthwhile.

rfcs/017 - Cluster Data Storage RFC.md
```diff
@@ -0,0 +1,221 @@
...
+* Fronting the datastore internal data structure with this logic instead of using it directly by the opcodes (when in daemon mode, leaving it as-is or having a "cluster" of one when running locally)
+* Exposing this new behavior into the docker container, and the library work necessary to make this feature usable and natural in other languages.
+* Adding atomic functionality to the datastore API and internals to make distributed mutation (not just key replacement) possible
+
+### Cluster Coordination Work
+
+The new endpoints have some extra logic that they need to tackle, as this is the first time the nodes of the cluster are actually communicating with each other. The LogRendezvousHash module needs to be updated on the *actual* state of the cluster versus the expected state defined in the DNS records: which nodes are actually up or down from its perspective? That determines which nodes are considered for ownership of the datastore data. RendezvousHash (and our own LogRendezvousHash) are stable, in that the top `n` nodes after one of them goes down will still contain `n - 1` of the original set, so not too much data thrashing occurs when a node goes down or comes online.
```
depombo, line 63: We are not proposing that nodes write to the DNS record, right?

rfcs/017 - Cluster Data Storage RFC.md
```diff
@@ -0,0 +1,221 @@
...
+
+Further, while a full network partition between two groups of nodes will result in a "split brain" scenario in an unavoidable fashion, with LogRendezvousHash it is possible to counteract a partial split by way of a request forwarding mechanism. If a node cannot directly reach another node, it should try to ask all other nodes in its list if any of them can reach the node. If so, the first one to respond in the positive becomes the new "router" that all requests to that node are forwarded through, with periodic attempts in the background to remove the extra hop by re-polling. Only if all nodes a node can reach report that they also cannot reach the node in question should it be marked as "down" and assignment of keys shuffled with the new, reduced set. This is fundamentally how the internet routes packets to unknown MAC addresses, and can guard against misconfigured firewalls/routers black hole-ing packets that they shouldn't be, but only works when there is an unaffected third party, so requires at least three separate sub-clusters to work.
```
depombo, line 65: Seeing the data structure used to represent the "report" and "router" for a cluster that every node in it will have would be useful?

rfcs/017 - Cluster Data Storage RFC.md
```diff
@@ -0,0 +1,221 @@
...
+
+For example, let's presume a cluster has been deployed to all three major cloud providers: AWS, Azure, and GCP. Now let's presume that GCP has a network configuration error that makes AWS nodes unreachable to it and vice versa. When the GCP nodes want to get data that was previously assigned to AWS, they will fail to contact the AWS node, then try to ask all other nodes (which would include the other AWS nodes) which ones can reach the AWS node in question. The GCP nodes would report a failure, but the Azure nodes would report a success, and one of those nodes will be chosen effectively at random to be the node that will route to the AWS node. This process would repeat across every GCP node and AWS node combination with an effectively random assignment of Azure nodes to handle the load, which should lead to an approximately even level of traffic distribution between the Azure nodes.
```
depombo, line 67: Do three separate sub-clusters necessarily mean three sub clusters across regions or cloud providers, or simply if there is one unreachable/failing node within a 3-node cluster in the same region and cloud provider? That seems like a more likely scenario that we should also cover where one node receives a request that blows up the process.

rfcs/017 - Cluster Data Storage RFC.md
```diff
@@ -0,0 +1,221 @@
...
+
+![A partial partition diagram of the paragraph above](./partial_partition.jpg)
+
+This all can be contained in a relatively simple `/ping/<public_ip_address>` control port endpoint. Normally the `/ping/<public_ip_address>` would be directed at `<public_ip_address>`, which should always return `ok` (as long as the cluster secret token is also provided as a header, otherwise `fail`, but that's at an earlier part of the control port implementation). But if a request for `/ping/<public_ip_address>` comes in and `<public_ip_address>` is not the node's own public IP address, then it immediately forwards the request on to the public IP address in the request URL and returns its response or `fail` if it times out (should be a short timeout as well, such as 50ms, but exact threshold depends on real world data here).
```
depombo, line 71: Will there any way to know if a request was forwarded/proxied? If not, why doesn't it matter?

rfcs/017 - Cluster Data Storage RFC.md
dfellis: No, this is their own internal record keeping of the cluster as seen from that nodes' perspective.

rfcs/017 - Cluster Data Storage RFC.md
dfellis: I considered it an implementation detail? The nodes need to track metadata on all of the other nodes to determine what nodes are in charge of what and how to route between the nodes, the exact structure will depend on what is clean and efficient to write in Rust. This high level description felt like more than enough?

rfcs/017 - Cluster Data Storage RFC.md
cdmistman: typically (reverse) proxies use `X-forwarded-for` - we can probably use that. I imagine we'll have to return the status of the daemon at the address

rfcs/017 - Cluster Data Storage RFC.md
dfellis: One unreachable/failed node is automatically already dealt with by the RendezvousHash logic as described back in the 90s, so I didn't consider it worth going over in this RFC because it's more trivial.

rfcs/017 - Cluster Data Storage RFC.md
dfellis: So if you call `/ping/<someIPaddress>` to `someIPaddress` you know there's no forwarding because you're pinging that IP address directly. If you call `/ping/<someIPaddress>` to `someOtherIPAddress` then you *know* that there's forwarding going on because you're explicitly asking for it to be forwarded?

Was this question meant for a different paragraph, maybe? The other endpoints could silently forward for you, but `/ping` is pretty simple, I thought.

rfcs/017 - Cluster Data Storage RFC.md
aguillenv, line 65: regarding the background polling, would be a time were we declare a node as lost or any number of retries oe will polling continuously?

rfcs/017 - Cluster Data Storage RFC.md
```diff
@@ -0,0 +1,221 @@
...
+
+Since all nodes in a cluster will be running the same version of Alan, which means they all have the same HandlerMemory, [protobuf](https://github.com/protocolbuffers/protobuf) is the best solution here for speed, size, and compatibility concerns, as it officially supports many languages and there is also a [Rust implementation](https://docs.rs/protobuf/2.22.1/protobuf/) that can be used. There are other protocols that are more widespread (JSON), faster (CapnProto), and smaller (Serde), but Protobuf hits a sweet spot where it's decent at all three (and it isn't an Apache project monstrosity, will veto any push for something out of apache.org).
+
+### Global Mutable State that Actually Works
```
aguillenv, line 83: The distributed store compute should be something we should take into account for the compute or might be significant for the performance of the node and our "autoscaling"?

rfcs/017 - Cluster Data Storage RFC.md
dfellis: So, I'm imagining that it would poll relatively infrequently continuously in the background as long as it still exists in DNS. Eventually the `deploy` service would either pick up the failure as well and remove the dead node and add a new one (at which point it won't be in DNS anymore so it would stop polling it) or just a full redeploy of the cluster would take out the node that is polling as well.

rfcs/017 - Cluster Data Storage RFC.md
dfellis: I'm not sure I understand the question here?

rfcs/017 - Cluster Data Storage RFC.md
depombo: It is worthwhile to point that out even if it is trivial since the reader of the RFC might not be familiar? I understand it makes the RFC longer and that was also probably a factor. I would have though about the cluster coordination and mutable distributed global state as separate RFCs, but I also understand why you made them into the same RFC after reading through it.

rfcs/017 - Cluster Data Storage RFC.md
```diff
@@ -0,0 +1,221 @@
...
+
+Providing the cluster secret string into the Docker container as an environment variable along with the 4142 port is the other piece that makes it possible, and these requests can be put into a simple library for the language the Anycloud app is written in to cover up these implementation details and make it fit in better with the language in question.
+
+But those endpoints make clear something else that's necessary: the `HandlerMemory` structs that represent the value of a particular key need to be able to be serialized and deserialized, and in a way that they can be done so not just in Rust, but any language we want to fully support in Anycloud. (That's not entirely correct, we could create a set of secondary endpoints just for Anycloud that only accepts strings, or rather `Array<int8>` arrays where strings are a special case of that, but that may limit/impede future efforts at RPC between the Alan and Anycloud for offloading CPU-intensive work to Alan, and we still need serialization/deserialization of HandlerMemory for Alan itself.)
```
depombo, line 79: I'm a bit confused. Isn't `@std/datastore` backed by a [Dashmap](https://crates.io/crates/dashmap)? Would that change?

rfcs/017 - Cluster Data Storage RFC.md
depombo: The question was in regards to all cluster coordination. I like that it is as simple as looking at the ip address in the url and comparing it to itself.

rfcs/017 - Cluster Data Storage RFC.md
depombo: Fair

rfcs/017 - Cluster Data Storage RFC.md
```diff
@@ -0,0 +1,221 @@
...
+
+#### Approvers
+
+- Luis De Pombo <luis@alantechnologies.com>
+- Alejandro Guillen <alejandro@alantechnologies.com>
+- Colton Donnelly <colton@alantechnologies.com>
+
+### Implementation
+
+- [ ] Implemented: [One or more PRs](https://github.com/alantech/alan/some-pr-link-here) YYYY-MM-DD
```
depombo, line 21: This should probably be a milestone or two?

rfcs/017 - Cluster Data Storage RFC.md
dfellis: Sure. I can add that, but I won't remove this example because it was the more important piece to figure out the high-level algorithm. (Just using RendezvousHash by itself here didn't solve the partial partition problem.)

rfcs/017 - Cluster Data Storage RFC.md
dfellis: So we can use `X-forwarded-for` as @cdmistman suggested for easier debugging, but from the API perspective the automatic forwarding was to abstract the routing logic from higher layers in the stack (and from the Anycloud apps). It also forced a more complete solution because all possible conditions of a smaller surface area were considered (all nodes up, one or more nodes actually down, partial partitioning of the cluster, and full partitioning of the cluster into two or more sub-clusters, where the latter was decided to be an unresolveable scenario, but *every other outage scenario* is now recoverable :) ).

rfcs/017 - Cluster Data Storage RFC.md
dfellis: I'm hesitant to break it up since as you can see from the diagram at the bottom of the RFC that there's a *lot* here and multiple ways we could slice it up. Even just the cluster status table may take multiple PRs, so I didn't want to state "must be one PR" in here.

depombo: Got it, that makes sense

rfcs/017 - Cluster Data Storage RFC.md
depombo: I think @aguillenv means how the different failure scenarios for cluster coordination affect how we autoscale?

rfcs/017 - Cluster Data Storage RFC.md
dfellis: No, that wouldn't change. What would change is that we need to put a thin layer on top of it to:

1. Go ask another node via its control port for the data if it's not the primary node for that data.
2. Any time it writes data to the local DashMap, also push that data to the 2 other replicas in the cluster (assuming there are at least 3 nodes in the cluster).
3. Every time the node assignments have changed, ask every other node if there's data it should now be in charge of and drop any data it is no longer even in charge of replicating from its own DashMap.

However, because we discovered that `DashMap` has a huge perf hit when dropping data, it may make sense to re-eval its use since dropping keys is expected in the normal lifetime of the datastore. But I also considered that an implementation detail that doesn't belong in the RFC. The responsibilities of the thin layer *should* be in the RFC, though.

depombo: It is an implementation detail, but I was missing the "thin layer on top of it" part. Thanks for clarifying!

depombo (approved):

rfcs/017 - Cluster Data Storage RFC.md
aguillenv: Yes, my comment it is non blocking but just to understand if is something that could affect significantly the load the system even when our autoscaler would manage it.

aguillenv (approved):

cdmistman (approved):

rfcs/017 - Cluster Data Storage RFC.md
dfellis: Discussed offline, but for future reference, the load that this routing adds to the cluster would be monitored by the existing stats monitoring in the AVM, since that is at the server level, not process level, so the autoscaler should still kick in if the increased load from rerouting messages becomes a big deal.

Further, the actual amount of compute needed for the rerouting is very small and unlikely to be more than a rounding error, so it shouldn't ever be the trigger of the autoscaler, anyways.

PR #473, opened 2021-04-05
Initial lnn support

cdmistman:
This is the first PR for [lnn](https://github.com/alantech/alan/blob/main/rfcs/016%20-%20Reworked%20Compiler%20Front%20End.md), the secondary Alan compiler front-end that will replace the current one.

This PR will be merge-able once basic programs are able to be compiled and successfully run. This means:
- [x] Functions where param types and the return type are defined
- [x] `const` and `let` variables with type definitions
- [ ] ~~Operators with only 1 possible function call~~
  - operators coming once function selection is done
- [ ] ~~Conditionals using the new conditional opcode format (see: #430)~~
  - this is just a TODO, still

Note that, while the above are the only goals for the current PR, there may be some boilerplate or incomplete logic for future PRs (just to get my thoughts down and to reduce work needed for said future PRs)


Makefile
```diff
@@ -46,7 +46,7 @@ node_modules: build
 
 .PHONY: clean
 clean:
-	git clean -ffdxe .vscode
+	git clean -ffdxe .vscode -e compiler/src/lnntoamm
```
dfellis, line 49: what is this and why?

cdmistman: ah when I was first setting up and figuring things out the dir was getting deleted because of `git clean`, so this was to prevent the dir from getting wiped - I just forgot to delete this

compiler/src/lnntoamm/Amm.ts
```diff
@@ -17,6 +17,22 @@ export default class Output {
...
+      }
+    }
+    for (let eventName of Object.keys(this.events)) {
+      res = res.concat('event ', eventName, ': ', this.events[eventName].ammName);
+    }
+    for (let handler of this.handlers) {
+      res = res.concat(handler);
+    }
+    return res;
+  }
```
dfellis: Believe it or not, but making this an array, `push`ing these pieces into it, and then `join`ing them at the end will be faster.

cdmistman: short story for anyone reading this comment:
I had seen people say that `concat` is faster than `join`, which is why I used `concat`. After seeing this comment, @dfellis and I both found separate benchmarks that tested this feature. His benchmark found that my original hypothesis is true, while my benchmark found that it's false. I think we'll just have to benchmark this ourselves...

compiler/src/index.ts
```diff
@@ -82,7 +85,7 @@ if (convert[getFormat(inputfile)] && convert[getFormat(inputfile)][getFormat(out
     const end = Date.now()
     console.log(`Done in ${formatTime(end - start)}`)
   } catch (e) {
-    console.error(e.message)
+    console.error(e)
```
dfellis, line 88: Restore this change back, please

compiler/src/lnntoamm/Event.ts
```diff
@@ -0,0 +1,99 @@
...
+      // }
+      // retConstraints.push(opcodes().get('void') as Type);
+      // really tsc????? you're gonna complain that handler could be an array
+      // here but not above? wat?
+      // if (!retConstraints.every(con => (handler as Fn).retTy.compatibleWithConstraint(con))) {
+      //   throw new Error(`failed to type-check: event handler ${handler}`);
+      // }
+      handler.asHandler(amm, this.ammName);
+    }
+  }
```
dfellis, line 77: Two general comments here:

1. You're allowed to have mutliple handlers registered to a single event. the runtimes will copy the payload to all of them an let them run, so it's not as much of a selection as an actual array of handlers to emit.
2. Is this commented code intended to be deleted before this is landed or no? If not, can a small `TODO` comment on top explaining what it's for be added?

compiler/src/lnntoamm/Event.ts
```diff
@@ -0,0 +1,99 @@
...
+  //     for (let {dec, constraints} of varConstraints) {
+  //       if (dec ===  null) {
+  //         retConstraints.push(...constraints);
+  //         continue;
+  //       }
+  //       console.log(dec.name, 'is', dec.ty.name, 'constraints:', constraints);
+  //       console.log(`${dec.name} is ${dec.ty.name}, constrained: ${constraints.map(t => t.name).join(',')}`)
+  //     }
+  //   }
+  // }
```
dfellis, line 98: Same. Is this relevant in the future, or actually dead code to be deleted?

compiler/src/lnntoamm/Amm.ts
```diff
@@ -0,0 +1,136 @@
...
+  emit(
+    eventName: string,
+    val: string,
+  ) {
+    const line = this.indent.concat('emit ', eventName, ' ', val);
+    DEBUG_MODE_PRINTING && console.log(line);
+    this.handlers[0] = this.handlers[0].concat(line.concat('\n'));
+  }
+
+  return(
```
dfellis, line 122: This works!? Not using the `return` keyword is why I called it `exit` before.

Maybe JS allows this but Java did not?

compiler/src/lnntoamm/Event.ts
cdmistman: the comments are deleted code, I'll go ahead and clean up.

RE: multiple handlers, good to know! I'll fix that

compiler/src/lnntoamm/Event.ts
cdmistman: dead code, will delete

compiler/src/lnntoamm/Amm.ts
cdmistman: yeah, Java has a lot of hard keywords, but JS is a lot more forgiving :) I'll go ahead and change it to `exit` though

compiler/src/lnntoamm/Event.ts
```diff
@@ -0,0 +1,66 @@
+import { LPNode } from "../lp";
+import Output from "./Amm";
+import Fn from './Fn';
+import Scope from "./Scope";
+import Type, { Builtin } from "./Types";
+import { genName, TODO } from "./util";
+
+var allEvents: Event[] = [];
```
dfellis, line 8: Nit: why not `let`?

compiler/src/lnntoamm/Fn.ts
```diff
@@ -0,0 +1,235 @@
+import { LPNode, NulLP } from '../lp';
+import Output from './Amm';
+import opcodes from './opcodes';
+import Scope from './Scope';
+import Stmt, { Assign, Dec, Emit, Exit, FnArg, MetaData, Ref } from './Statement';
+import Type, { FunctionType } from './Types';
+import { TODO } from './util';
+
+// the value is null if the type is to be inferred
+export type Args = {[name: string]: FnArg};
```
dfellis, line 10: Then why not `{[name: string]: FnArg | null}`

compiler/src/lnntoamm/Fn.ts
```diff
@@ -0,0 +1,235 @@
...
+      this.name,
+      Object.values(this.args).map(a => a.ty),
+      this.retTy
+    );
+  }
+
+  static fromFunctionsAst(
+    ast: LPNode,
+    scope: Scope,
+    // metadata: MetaData = null,
```
dfellis, line 60: Delete this?

compiler/src/lnntoamm/Fn.ts
```diff
@@ -0,0 +1,235 @@
...
+  ): Fn {
+    // TODO: inheritance
+    let metadata = new MetaData(scope);
+
+    let work = ast;
```
dfellis, line 65: Why this renaming here? I don't think it serves any purpose?

compiler/src/lnntoamm/Fn.ts
```diff
@@ -0,0 +1,235 @@
...
+      bodyAsts = bodyAsts.get('functionbody').get('statements').getAll().map(s => s.get('statement'));
+      bodyAsts.forEach(ast => body.push(...Stmt.fromAst(ast, metadata)));
+    } else {
+      bodyAsts = bodyAsts.get('assignfunction');
+      body = Stmt.fromAst(bodyAsts, metadata);
+      const retVal = body[body.length - 1];
+      if (!(retVal instanceof Dec)) {
+        throw new Error(`illegal function body: ${bodyAsts}`);
+      }
+      body.push(new Exit(bodyAsts, retVal.ref()));
```
dfellis, line 93: A few things here.

1. isn't `assignfunction` going to have a single assignable statement inside of it, so this isn't an array at all?
2. won't `body.length` always equal `1` so you can just do `const retVal = body[0]`
3. won't the `retVal` be an assignable, not a declaration so won't this always throw an error?

compiler/src/lnntoamm/Fn.ts
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    return new Fn(
+      ast,
+      new Scope(scope),
+      name,
+      args,
+      Type.getFromTypename(retTy, scope),
```
dfellis, line 101: This breaks return type inference, doesn't it? (I didn't see the return type being taken from the exit statement(s) so it would be `void` if not explicitly defined.

compiler/src/lnntoamm/Fn.ts
```diff
@@ -0,0 +1,235 @@
...
+  ): Fn {
+    let body = [];
+    let metadata = new MetaData(scope);
+    ast.get('statements').getAll().map(s => s.get('statement')).forEach(ast => body.push(...Stmt.fromAst(ast, metadata)));
+    return new Fn(
+      ast,
+      new Scope(scope),
+      null,
+      {},
+      opcodes().get('void'),
```
dfellis, line 118: Can be fixed later, but conditional function bodies could be something other than void

compiler/src/lnntoamm/Fn.ts
```diff
@@ -0,0 +1,235 @@
...
+  asHandler(amm: Output, event: string) {
+    let handlerArgs = [];
+    for (let arg of Object.keys(this.args)) {
+      handlerArgs.push([arg, this.args[arg].ty.breakdown()]);
+    }
+    amm.addHandler(event, handlerArgs, this.retTy.breakdown());
+    let isReturned = false;
+    for (let ii = 0; ii < this.body.length; ii++) {
+      const stmt = this.body[ii];
+      // console.log(stmt);
```
dfellis, line 137: delete this?

compiler/src/lnntoamm/Module.ts
```diff
@@ -0,0 +1,278 @@
...
+            module.moduleScope.put(importName, thing);
+          }
+        });
+      }
+    });
+
+    // now we're done with imports, move on to the body
+    const body = ast.get('body').getAll();
+
+    // type
```
dfellis, line 106: I don't think these comments add much here

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    scope: Scope,
+    variables: { [name: string]: Dec } = null,
+    retConstraints: Type[] = null,
+  ) {
+    this.scope = scope;
+    this.variables = variables !== null ? variables : {};
+    this.retConstraints = retConstraints !== null ? retConstraints : [];
+  }
+
+  var(name: string): Dec {
```
dfellis, line 25: 1. `var` as a method name
2. The return type here is `Dec | null` right?

compiler/src/lnntoamm/Event.ts
cdmistman: I can change it to `let`, it just felt wrong to me to use `let` with my go/rust background

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    if (this.variables[name] == null) {
```
dfellis, line 26: `==` instead of `===` maybe changing it explicitly to `this.variables.hasOwnProperty(name)` instead?

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    }
+    return stmts;
+  }
+
+  static fromAssignables(ast: LPNode, metadata: MetaData): Stmt[] {
+    let stmts = [];
+
+    let asts: LPNode[] = ast.getAll().map(a => a.get('withoperators'));
+    if (asts.length > 1) TODO('operators');
+    asts = asts.length === 1 ? asts.pop().get('baseassignablelist').getAll().map(a => a.get('baseassignable')) : [];
```
dfellis, line 80: Nit: may want to check if the only value is a `baseassignablelist` and throw an error if the only value is an operator?

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+              fnname: work.get('variable'),
+              fncall: next.get('fncall'),
+            },
+            (work as NamedOr).filename,
+            work.line,
+            work.char,
+          );
+          stmts.push(...Call.fromAsts(
+            callAst,
+            // null,
```
dfellis, line 112: Delete this?

compiler/src/lnntoamm/Fn.ts
cdmistman: this comment is slightly outdated - it used to be `export type Args = {[name: string]: Type | null};`, but I've since added the `FnArg` class and ensure that each instance's type isn't `null`. I'll delete this comment since it no longer applies

compiler/src/lnntoamm/Fn.ts
cdmistman: I was reassigning `work` in a previous iteration so that I can keep the original `ast`, but it's no longer needed

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    return stmts;
+  }
+
+  exprTy(): Type {
+    throw new Error(`assignments aren't expressions`);
+  }
+
+  inline(amm: Output) {
+    const name = this.upstream.ammName;
+    const ty = this.upstream.ty.breakdown(); // always use the declaration's type, since it's been reduced.
```
dfellis, line 180: what about something like:

```ln
let foo = none();
foo = some(5);
```

Using the two resolved types for constraints seems to be in order here? But this depends on generic types being in, so I can see why it's not here yet.

That's also not entirely academic, I could see someone doing:

```ln
let out = none();
if myCondition {
  out = some(3.14);
} else if myOtherCondition {
  out = some(6.28);
}
return out;
```

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+  ) {
+    super(ast)
+    this.fn = fn;
+    this.args = args;
+  }
+
+  static fromAsts(
+    wholeAst: LPNode,
+    fnName: string,
+    fnCallAst: LPNode,
```
dfellis, line 208: Why does this exist? Why not just extract the `fnCallAst` from `wholeAst`?

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+      if (stmts.length <= snap) {
+        throw new Error(`didn't generate any statements for arg ${a.t.trim()}`);
+      }
+      let dec: Stmt = stmts[stmts.length - 1];
+      if (dec instanceof Dec) {
+        return dec.ref();
+      } else if (dec instanceof Ref) {
+        return stmts.pop() as Ref;
+      } else {
+        console.log(dec);
```
dfellis, line 230: Delete this?

compiler/src/lnntoamm/Fn.ts
cdmistman: So yes, it will have a single assignable statement inside of it, but while I'm generating the graph I also generate intermediate steps. So input like this:
```alan
fn addOne(x: int64) = add(1, x);
```
will generate this array with `Stmt.fromAst`:
```javascript
[
  Dec {
    name: '_abc123',
    val: Lit {
      val: '1',
      ty: Type.oneOf(['float32', 'float64', 'int8', 'int16', 'int32', 'int64']),
  },
  Dec {
    name: '_abc124',
    val: Call {
      fn: <opcodes.get('add')>,
      args: [
        <Ref to _abc123>,
        <Ref to x>,
      ],
  },
]
```

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    const argAsts: LPNode[] = [
+      fnCallAst.get('assignables'),
+      ...fnCallAst.get('cdr').getAll().map(a => a.get('assignables')),
+    ];
+    const args: Ref[] = argAsts.map(a => {
+      const snap = stmts.length;
+      stmts.push(...Stmt.fromAssignables(a, metadata));
+      if (stmts.length <= snap) {
+        throw new Error(`didn't generate any statements for arg ${a.t.trim()}`);
+      }
```
dfellis, line 223: Perhaps the error should be in `Stmt.fromAssignables`? If it doesn't generate *anything* that can't be a valid assignable statement, right?

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+      } else {
+        console.log(dec);
+        throw new Error(`got unexpected statement for arg ${a.t.trim()}: ${dec}`);
+      }
+    });
+
+    let fns = metadata.scope.deepGet(fnName);
+    if (fns === null) {
+      fns = [];
+    }
```
dfellis, line 238: I don't think this is necessary `null instanceof Array` returns false so it shouldn't try to do the length check below.

compiler/src/lnntoamm/Fn.ts
cdmistman: the exit statements do contribute to the return type (look at `Exit.fromAst` - you'll see `metadata.retConstraints.push(...)`). However, you're right - instead of defaulting to `void` when there's no return type specified, I should default to `Type.oneOf([Type.generate(), opcodes().get('void')])` so that it prefers the `void` type (since there is no return) but will get constrained to a non-void return type once the constraints get processed.

compiler/src/lnntoamm/Fn.ts
cdmistman: I thought we're doing if statements, not if expressions?

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    } else if (dec instanceof Ref) {
+      const upTy = dec.ty;
+      dec = new Dec(
+        ast,
+        mutable,
+        name,
+        ty,
+        dec,
+      );
+      (dec as Dec).ty.constrain(upTy);
```
dfellis, line 498: Is this explicit type override actually necessary here? You literally declare a `new Dec` right before this?

compiler/src/lnntoamm/Statement.ts
cdmistman: that'll still work - the generated type for `foo` will initially be `Maybe<Generated>`, then the assignments will constrain the generated type to `oneOf([float32, float64, int8, int16, int32, int64])` in your first example and both reassignments will constrain it to `oneOf([float32, float64])` in your second example

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    throw new Error(`declarations can't be used as expressions`);
+  }
+
+  inline(amm: Output) {
+    const name = this.ammName;
+    let ty: Builtin;
+    try {
+      ty = this.ty.breakdown();
+    } catch (e) {
+      console.log('~~~', this);
```
dfellis, line 531: Delete this?

compiler/src/lnntoamm/Statement.ts
cdmistman: `wholeAst` is only used for eventual error reporting - see lines 100-109

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    if (argTy === null) {
+      TODO('args with implicit types');
+    } else if (!(argTy instanceof Type)) {
+      throw new Error(`Function argument is not a valid type: ${typename.t}`);
+    }
+    const arg = new FnArg(ast, name, argTy);
+    metadata.define(arg);
+    const dec = metadata.var(arg.name);
+    if (dec !== arg) {
+      throw new Error('ugggghhhhh');
```
dfellis, line 590: Is this expected to ever be hit? If so, slightly better error message needed :)

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    }
+    return arg;
+  }
+
+  ammOut(): [string, Builtin] {
+    return TODO('TODO:')
```
dfellis, line 596: Better TODO here?

compiler/src/lnntoamm/Statement.ts
```diff
@@ -0,0 +1,785 @@
...
+    return stmts;
+  }
+
+  pushAMM(indent: string, output: string) {
+    output.concat(
+      indent,
+      'emit ',
+      this.event.ammName,
+      ' ',
+      this.emitVal.ammName,
```
dfellis, line 645: Maybe only include these two if the event has a value to emit and isn't just a `void` event?

compiler/src/lnntoamm/Statement.ts
cdmistman: I don't want that to happen though - if there aren't any functions in the module scope there could still be a closure defined in the current fn scope that I'd have to find.

compiler/src/lnntoamm/Statement.ts
cdmistman: you would think? but no, unfortunately TS was giving me a type error :(

compiler/src/lnntoamm/Statement.ts
cdmistman: no it's not, I'll go ahead and delete

compiler/src/lnntoamm/Statement.ts
cdmistman: that's for one of my extensions that highlights `TODO:` but not `TODO` - this shouldn't get called at all (it's from a previous abstract fn that I defined but didn't get rid of) - will delete

compiler/src/lnntoamm/Types.ts
```diff
@@ -0,0 +1,290 @@
...
+    return Interface.fromAst(ast, scope);
+  }
+
+  static fromTypesAst(ast: LPNode, scope: Scope): Type {
+    return Struct.fromAst(ast, scope);
+  }
+
+  static generate(): Type {
+    return new Generated();
+  }
```
dfellis, line 42: Is this supposed to be replaced by the other classes? Can it be declared `abstract` too? Or should we just leave a comment?

cdmistman: Not sure what you mean? This is just a generated type - a type that hasn't been declared or reduced to a concrete type yet.

compiler/src/lnntoamm/opcodes.ts
```diff
@@ -0,0 +1,66 @@
...
+    void: [],
+    int8: [],
+    int16: [],
+    int32: [],
+    int64: [],
+    float32: [],
+    float64: [],
+    bool: [],
+    string: [],
+  }).forEach(([name, generics]: [string, string[]]) => {
```
dfellis, line 28: Am I reading this correctly? Are you destructuring one argument into two and typing them in the arguments declaration of this anonymous function?

compiler/temp.js
```diff
@@ -0,0 +1,17 @@
+const r = require('alan-js-runtime')
```
dfellis, line 1: Delete this file

compiler/temp.lnn
```diff
@@ -0,0 +1,7 @@
+from @std/app import start, exit, stdout
```
dfellis, line 1: Delete this file

compiler/src/lnntoamm/opcodes.ts
cdmistman: `Object.entries` returns a list of tuples, so in this case it's `[string, any[]][]`. The only the argument's type declaration is doing is specifying that the value of each key isn't just `any[]`, it's actually `string[]`. Otherwise, I'm just destructuring the tuple `[key, value]`

compiler/src/lnntoamm/Fn.ts
dfellis: Oh, I mean that `if myConditional { <fnBody> }` is composed of `ifLiteral` `conditionalCheck` `fn` where the block-looking function form is allowed, but it could also be `if myConditional someFunction` or even (though silly) `if myConditional fn (): retType { ... }` or `if myConditional fn () = myRetVal`

compiler/src/lnntoamm/Statement.ts
dfellis: I just mean it's de-centralizing extraction of the function call logic, so it doesn't seem DRY?

compiler/src/lnntoamm/Statement.ts
dfellis: Well, I mean the next conditional then checks if the length is zero so you still hit the error path, don't you?

compiler/src/lnntoamm/Types.ts
dfellis: Got it. I didn't understand what this is at first. Calling "generate" on a type sounds like it should do something, not create a type constraining object.

compiler/src/lnntoamm/Fn.ts
cdmistman: right - would this allow for things like this? if not, does it make sense to require that the fn returns `void` to avoid confusion?
```alan
fn add1(n: int64) = add(1, n);
let x = if y == 0 add1(1) else add1(y);
```

compiler/src/lnntoamm/Statement.ts
cdmistman: yes, but I have the TODO on the line before the empty-array-check that'll possibly populate the 0-length-array with a single value. I just have to complete the `Closure` class before I can implement that `TODO`

compiler/src/lnntoamm/Types.ts
cdmistman: the object isn't intended to constrain another type - instead, it's simply a way of converting
```alan
let x = 'hello, world!';
```
into
```alan
let x: T = 'hello, world!';
```
which will allow `T` to get reduced into some type `U` once additional constraints are added. In this case, `T` is constrained to a `string` type, so internally it will delegate to the `string` Type. I couldn't think of a better way to describe this than "generate a type", so that's why I went with `generate`

compiler/src/lnntoamm/Event.ts
cdmistman: added a comment to clarify that the handlers have to be filtered to ensure that they accept the event type and return `void`

compiler/src/lnntoamm/Fn.ts
dfellis: Not exactly, but this:

```ln
fn myFn(x: int64): Result<int64> {
  fn add1(n: int64) = add(1, n);
  if n == 0 add1(1) else add1(x)
}
```

Would work? The return type of the inner function becomes the return type of the outer function.

dfellis: That makes it identical in behavior to:

`fn myFn(x: in64): Result<int64> {
  if x == 0 {
    return ok(2);
  } else {
    return x + 1;
  }
}
```

File: compiler/src/lnntoamm/opcodes.ts (hunk 31 above)
Comment (dfellis) [reply compiler/src/lnntoamm/opcodes.ts:28]: Got it. I didn't notice the `entries` usage.

File: std/root.lnn [hunk 34]
```diff
@@ -0,0 +1,19 @@
+// note: most-recently-defined wins as tie-breaker
```
depombo, line 1: don't we have multiple dispatch or is there a case when a function has the same name and arguments?

depombo (approved):

std/root.lnn
dfellis: Even when they have different argument types it could come into play. If one function has an interface type and another a *different* interface type or an actual type and they both match, choosing the most-recently-defined makes it possible for users to override behavior from imported modules while anything else becomes effectively "unpredictable" (if it's based on which is the "best match" the developer needs to know those rules, and there still needs to be a tiebreaker rule there, too).

compiler/src/lnntoamm/Ast.ts
```diff
@@ -0,0 +1,258 @@
...
+  // Essentially, there are two recursively-found directories that global modules can be found,
+  // the `modules` directory and the `dependencies` directory (TBD: are these the final names?)
+  // The `modules` directory is recursively checked first (with a special check to make sure it
+  // ignores self-resolutions) and the first one found in that check, if any, is used. If not,
+  // there's a special check if the dependency is an `@std/...` dependency, and if so to return
+  // that string as-is so the built-in dependency is used. Next the same recursive check is
+  // performed on the `dependencies` directories until the dependency is found. If that also
+  // fails, then there will be a complaint and the process will exit.
+  //
+  // The idea is that the package manager will install dependencies into the `dependencies`
```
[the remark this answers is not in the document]
dfellis: I think this was just copy-pasted from the current `lntoamm/Ast.ts` file?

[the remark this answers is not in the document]
cdmistman: yup - I don't intend to change dependency management

std/root.lnn
cdmistman: this is behavior that is unchanged from the current front-end. I just left this comment here to remind myself :)

dfellis (approved):

PR #484, opened 2021-04-08
Protobuf memory handler

aguillenv:


avm/src/vm/memory.rs
```diff
@@ -883,4 +886,127 @@ impl HandlerMemory {
     let s = str::from_utf8(&s_bytes[0..s_len]).unwrap();
     s.to_string()
   }
+
```
depombo, line 889: maybe serde might make this mapping simpler?
https://docs.rs/serde-protobuf/0.8.1/serde_protobuf/
https://serde.rs

aguillenv: Taking a look to https://docs.rs/serde-protobuf/0.8.1/serde_protobuf/ they do not have serialization implemented yet.
With the current crate would be possible to add `serde` derive but it add it over the autogenerated struct based on `.proto` schema so that will not simplify that mucho, we still will need to make the double conversion work if I'm not wrong.

aguillenv: Probably we can add it and leave it if someday it is necessary to pass from the `proto` struct to a json?

dfellis (approved):

depombo (approved):

PR #485, opened 2021-04-13
sample vm stats for 100ms every min and send every ~5min

depombo:


avm/src/daemon/stats.rs
```diff
@@ -40,15 +40,15 @@ pub struct VMStatsV1 {
 }
 
 // calculate the cpu times usage per process using the process'
-// total cpu time delta in a given time window
+// total cpu time delta in a 100ms time window
 async fn get_proc_usages() -> Vec<f64> {
-  let duration = 1.0;
+  let duration = 0.1; // seconds
```
dfellis, line 45: Nit: why not make this `100; // miliseconds` and then not convert the duration below on line 51 and also use it on line 124?

dfellis (approved):

PR #486, opened 2021-04-14
bring the anycloud cli into the monorepo

depombo:


anycloud/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2020 Alan Technologies, Inc
```
dfellis, line 189: 2021, and probably should've always been that, since we started this year?

depombo: Ah yes!

dfellis (approved):

aguillenv (approved):
Not a huge fan of monorepos but since the job it's done I'll approve. I think interoperability issues might still happen since we still need to check the `alan` version on `PATH` and which `alan` version is being used by the daemon.

depombo:
> Not a huge fan of monorepos but since the job it's done I'll approve. I think interoperability issues might still happen since we still need to check the `alan` version on `PATH` and which `alan` version is being used by the daemon.

Do you mean anycloud cli instead of the daemon?

aguillenv:
> > Not a huge fan of monorepos but since the job it's done I'll approve. I think interoperability issues might still happen since we still need to check the `alan` version on `PATH` and which `alan` version is being used by the daemon.
> 
> Do you mean anycloud cli instead of the daemon?

I mean both, anycloud cli works with current alan version on local `PATH` and will need to be aligned with `daemon` alan version, which is the latest release, no?

depombo:
> > > Not a huge fan of monorepos but since the job it's done I'll approve. I think interoperability issues might still happen since we still need to check the `alan` version on `PATH` and which `alan` version is being used by the daemon.
> > 
> > 
> > Do you mean anycloud cli instead of the daemon?
> 
> I mean both, anycloud cli works with current alan version on local `PATH` and will need to be aligned with `daemon` alan version, which is the latest release, no?

So because we now release them together and we build Alan first, Anycloud will have Alan in the `PATH` and they both simply have the same version. The daemon version still needs to be passed, but that becomes the only bit of interop.

depombo:
> > > > Not a huge fan of monorepos but since the job it's done I'll approve. I think interoperability issues might still happen since we still need to check the `alan` version on `PATH` and which `alan` version is being used by the daemon.
> > > 
> > > 
> > > Do you mean anycloud cli instead of the daemon?
> > 
> > 
> > I mean both, anycloud cli works with current alan version on local `PATH` and will need to be aligned with `daemon` alan version, which is the latest release, no?
> 
> So because we now release them together and we build Alan first, Anycloud will have Alan in the `PATH` and they both simply have the same version and for the prerelease we download Anycloud bin directly instead of building it from the latest Alan. The daemon version still needs to be passed, but that becomes the only bit of interop?

Talked about it offline. The plan is to try to squash the remaining bits of interop that Alejandro is pointing out that still remain after this PR.

PR #488, opened 2021-04-14
pass the cli version on every request

depombo:
This will allow handling unsupported CLI version errors


dfellis (approved):
I love net-negative PRs that improve functionality at the same time! :)

PR #489, opened 2021-04-14
Split out the control port to a separate file and more tightly integerate with LogRendezvousHash

dfellis:


avm/src/daemon/ctrl.rs
```diff
@@ -0,0 +1,200 @@
...
+    // If this is an Anycloud deployment and the child process hasn't started, mark as a failure
+    // TODO: Any way to generalize this so we don't have special logic for Anycloud?
+    Ok(Response::builder().status(500).body("fail".into()).unwrap())
+  } else {
+    // Everything passed, send an ok
+    Ok(Response::builder().status(200).body("ok".into()).unwrap())
+  }
+}
+
+mod naive {
```
depombo, line 127: we are not putting this in its own file to avoid circular dependencies, right? if so, can we leave a comment about it?

avm/src/daemon/ctrl.rs
```diff
@@ -0,0 +1,200 @@
...
+
+pub struct ControlPort {
+  lrh: LogRendezvousHash,
+  client: Client<HttpsConnector<HttpConnector>>,
+  // TODO: Once the crazy type info of the server can be figured out, we can attach it to this
+  // struct and then make it possible to wind down the control port server
+  // server: &'a dyn Service<std::convert::Infallible>,
+}
+
+async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible> {
```
depombo, line 105: maybe `health` is a better name now?

avm/src/daemon/ctrl.rs
dfellis: Oh, the [original code I based this on](https://github.com/ctz/rustls/blob/51961e1e53380441fb49efdd0e688181a9b275be/rustls-mio/examples/tlsclient.rs#L449-L468) did it that way. I just refactored it until I got it up-to-date and working. I can remove the inner `mod` if you prefer.

avm/src/daemon/ctrl.rs
dfellis: So I split the cluster routing/health task in two. The very next PR would add the new `/ping` endpoint so I'd prefer to keep it named this way?

depombo: sounds good!

avm/src/daemon/ctrl.rs
depombo: got it, definitely not something blocking / can happen in the next PR. maybe that is idiomatic rust I am just unfamiliar with it so it looks a bit weird

depombo (approved):

aguillenv (approved):

.version.sh
```diff
@@ -34,8 +34,8 @@ cd -
 # Commit and tag the update
 git add js-runtime/package.json js-runtime/yarn.lock compiler/package.json compiler/yarn.lock avm/Cargo.toml avm/Cargo.lock anycloud/cli/Cargo.toml anycloud/cli/Cargo.lock
```
depombo, line 35: nit: should we call script `.release.sh` ?

Makefile
```diff
@@ -58,4 +66,8 @@ uninstall:
 
 .PHONY: version
```
depombo, line 67: nit: same as above, maybe just calling this `release`?

.github/workflows/release.yml
```diff
@@ -93,35 +67,30 @@ jobs:
...
+        file: ./alan-windows.zip
         asset_name: alan-windows.zip
-        asset_content_type: application/zip
 
     - name: Generate Anycloud zip
       run: |
         cd anycloud/cli
         cargo build --release
-        cd ..
+        cd ../..
```
depombo, line 81: nice, thanks for fixing this

.prerelease.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# Be very *vexing* with the output, but good for debugging if something goes wrong
```
depombo, line 3: 😂

.prerelease.sh
```diff
@@ -0,0 +1,38 @@
...
+
+cd avm
+cargo build
+cd -
+
+# Commit and tag the update
+git add js-runtime/package.json js-runtime/yarn.lock compiler/package.json compiler/yarn.lock avm/Cargo.toml avm/Cargo.lock anycloud/cli/Cargo.toml anycloud/cli/Cargo.lock
+git commit -m "v${VERSION}"
+git push origin $(git branch --show-current)
+gh release create v${VERSION} -t v${VERSION} -n Pre-release --target $(git branch --show-current) -p
```
depombo, line 38: amazing!

depombo (approved):

Makefile
dfellis: Discussed offline, but having the two with distinct names makes it harder to brainfart and type the wrong one in. Since anything published to npm is un-revocable, it would be good to prevent accidental publishing of beta code as a release.

PR #490, opened 2021-04-15
Add min/max optional number of replicas

aguillenv:


anycloud/cli/src/deploy/mod.rs
```diff
@@ -524,10 +528,43 @@ pub async fn add_deploy_config() {
       .with_prompt("Virtual machine type")
       .interact_text()
       .unwrap();
+    let min_replicas: Option<u32>;
```
cdmistman, line 531: in Rust it's preferred to instantiate variables wherever they're declared. So this line could be put after the next statement as:
```rust
let min_replicas: Option<u32> = if let Ok(replicas) = replicas.parse::<u32>() {
  Some(replicas)
} else { ... };
```

cdmistman: Another option would be something like this:
```rust
let min_replicas = replicas.parse::<u32>().unwrap_or_else(|_| {
  eprintln!(...);
  exit(1);
});
```

anycloud/cli/src/deploy/mod.rs
```diff
@@ -587,10 +624,43 @@ pub async fn edit_deploy_config() {
       .with_initial_text(config.vmType.to_string())
       .interact_text()
       .unwrap();
+    let min_replicas: Option<u32>;
```
cdmistman, line 627: same as above - i'm not sure why `min_replicas` is an `Option` anyways? it's never set to `None`...

aguillenv: Add it as `Option` to have "backwards compatibility", if in your current config theres no minReplicas or maxReplicas still works. Thats why I added it but yes, initially I thought it as pure `u32`, not and `Option`.

anycloud/cli/src/deploy/mod.rs
aguillenv: Got it

dfellis: Why not use the option map here? If it is `Some` it will transform the data and if it is `None` it skips your callback entirely.

anycloud/cli/src/deploy/mod.rs
depombo: More than backwards compatibility, it is an optional parameter so it parsing `anycloud.json` should not fail if it is missing.

depombo (approved):

anycloud/cli/src/deploy/mod.rs
cdmistman: it could be something like this?
```rust
let min_replicas: Option<u32> = replicas.parse::<u32>().into::<Option<_>>().unwrap_or_else(...);
```

aguillenv: Applied second suggestion

cdmistman (approved):
minmaxing is generally frowned upon in the RPG world, but I think it's good here

dfellis (approved):

cdmistman (approved):

anycloud/cli/src/deploy/mod.rs
```diff
@@ -543,6 +549,52 @@ pub async fn add_deploy_config() {
       break;
     }
   }
+  let prompt = if creds.len() > 1 {
+    "Minimum number of replicas per region or cloud"
```
dfellis, line 553: Nit: I don't know if "replicas" is how the end user is going to think of this. Maybe just `s/replicas/VMs/g` for all of the user-facing strings?

dfellis (approved):

anycloud/cli/src/deploy/mod.rs
aguillenv: sounds good

depombo:
Can we add a `how-to` guide in the docs?

cdmistman (approved):

PR #493, opened 2021-04-16
Add warn macro and set every user fault a warning

aguillenv:
paired with: https://github.com/alantech/deploy/pull/86

### DIFF
--- a/anycloud/cli/src/logger/mod.rs
+++ b/anycloud/cli/src/logger/mod.rs
@@ -24,12 +24,12 @@ pub enum ErrorType {
 macro_rules! error {
   ($errCode:expr, $($message:tt)+) => {async{
     eprintln!($($message)+);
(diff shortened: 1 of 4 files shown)


cdmistman (approved):

dfellis (approved):

anycloud/cli/src/logger/mod.rs
```diff
@@ -24,11 +24,24 @@ pub enum ErrorType {
 macro_rules! error {
   ($errCode:expr, $($message:tt)+) => {async{
     eprintln!($($message)+);
-    $crate::deploy::client_error($errCode, &format!($($message)+)).await;
+    $crate::deploy::client_error($errCode, &format!($($message)+), true).await;
```
depombo, line 27: can we just pass the level directly? that way we can also include info later

depombo (approved):

PR #495, opened 2021-04-19
POST ctrl port

aguillenv:
Solves https://github.com/alantech/meta/issues/180


avm/src/daemon/ctrl.rs
```diff
@@ -124,6 +140,66 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
...
+  let bytes = body::to_bytes(req.into_body()).await?;
+  let body: DaemonProperties = serde_json::from_slice(&bytes).unwrap();
+  might_dump_files(&body)?;
+  DAEMON_PROPS
+    .set(DaemonProperties {
+      clusterId: body.clusterId,
+      agzB64: body.agzB64,
+      deployToken: body.deployToken,
+      domain: body.domain,
+      dockerfileB64: body.dockerfileB64,
```
depombo, line 164: why did we change the capitalization? It is usually capitalized, but also this might cause a bug in other `deploy` endpoints and the CLI that still expect the `DockerfileB64` instead of `dockerfileB64`

avm/src/daemon/ctrl.rs
```diff
@@ -124,6 +140,66 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
...
+      appTarGzB64: body.appTarGzB64,
+      envB64: body.envB64,
+    })
+    .unwrap();
+  Ok(())
+}
+
+fn might_dump_files(daemon_props: &DaemonProperties) -> DaemonResult<()> {
```
depombo, line 172: nit: `maybe_dump_files`

avm/src/daemon/ctrl.rs
```diff
@@ -143,26 +219,44 @@ mod naive {
...
+          let mut http_connector = HttpConnector::new();
+          http_connector.enforce_http(false);
+
+          ControlPort {
+            lrh: LogRendezvousHash::new(vec![]),
+            client: Client::builder().build::<_, Body>(HttpsConnector::from((http_connector, tls))),
+          }
+        } else {
+          let err = "Failed getting ssl certificate or key";
+          error!(ErrorType::CtrlPortStartFailed, "{}", err).await;
```
depombo, line 251: this should probably be it's own error type

avm/src/daemon/ctrl.rs
aguillenv: Hmm I did it since it's true that it is capitalize, it is still a variable name? To avoid unnecessary change in this PR renaming it in several places will revert it.

avm/src/daemon/ctrl.rs
aguillenv: Sure

avm/src/daemon/ctrl.rs
```diff
@@ -124,6 +140,66 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
   }
 }
 
+async fn handle_start(req: Request<Body>) -> Result<Response<Body>, Infallible> {
+  // Receive POST and save daemon properties
+  match get_daemon_props(req).await {
+    Ok(_) => Ok(Response::builder().status(200).body("ok".into()).unwrap()),
+    Err(err) => {
+      error!(ErrorType::DaemonStartFailed, "{:?}", err).await;
```
cdmistman, line 148: ... we can do something really clever on the `macro_rules! error` so that we can just say `error!(DaemonStartFailed, ...)` ...

see <https://github.com/clap-rs/clap/blob/master/src/macros.rs#L321-L326>:
```rust
macro_rules! error {
  (...) => {
    let err = anycloud::error::ErrorType::$err;
    ...
  }
};
```

anycloud/cli/Cargo.toml
```diff
@@ -5,7 +5,7 @@ license = "Apache-2.0"
 homepage = "https://github.com/alantech/anycloud"
 documentation = "https://alantech.gitbook.io/anycloud"
 repository = "https://github.com/alantech/anycloud"
-version = "0.1.37-beta8"
+version = "0.1.37-beta26"
 authors = ["Luis de Pombo <luis@alantechnologies.com>", "David Ellis <david@alantechnologies.com>"]
```
dfellis, line 9: @cdmistman and @aguillenv you guys should add yourselves to this (and the avm Cargo.toml) authors array. :)

avm/src/daemon/ctrl.rs
```diff
@@ -107,8 +113,18 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
   if cluster_secret.is_some() && !req.headers().contains_key(cluster_secret.as_ref().unwrap()) {
     // If this control port is guarded by a secret string, make sure there's a header with that
     // secret as the key (we don't care about the value) and abort otherwise
-    Ok(Response::builder().status(500).body("fail".into()).unwrap())
-  } else if TcpStream::connect("127.0.0.1:443").is_err() {
+    return Ok(Response::builder().status(500).body("fail".into()).unwrap());
+  }
+  match req.uri().path() {
+    "/health" => Ok(Response::builder().status(200).body("ok".into()).unwrap()),
+    "/clusterHealth" => handle_cluster_health(),
```
dfellis, line 120: Why are these different?

avm/src/daemon/ctrl.rs
```diff
@@ -143,26 +219,44 @@ mod naive {
...
-
-    ControlPort {
-      lrh: LogRendezvousHash::new(vec![]),
-      client: Client::builder().build::<_, Body>(HttpsConnector::from((http_connector, tls))),
+  pub async fn start() -> ControlPort {
+    let pwd = env::current_dir();
+    match pwd {
+      Ok(pwd) => {
+        let priv_key_b64 = read(format!("{}/priv_key_b64", pwd.display()));
+        let cert_b64 = read(format!("{}/cert_b64", pwd.display()));
```
dfellis, line 227: These are being read from the environment variables but are not being set as environment variables in the other PR, so this won't work?

dfellis: Also, let's not put these in the environment variables, that would be a very bad idea.

avm/src/daemon/ctrl.rs
aguillenv: `/health` tell us control port is ready and `clusterHealth` checks for 443 port abd all that logic

aguillenv: / health is being used to know when control port is active to send the `POST` to start the VM

avm/src/daemon/ctrl.rs
aguillenv: It is getting it from file system

cdmistman (approved):

avm/src/daemon/ctrl.rs
```diff
@@ -124,6 +140,66 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
...
+  maybe_dump_files(&body)?;
+  DAEMON_PROPS
+    .set(DaemonProperties {
+      clusterId: body.clusterId,
+      agzB64: body.agzB64,
+      deployToken: body.deployToken,
+      domain: body.domain,
+      DockerfileB64: body.DockerfileB64,
+      appTarGzB64: body.appTarGzB64,
+      envB64: body.envB64,
```
dfellis, line 166: I would like to make a suggestion on these three properties.

Can we make them sub-properties on a `filesB64` object where the key is literally just the name of the file and the value is the base64-encoded data? Then we can generalize all of this to just: "here are the files you need to put on disk during startup for whatever reason needed by the `agz` payload".

This will let us eliminate all of the special hackery within the AVM for Anycloud and eventually give the power to include static files to anyone running `alan deploy new` / `anycloud new`.

aguillenv: Good idea, makes sense to me

avm/src/daemon/ctrl.rs
aguillenv: Like this, will try it out

aguillenv: Applied!

avm/src/daemon/daemon.rs
```diff
@@ -116,118 +142,150 @@ async fn post_v1_stats(
...
+          let dns = DNS::new(&domain);
+          if let Ok(dns) = &dns {
+            loop {
+              let vms = match dns.get_vms(&cluster_id).await {
+                Ok(vms) => vms,
+                Err(err) => {
+                  error!(NoDnsVms, "{}", err).await;
+                  Vec::new()
+                }
+              };
```
cdmistman, line 230: nit: can be `unwrap_or_else`

avm/src/daemon/daemon.rs
```diff
@@ -116,118 +142,150 @@ async fn post_v1_stats(
...
+                if factor != "1" {
+                  post_v1_scale(&cluster_id, &agz_b64, &deploy_token, &factor).await;
+                }
+              }
+              control_port.check_cluster_health().await;
+              sleep(period).await;
+            }
+          } else if let Err(dns_err) = &dns {
+            error!(NoDns, "DNS error: {}", dns_err).await;
+            panic!("DNS error: {}", dns_err);
```
cdmistman, line 273: shouldn't this just be
```suggestion
            error!(NoDns, "DNS error: {}", dns_err).await;
            std::process::exit(1);
```

avm/src/daemon/daemon.rs
```diff
@@ -116,118 +142,150 @@ async fn post_v1_stats(
...
           }
+        });
+        if let Err(err) = run_agz_b64(&agz_b64).await {
+          error!(RunAgzFailed, "{:?}", err).await;
+          panic!("{:?}", err);
```
cdmistman, line 278: same as above

avm/src/daemon/daemon.rs
```diff
@@ -116,118 +142,150 @@ async fn post_v1_stats(
...
         }
-        control_port.check_cluster_health().await;
-        sleep(period).await;
+      } else {
+        let msg = "No daemon properties defined";
+        error!(NoDaemonProps, "{}", msg).await;
+        panic!("{}", msg);
```
cdmistman, line 283: another one

avm/src/daemon/daemon.rs
```diff
@@ -116,118 +142,150 @@ async fn post_v1_stats(
...
       }
-    } else if let Err(dns_err) = &dns {
-      error!(ErrorType::NoDns, "DNS error: {}", dns_err).await;
-      panic!("DNS error: {}", dns_err);
-    } else if let Err(self_ip_err) = &self_ip {
-      error!(ErrorType::NoPrivateIp, "Private ip error: {}", self_ip_err).await;
+    }
+    Err(self_ip_err) => {
+      error!(NoPrivateIp, "Private ip error: {}", self_ip_err).await;
       panic!("Private ip error: {}", self_ip_err);
```
cdmistman, line 288: here too

depombo (approved):

avm/src/daemon/daemon.rs
```diff
@@ -24,6 +24,32 @@ use crate::vm::run::run;
...
+}
+
+#[allow(non_snake_case)]
+#[derive(Deserialize, Debug, Serialize)]
+pub struct DaemonProperties {
+  pub clusterId: String,
+  pub agzB64: String,
+  pub deployToken: String,
+  pub domain: String,
+  pub filesB64: Option<DaemonFilesB64>,
```
depombo, line 51: so this is weird because the agz is also base64 encoded can we rename `filesB64` to `optionalB64files` or something? This also complicates the data structure because of the options nested within an option.

avm/src/daemon/daemon.rs
```diff
@@ -24,6 +24,32 @@ use crate::vm::run::run;
...
+  pub content: String,
+}
+
+#[allow(non_snake_case)]
+#[derive(Deserialize, Debug, Serialize)]
+pub struct DaemonFilesB64 {
+  pub Dockerfile: Option<DaemonFileB64>,
+  pub appTarGz: Option<DaemonFileB64>,
+  pub envVariables: Option<DaemonFileB64>,
+}
```
depombo, line 42: This struct should probably not exist and `filesB64` is just a `Vec< DaemonFileB64>`?

dfellis: Better if it was a `HashMap<String, String>` where the key is the file name and the value is the file contents base64 encoded. Then we can support *any* files needed.

avm/src/daemon/daemon.rs
aguillenv: Hmm since we are already logging the error I think makes sense

avm/src/daemon/daemon.rs
aguillenv: I'll try this with the `HashMap`, did not thought about it the first time. This way we remove the struct complexity I'm adding.

avm/src/daemon/daemon.rs
aguillenv: But, how could I handle the `error!` since it's a future? I cannot define an `async` block inside a `unwrap_or_else`, right?

avm/src/daemon/daemon.rs
```diff
@@ -116,118 +128,157 @@ async fn post_v1_stats(
...
-      )
-      .await;
-    } else {
-      return Err("AGZ file appears to be corrupt.".into());
+    let pwd = env::current_dir();
+    match pwd {
+      Ok(pwd) => {
+        let priv_key = read(format!("{}/key.pem", pwd.display()));
+        let cert = read(format!("{}/certificate.pem", pwd.display()));
+        if let (Ok(priv_key), Ok(cert)) = (priv_key, cert) {
```
dfellis, line 144: Since the control port itself fails and exits the app, this is technically impossible, but supposing someone deleted the cert after spinning up the control port,  this `if let` could fail and the lack of an `else` on it means it would fall through to return an `Ok` while not actually spinning up the VM.

aguillenv: Good catch, solved

cdmistman (approved):

dfellis (approved):

PR #496, opened 2021-04-19
Resultify AVM

### LINKED ISSUE #472: Refactor error handling in the AVM
Stop throwing panics and instead return errors that the daemon can log and then panic on.

cdmistman:
fixes #472


avm/src/vm/memory.rs
```diff
@@ -591,7 +592,7 @@ impl HandlerMemory {
       if len == 0 {
         // Assume zero is not a string
         dest.write_fixed(dest_addr, mem_slice[0].1);
-        return;
+        return Ok(())
```
aguillenv, line 595: missign `;`?

avm/src/vm/event.rs
```diff
@@ -248,36 +259,44 @@ impl HandlerFragment {
...
-            //eprintln!("{} {:?}", i.opcode._name, i.args);
-            func(i.args.clone(), HandlerMemory::fork(hand_mem.clone()))
-              .then(HandlerMemory::drop_parent_async)
-          } else {
-            eprintln!("expected another IO instruction");
-            std::process::exit(1);
+          let hand_mem = hand_mem.clone();
+          async move {
+            if let OpcodeFn::Io(func) = i.opcode.fun {
+              //eprintln!("{} {:?}", i.opcode._name, i.args);
```
aguillenv, line 280: These comments are for future debugging?

avm/src/vm/protos/mod.rs
```diff
@@ -1 +1,3 @@
-// Dummy mod. Will be auto generated by protoc-rust
```
aguillenv, line 1: Hmmm  will think how to avoid this and be transparent for us

avm/src/vm/event.rs
cdmistman: yes - I didn't remove it since it's been there for a while (and could prove useful in the future)

avm/src/vm/protos/mod.rs
cdmistman: sounds good. I brought it back in the next commit, but I keep forgetting the 1 extra step after `git add .`

dfellis (changes requested):
Undo these changes

avm/src/vm/event.rs
```diff
@@ -8,21 +7,26 @@ use crate::vm::memory::HandlerMemory;
...
+#[repr(i64)]
 /// Special events in alan found in standard library modules, @std.
 /// The IDs for built-in events are negative to avoid collision with positive, custom event IDs.
 /// The first hexadecimal byte of the ID in an integer is between 80 and FF
 /// The remaining 7 bytes can be used for ASCII-like values
 pub enum BuiltInEvents {
   /// Alan application start
   /// '"start"' in ASCII or 2273 7461 7274 22(80)
-  START,
+  START = -9213673853036498142,
```
dfellis, line 26: What's going on here?

cdmistman: with this, we can use `BuiltInEvents::START as i64`

avm/src/vm/event.rs
```diff
@@ -204,25 +208,31 @@ impl HandlerFragment {
...
+            if let Some(event) = event? {
+              let event_tx = EVENT_TX.get().unwrap();
+              let event_sent = event_tx.send(event);
+              if event_sent.is_err() {
+                eprintln!("Event transmission error");
+                std::process::exit(2);
+              }
             }
+            Ok(())
+          } else {
```
dfellis, line 228: This whole branch is known (to us) to never happen because of the way the opcode fragments work. It would be nice to be able to assert this with the type during the fragment generation, perhaps having another enum for the "fragment type" wrapping around the vec and only allowing the particular opcode function enum we care about within each of these outer enums and being able to avoid this looping check here?

Not something for this PR, but it does bother me that we have to define an error path here that will never be triggered and that we have to pay a price doing these useless checks.

avm/src/vm/http.rs
```diff
@@ -57,7 +57,7 @@ macro_rules! make_server {
         let port_num = http.port;
         let addr = std::net::SocketAddr::from(([0, 0, 0, 0], port_num));
         let make_svc = hyper::service::make_service_fn(|_conn| async {
-          Ok::<_, std::convert::Infallible>(hyper::service::service_fn($listener))
+          Ok::<_, $crate::vm::VMError>(hyper::service::service_fn($listener))
```
dfellis, line 60: Why this change?

avm/src/vm/memory.rs
```diff
@@ -1,11 +1,15 @@
...
 
-use protobuf::{Message, ProtobufError, ProtobufResult};
+use byteorder::ByteOrder;
+use byteorder::NativeEndian;
 
 use crate::vm::program::Program;
 use crate::vm::protos;
+use crate::vm::InvalidState;
+use crate::vm::VMError;
+use crate::vm::VMResult;
```
dfellis, line 12: Nit: why not

```rs
use crate::vm::{program::Program, protos, InvalidState, VMError, VMResult};
```

avm/src/vm/memory.rs
```diff
@@ -1,11 +1,15 @@
-use std::convert::TryInto;
+use std::array::IntoIter;
 use std::str;
 use std::sync::Arc;
 
-use protobuf::{Message, ProtobufError, ProtobufResult};
+use byteorder::ByteOrder;
+use byteorder::NativeEndian;
```
dfellis, line 6: Same nit here.

avm/src/vm/memory.rs
```diff
@@ -136,15 +144,19 @@ impl HandlerMemory {
     event_id: i64,
     curr_addr: i64,
     curr_hand_mem: &Arc<HandlerMemory>,
-  ) -> Option<Arc<HandlerMemory>> {
-    let pls = Program::global().event_pls.get(&event_id).unwrap().clone();
+  ) -> VMResult<Option<Arc<HandlerMemory>>> {
```
dfellis, line 147: We're wrapping another layer around `HandlerMemory` here.

You know when we want to improve performance we're going to be undoing a lot of this and going back to "we have thoroughly tested/audited our code so we don't have to pay a runtime penalty for this safety" :/

avm/src/vm/memory.rs
```diff
@@ -876,15 +878,17 @@ impl HandlerMemory {
...
-    }
-    let s = str::from_utf8(&s_bytes[0..s_len]).unwrap();
-    s.to_string()
+    let s_bytes = f
+      .block
+      .iter()
+      .skip(1)
+      // TODO: `IntoIter::new` will be deprecated once `rust-lang/rust#65819` is merged
+      .flat_map(|(_, chars)| IntoIter::new(chars.to_ne_bytes()))
+      .collect::<Vec<u8>>();
```
dfellis, line 889: Why switch from the `for in` loop to this if it depends on discourage usage of a built-in type?

avm/src/vm/opcode.rs
```diff
@@ -222,2012 +224,2014 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    hand_mem.write_fixed(args[2], i64::from_ne_bytes(out.to_ne_bytes()))?;
+    Ok(None)
   });
   cpu!(boolf64 => fn(args, hand_mem) {
-    let out = hand_mem.read_fixed(args[0]) as f64;
-    hand_mem.write_fixed(args[2], i64::from_ne_bytes(out.to_ne_bytes()));
-    None
+    let out = hand_mem.read_fixed(args[0])? as f64;
+    hand_mem.write_fixed(args[2], i64::from_ne_bytes(out.to_ne_bytes()))?;
+    Ok(None)
```
dfellis, line 260: These opcodes sound very confused and unsure of themselves "read fixed argument 0? I guess? Then write it as a bool to argument 2? Ok, nothing else."

avm/src/vm/opcode.rs
```diff
@@ -222,2012 +224,2014 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    hand_mem.write_fractal(args[2], &HandlerMemory::str_to_fractal(&out_str))?;
+    Ok(None)
   });
   cpu!(matches => fn(args, hand_mem) {
-    let a_str = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[0]));
-    let b_str = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[1]));
-    let b_regex = Regex::new(&b_str).unwrap();
+    let a_str = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[0])?)?;
+    let b_str = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[1])?)?;
+    let b_regex = Regex::new(&b_str).map_err(|regex_err| VMError::Other(format!("Bad regex construction: {}", regex_err)))?;
```
dfellis, line 2137: This is good, but makes me wonder what we can/should do at compile time for this. A bad regex should be detectable during the compilation phase.

avm/src/vm/opcode.rs
```diff
@@ -3031,779 +3054,784 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    Ok(None)
   });
 
   // "Special" opcodes
   cpu!(exitop => fn(args, hand_mem) {
-    io::stdout().flush().unwrap();
-    io::stderr().flush().unwrap();
-    std::process::exit(hand_mem.read_fixed(args[0]) as i32);
+    let _ = io::stdout().flush();
+    let _ = io::stderr().flush();
```
dfellis, line 3344: I feel like these lines are missing `?`s

avm/src/vm/telemetry.rs
```diff
@@ -27,15 +27,11 @@ pub async fn log(event: &str) {
   });
   let client =
     Client::builder().build::<_, Body>(hyper_rustls::HttpsConnector::with_native_roots());
-  if client
+  let _ = client
```
dfellis, line 30: What's the assignment here for?

avm/src/vm/event.rs
cdmistman: I think there's a way to unify the function types of the different opcode kinds, so then we can entirely get rid of these checks and only use the additional information for building up the instruction fragments. I imagine it'd be a struct with 2 fields, and running them would be much faster.

The new opcode defs would look like:
```rust
pub enum OpcodeKind {
  CPU,
  IO,
  UnpredCPU,
}
pub struct Opcode {
  name: &'static str,
  kind: OpcodeKind,
  exec: fn(&[i64], Arc<HandlerMemory>) -> Box<Pin<VMResult<Arc<HandlerMemory>>>>,
}
```
so execution can be done with purely an iterator (simplified):
```rust
handler.instructions
  .iter()
  .map(|fragment| {
    fragment
      .iter()
      .map(|(op, instrs)| exec(&instrs, hmem.fork()))
      .collect::<UnorderedFutures<VMResult<Vec<Arc<HandlerMemory>>>>>>()
  })
  .collect::<OrderedFutures<VMResult<Vec<Arc<HandlerMemory>>>>>()
```

avm/src/vm/http.rs
cdmistman: the service function could return a `VMError` since it has to work with the memory of the event that handles each request, so I allow the hyper service to propagate such errors - although we could handle the error directly in the service function?

avm/src/vm/memory.rs
cdmistman: that makes sense, although many of these checks are happening anyways when we call `unwrap` or `expect`. This is just to allow us to prevent these panics and recover from them if the daemon is running.

avm/src/vm/memory.rs
cdmistman: the `for in` loop from before had a non-constant number of allocations - this iterator implementation has 1 allocation total, guaranteed (on `iter()` - `skip` doesn't allocate, `IntoIter::new` recasts the array from `chars.to_ne_bytes` to an iterator, and `collect` inherits the allocation from `iter()`)

cdmistman: (side note: `impl IntoIterator for [T; N]` will be stabilized soon, but it'll actually call `impl IntoIter for Slice`, which will result in dropping the array while returning the iterator on a slice of the array, which is a compiler error. In edition 2021, this will delegate to `array::IntoIter`, which won't be deprecated)

avm/src/vm/opcode.rs
cdmistman: unfortunately not :/ regex does some heap allocations that can't be done at compile time* - see <https://docs.rs/regex/1.4.5/src/regex/exec.rs.html>

\* allocations in const-fns is planned, per <https://github.com/rust-lang/const-eval/issues/20>

avm/src/vm/opcode.rs
cdmistman: I was thinking that we don't care about the `Result`s returned from flushing, but it does make sense to do so.

avm/src/vm/telemetry.rs
cdmistman: it ignores the `Result` returned, which is what we were doing before, but I can change it back?

avm/src/vm/http.rs
dfellis: So for the web server I would prefer it is infallible and if it encounters any `VMError` responses it returns a `500 Internal Server Error` to the end user with some sort of short message indicating an unexpected failure within the AVM itself? Simply crashing out will make it harder to track down what's going wrong, I think.

avm/src/vm/opcode.rs
dfellis: Oh, I was thinking something like "Should we have a fake rust file that we inject the users' regex into and test if it parses correctly and just emit an error during compilation when the regex string is static"

But maybe the correct answer here is: "the regex API in Alan needs to be changed to be a `Result` and we return an Error to the user if it fails to build correctly"

avm/src/vm/telemetry.rs
dfellis: I was actually better with your refactoring, but I didn't understand why we needed the assignment at all. Just removing the `let _ =` means it won't compile?

avm/src/vm/opcode.rs
cdmistman: ah I see - we could also ship `ripgrep` with our compiler binary, and when we detect regex usage we
1. create an empty file
2. run `rg` on that empty file using the user's regex
3. continue compiling if the exit code is 0 or throw a compiler error if the exit code is non-zero

We could also shave `regex` off our dependency tree and save some space in the binary size by just shipping `ripgrep` in the avm regardless of if the compiler is there, and every time the user wants to do a regex query we do the same but on a file containing the input.

cdmistman: In all seriousness though, we *might* be able to pre-compile the Regex in the compiler by calling `regex` as a wasm module, store the object as a `Buffer`, and include it in the aga/agc output, but that would take a solid amount of time to get working...

cdmistman: actually it's not worth it to store the memory representation of `regex::Regex`, but we could still run it from wasm to validate the input

dfellis: I think all of this makes it clear that we need to change the regex-related APIs in Alan to be `Result`-based so it can provide the failure at runtime. At least the user will need to have defined some code to handle what happens if it can't parse the regex, though that is almost certainly going to be a logic bug. :/

avm/src/main.rs
```diff
@@ -131,7 +134,10 @@ fn main() {
           agc_file
         );
         telemetry::log("avm-run").await;
-        run_file(&fp, false).await;
+        if let Err(ee) = run_file(&fp, false).await {
+          eprintln!("{}", ee);
+          std::process::exit(2);
```
aguillenv, line 139: Why exit code `2` and not `1`?

avm/src/vm/opcode.rs
```diff
@@ -2992,32 +3013,33 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
     let (tx, rx): (Sender<Arc<HandlerMemory>>, Receiver<Arc<HandlerMemory>>) = oneshot::channel();
     let tx_ptr = Box::into_raw(Box::new(tx)) as i64;
-    event.push_fixed(0, tx_ptr);
+    event.push_fixed(0, tx_ptr)?;
     let event_emit = EventEmit {
       id: i64::from(BuiltInEvents::HTTPCONN),
       payload: Some(event),
     };
-    let event_tx = EVENT_TX.get().unwrap();
+    let event_tx = EVENT_TX.get().ok_or(VMError::ShutDown)?;
```
aguillenv, line 3042: Hmmm interesting, does this work as `unwrap_or`?

avm/src/vm/opcode.rs
```diff
@@ -3031,779 +3053,795 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
     Box::pin(async move {
-      make_server!(&Program::global().http_config, http_listener);
-      return hand_mem;
+      // this extra fn is so that we can just use `?` inside of http_listener instead of
+      // having a bunch of `match`es that call a closure
+      async fn listen(req: Request<Body>) -> Result<Response<Body>, Infallible> {
+        match http_listener(req).await {
+          Ok(res) => Ok(res),
+          Err(_) => {
+            // TODO: log the error?
```
aguillenv, line 3088: the chances to go through this brach is quite small right? If would represent an important error and difficult to track might be interesting to log it

aguillenv (approved):

avm/src/vm/event.rs
dfellis: So the CPU-bound opcodes can return an event payload, they don't return the HandlerMemory.

However, I think that weirdness is used only by the `emitop` opcode, so if we can refactor things to let event emission be done specially by that opcode and not require checking the return value, that'd be great. But what I'd prefer is that CPU opcodes still have a special return type, it's just that it's now `void` and we don't have to do any conditional checking on every single CPU opcode call, which is probably a significant perf impact.

depombo (approved):
LGTM once all tests are passing (only the AVM style right now?) and we verify there is no change in behavior/functionality for the AVM

avm/src/main.rs
cdmistman: that was what we were using in the majority of cases where we did `eprintln!()` and then `exit()`

avm/src/vm/opcode.rs
cdmistman: no, since `ok_or` still returns the `Result`, whereas `unwrap_or` is the same as `getOr` in Alan (if it's an `Err`, returns an `Ok` with the given default value)

avm/src/vm/opcode.rs
cdmistman: yes, but I'll let the daemon error-catching PR handle that

dfellis (approved):

avm/src/vm/opcode.rs
aguillenv: Got it

dfellis:
@cdmistman it appears that there was another accidental behavior change here, the double-`send` guard on http server responses apparently broke: https://github.com/alantech/alan/pull/496/checks?check_run_id=2403147748#step:4:1373

cdmistman:
> there was another accidental behavior change here, the double-`send` guard on http server responses apparently broke

As mentioned offline, this is from observing the results from the JS runtime, indicating that there isn't a change in behavior, but rather a race condition in our http tests.

Issue #478, opened 2021-04-07
Datastore LogRendezvousHash Primary + Copies Assignment


dfellis:
Using LogRendezvousHash, our modification of RendezvousHash to limit the number of redundant hashes computed, in order to determine which node in the cluster should be the primary owner of a key in the Datastore hash, and which nodes should maintain backup copies.

Like Consistent Hashing, new nodes added to the cluster cause only 1/n keys to be re-assigned, and the prior primary node would push to the new primary node in this situation, while nodes dropping cause an existing backup node to promote itself to primary and start pushing state to the new set of secondary nodes.

dfellis:
Traditional RendezvousHash uses the following algorithm to determine the primary and replica nodes in the network:

1. Concatenate the key to each of the ids of the nodes in the network.
2. Hash these strings and sort the results.
3. Take the top `n` records with the first one being the primary and the remaining `n - 1` being the replicas.

This depends on the hashing function being as cheap as possible, but it scales linearly with the size of the cluster, so cluster routing performance *drops* as cluster size increases through no fault of the individual nodes themselves.

ConsistentHash resolves this by having the concept of a "hash ring". You hash the cluster nodes `n` times and determine the many places on the ring that the node belongs to (using different hashing functions to get different locations, or some other way to quasi-randomize the assignment across the ring). This is done out-of-band of the actual routing path. During routing, you just hash the key to find its place in the ring and use some mechanism to find the "right" node (nearest or largest-not-going-over or etc) and then route to that one for the primary key. Copies can be determined by running other hash functions (or concatenate a static string to the key and re-hash, etc) but extra logic has to be added to deal with duplicates in the replica list not necessary for RendezvousHash. The routing path doesn't scale with the number of nodes, though, so performance is not affected by cluster size in the main path (updating the ring does scale with the number of nodes, but as long as the cluster isn't constantly scaling up and down, it's far below the network traffic load and not that important).

LogRendezvousHash is an attempt to find a middle-ground and still retain the `1/n` data churn property when a node is added or removed. The reason it works for RendezvousHash without explicitly seeding a ring data structure with virtual nodes is because the randomizing source of the key data itself in the hash provides an approximately even distribution.

If we create a binary tree that the nodes are added to with the nodes all being leaf nodes, and a hash determines which child you choose, then you can do approximately `log2(n)` hashes to reach a leaf node, using the node's distance from the parent as the string value to concatenate to the key and using `hash size / 2` as the toggle between `left` and `right` node.

[...]

PR #498, opened 2021-04-21
Rendezvous hash node assignment

aguillenv:
Related issue: https://github.com/alantech/alan/issues/478


avm/src/daemon/ctrl.rs
```diff
@@ -76,9 +76,32 @@ impl LogRendezvousHash {
...
+        .map(|hash| &hash.id as &str)
+        .collect::<Vec<&str>>();
+    }
+    let mut idx = self._get_idx_for_key(key);
+    let mut ids: Vec<&str> = Vec::new();
+    for _ in 0..top {
+      ids.push(&self.sorted_hashes[idx].id);
+      idx = (idx + 1) % self.sorted_hashes.len();
+    }
+    ids
```
dfellis, line 99: This will contain a node twice if the cluster size is 2.

While short-circuiting for a single node is a decent idea, it's also saving compute time on the cheapest path, anyways, so I don't know if we need to worry about that special casing, maybe we can replace this with something like:

`pub fn get_assigned_nodes_id(&self, key: &str) -> Vec<&str> {
  let top = cmp::min(3, self.sorted_hashes.len());
  let mut idx = self.get_idx_for_key(key);
  let mut ids = Vec::new();
  for _ in 0..top {
    ids.push(&self.sorted_hashes[idx].id);
    idx = (idx + 1) % self.sorted_hashes.len();
  }
  ids
}
```

File: avm/src/daemon/ctrl.rs [hunk 2]
```diff
@@ -76,9 +76,32 @@ impl LogRendezvousHash {
...
+  }
+
   // Runs a binary search for the record whose hash is closest to the key hash without
   // going over. If none are found, the *last* record in the list is returned as it wraps around.
-  pub fn _get_id_for_key(&self, key: &str) -> &str {
+  fn _get_idx_for_key(&self, key: &str) -> usize {
```
dfellis, line 104: Functions that are in use (this one as `_get_assigned_nodes_id` above) don't need the prefixed underscore because they are being used, so can we remove that?

avm/src/daemon/ctrl.rs
aguillenv: Oh yes, my bad that I did not handle that case. This looks cleaner!

avm/src/daemon/ctrl.rs
aguillenv: Yes, of course

dfellis (approved):

PR #501, opened 2021-04-27
Self load balancing (and complete reworking of the intra-cluster client)

dfellis:


aguillenv (approved):

.github/workflows/release.yml
```diff
@@ -60,6 +60,12 @@ jobs:
       with:
         node-version: 14.x
 
+    # The AVM requires openssl dev files to build now, and apparently the rust openssl code can auto-acquire it through vcpkg, so set that up
+    - name: Set up vcpkg
+      run: |
+        git clone https://github.com/microsoft/vcpkg
```
depombo, line 66: should we pin to a specific tag?

anycloud/cli/src/logger/mod.rs
```diff
@@ -18,6 +18,7 @@ pub enum ErrorType {
   ScaleFailed = 120,
   PostFailed = 121,
   RunAgzFailed = 122,
+  DuplicateDnsPrivateIp = 123,
```
depombo, line 21: we need to duplicate this in `deploy`. hmm maybe there is a way to get rid of this interop

avm/src/daemon/ctrl.rs
```diff
@@ -169,25 +188,120 @@ impl ControlPort {
...
+      // TODO: Should this error propagate up to the stats loop or no?
+      return;
+    } else if self_vm_vec.len() > 1 {
+      // This hopefully never happens, but if it does, we need to change the daemon initialization
+      error!(
+        ErrorType::DuplicateDnsPrivateIp,
+        "Private IP address collision detected! I don't know who I really am!"
+      )
+      .await;
+      // TODO: Should this error propagate up to the stats loop or no?
```
depombo, line 260: hmmm that is a good point

avm/src/vm/opcode.rs
```diff
@@ -2982,16 +2988,107 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
   });
 
   async fn http_listener(req: Request<Body>) -> VMResult<Response<Body>> {
+    // Grab the headers
+    let headers = req.headers();
+    // Check if we should load balance this request
+    if !headers.contains_key("x-alan-rr") {
```
depombo, line 2994: can we expand the comment a bit to explain how the load balancing works with this secret header?

depombo (approved):

avm/src/daemon/ctrl.rs
```diff
@@ -169,25 +188,120 @@ impl ControlPort {
...
+      Some(self_vm) => self.lrh.get_leader_id() == self_vm.private_ip_addr,
+      None => false,
+    }
+  }
+
+  pub fn get_leader(self: &ControlPort) -> Option<&VMMetadata> {
+    match self.vms.get(self.lrh.get_leader_id()) {
+      Some(vm) => Some(&vm),
+      None => None,
+    }
```
cdmistman, line 304: nit: could be
```suggestion
    self.vms.get(self.lrh.get_leader_id()).as_ref()
```

cdmistman (approved):

.github/workflows/release.yml
dfellis: So this is a Microsoft repository for a Microsoft compiler and Microsoft operating system. [I am also following the official install instructions](https://docs.microsoft.com/en-us/cpp/build/install-vcpkg?view=msvc-160&tabs=windows) so I think this is fine?

anycloud/cli/src/logger/mod.rs
dfellis: `deploy` could have a build script to extract this definition from the Alan repo, or we could potentially have some sort of common definition (likely still in this repo, but easier than parsing Rust code and getting the relevant data).

avm/src/vm/opcode.rs
dfellis: Sure

avm/src/daemon/ctrl.rs
dfellis: Wouldn't that be `&Option<VMMetadata>` instead of `Option<&VMMetadata>` though?

cdmistman: no - `as_ref` inverts the reference so it turns a `&Option<VMMetadata>` into a `Option<&VMMetadata>`

anycloud/cli/src/logger/mod.rs
depombo: Yeah, probably just a JSON file here

dfellis: > Yeah, probably just a JSON file here

The issue is turning that JSON file into an enum in Rust will also be difficult.

PR #502, opened 2021-04-28
Saturating Math

dfellis:
Finishing off RFC 004.


std/root.ln
```diff
@@ -2762,6 +3337,11 @@ export infix split as / precedence 3
 export infix mod as % precedence 3
 // export infix template as % precedence 3
 export infix pow as ** precedence 4
+export infix sadd as +. precedence 2
+export infix ssub as -. precedence 2
+export infix smul as *. precedence 3
+export infix sdiv as /. precedence 3
```
dfellis, line 3343: [/.](https://slashdot.org)

aguillenv (approved):

depombo (approved):

avm/src/vm/opcode.rs
```diff
@@ -1522,6 +1522,364 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    if a > 0 && b > 0 && a > std::i8::MAX - b {
+      hand_mem.write_fixed(args[2], std::i8::MAX as i64)?;
+      return Ok(None);
+    }
+    if a < 0 && b < 0 && a < std::i8::MIN - b {
+      hand_mem.write_fixed(args[2], std::i8::MIN as i64)?;
+      return Ok(None);
+    }
+    hand_mem.write_fixed(args[2], (a + b) as i64)?;
+    Ok(None)
```
cdmistman, line 1538: why not use rust's [saturating add](https://doc.rust-lang.org/std/primitive.i8.html#method.saturating_add)?

dfellis: 1. I forgot about it.
2. Particularly for division, we need to make sure it has the same semantics as here (true of all of them, but unlikely to differ for adding and subtracting)

cdmistman: there's no saturating division implementation so I think it makes sense to maintain the manual implementation

avm/src/vm/opcode.rs
```diff
@@ -1522,6 +1522,236 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
     Ok(None)
   });
 
+  // Saturating Arithmetic opcodes
+  cpu!(saddi8 => fn(args, hand_mem) {
```
cdmistman, line 1526: don't be sad 😢

avm/src/vm/opcode.rs
```diff
@@ -1522,6 +1522,236 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    Ok(None)
+  });
+  cpu!(smulf32 => fn(args, hand_mem) {
+    let a = f32::from_ne_bytes((hand_mem.read_fixed(args[0])? as i32).to_ne_bytes());
+    let b = f32::from_ne_bytes((hand_mem.read_fixed(args[1])? as i32).to_ne_bytes());
+    let out = a * b;
+    let num = i32::from_ne_bytes(out.to_ne_bytes()) as i64;
+    hand_mem.write_fixed(args[2], num)?;
+    Ok(None)
+  });
```
cdmistman, line 1639: does it make sense to have saturating versions for floats since they're not at all different from the original versions?

dfellis: They are different, though? The "normal" floating point math opcodes actually simulate the integer error failures -- hitting `Infinity` or `Div-By-Zero` causes an Error to be returned instead of a Result-wrapped float.

cdmistman: ah, you're right - my bad

cdmistman (approved):

dfellis:
One thing we realized through this is that `neg` and `abs` have a potential bug at the `INTMIN` value for their size because of Two's complement (-128 to 12*7*) so we need to Result-ify and Saturate those opcodes, too. I will do that in this PR.

cdmistman (approved):

PR #503, opened 2021-04-28
more consistent cli api naming proposal

depombo:
now top-level `app` cli commands and `credential`/`config` only differ in the nomenclature for editing them which is `upgrade` and `edit` respectively

### DIFF
--- a/anycloud/cli/src/main.rs
+++ b/anycloud/cli/src/main.rs
@@ -129,7 +129,7 @@ pub async fn main() {
       .about("Deploys your repository to a new App with a Deploy Config from anycloud.json")
       .arg_from_usage("-e, --env-file=[ENV_FILE] 'Specifies an optional environment file'")
     )
-    .subcommand(SubCommand::with_name("info")
+    .subcommand(SubCommand::with_name("list")
       .about("Displays all the Apps deployed with the Deploy Configs from anycloud.json")
     )
     .subcommand(SubCommand::with_name("terminate")
@@ -142,7 +142,7 @@ pub async fn main() {
     .subcommand(SubCommand::with_name("config")
       .about("Manage Deploy Configs used by Apps from the anycloud.json in the current directory")
       .setting(AppSettings::SubcommandRequiredElseHelp)
-      .subcommand(SubCommand::with_name("add")
+      .subcommand(SubCommand::with_name("new")
         .about("Add a new Deploy Config to the anycloud.json in the current directory and creates the file if it doesn't exist.")
       )
       .subcommand(SubCommand::with_name("list")
@@ -158,7 +158,7 @@ pub async fn main() {
     .subcommand(SubCommand::with_name("credentials")
       .about("Manage all Credentials used by Deploy Configs from the credentials file at ~/.anycloud/credentials.json")
       .setting(AppSettings::SubcommandRequiredElseHelp)
-      .subcommand(SubCommand::with_name("add")
+      .subcommand(SubCommand::with_name("new")
         .about("Add a new Credentials")
       )
       .subcommand(SubCommand::with_name("list")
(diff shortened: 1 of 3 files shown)


dfellis (approved):

cdmistman (approved):

aguillenv:
> now top-level `app` cli commands and `credential`/`config` only differ in the nomenclature for editing or removing them which is, `upgrade`/`terminate` and `edit`/`remove` respectively

Should we merge `terminate` and `remove` concepts eventually?

aguillenv (approved):

depombo:
> > now top-level `app` cli commands and `credential`/`config` only differ in the nomenclature for editing or removing them which is, `upgrade`/`terminate` and `edit`/`remove` respectively
> 
> Should we merge `terminate` and `remove` concepts eventually?

I thought of it a bit, but the verbs feel more adequate to describe the respective nouns. Terminating a config/credential definitely does not make sense, so it would have to be `remove`. `edit`ing an app sounds pretty weird though.

aguillenv:
> > > now top-level `app` cli commands and `credential`/`config` only differ in the nomenclature for editing or removing them which is, `upgrade`/`terminate` and `edit`/`remove` respectively
> > 
> > 
> > Should we merge `terminate` and `remove` concepts eventually?
> 
> I thought of it a bit, but the verbs feel more adequate to describe the respective nouns. Terminating a config/credential definitely does not make sense, so it would have to be `remove`. `edit`ing an app sounds pretty weird though.

No, for sure `edit` and `upgrade` mean different things. I'm referring to the concepts of `terminate` and `remove` (might be a different verb like `delete`). Might be easy in the future for the users to remember commands in my opinion. Remembering k8s, the `kubectl` command has almost the same verbs/actions (add, delete, edit) for different resources. But it's not a blocker, just something we can think about.

depombo:
> > > > now top-level `app` cli commands and `credential`/`config` only differ in the nomenclature for editing or removing them which is, `upgrade`/`terminate` and `edit`/`remove` respectively
> > > 
> > > 
> > > Should we merge `terminate` and `remove` concepts eventually?
> > 
> > 
> > I thought of it a bit, but the verbs feel more adequate to describe the respective nouns. Terminating a config/credential definitely does not make sense, so it would have to be `remove`. `edit`ing an app sounds pretty weird though.
> 
> No, for sure `edit` and `upgrade` mean different things. I'm referring to the concepts of `terminate` and `remove` (might be a different verb like `delete`). Might be easy in the future for the users to remember commands in my opinion. Remembering k8s, the `kubectl` command has almost the same verbs/actions (add, delete, edit) for different resources. But it's not a blocker, just something we can think about.

Ah I like `delete`

depombo:
> > > > > now top-level `app` cli commands and `credential`/`config` only differ in the nomenclature for editing or removing them which is, `upgrade`/`terminate` and `edit`/`remove` respectively
> > > > 
> > > > 
> > > > Should we merge `terminate` and `remove` concepts eventually?
> > > 
> > > 
> > > I thought of it a bit, but the verbs feel more adequate to describe the respective nouns. Terminating a config/credential definitely does not make sense, so it would have to be `remove`. `edit`ing an app sounds pretty weird though.
> > 
> > 
> > No, for sure `edit` and `upgrade` mean different things. I'm referring to the concepts of `terminate` and `remove` (might be a different verb like `delete`). Might be easy in the future for the users to remember commands in my opinion. Remembering k8s, the `kubectl` command has almost the same verbs/actions (add, delete, edit) for different resources. But it's not a blocker, just something we can think about.
> 
> Ah I like `delete`, let's go with that!

Talked offline with @dfellis and `delete`-ing a cluster still does not have the same implications as `terminate`-ing

PR #504, opened 2021-04-30
Run daemon locally

aguillenv:


anycloud/cli/src/common/mod.rs
```diff
@@ -0,0 +1,99 @@
+use std::env;
+use std::fs::read;
+use std::process::Command;
+
+pub fn get_base_agz_b64() -> String {
+  base64::encode(include_bytes!("../../alan/anycloud.agz"))
```
dfellis, line 6: I wonder if it's possible to make this a `const` and have the compiler do it at build time? I'm not sure what restrictions are there, but it seems like it would be more efficient? @cdmistman ?

avm/src/daemon/daemon.rs
```diff
@@ -178,7 +192,92 @@ async fn run_agz_b64(agz_b64: &str) -> DaemonResult<()> {
...
+  post_v1("localDaemonToken", body).await
+}
+
+async fn set_local_daemon_props() -> () {
+  let files_b64 = get_files_b64().await;
+  create_app_tar(files_b64.get(&"app.tar.gz".to_string()));
+  DAEMON_PROPS
+    .set(DaemonProperties {
+      clusterId: "daemon-local-cluster".to_string(),
+      agzB64: get_base_agz_b64(),
```
dfellis, line 239: What is this doing here? This shouldn't be here, because that's for `anycloud` not `alan daemon`.

dfellis: How do I run a local cluster that's an actual alan application?

avm/src/daemon/daemon.rs
```diff
@@ -178,7 +192,92 @@ async fn run_agz_b64(agz_b64: &str) -> DaemonResult<()> {
...
+    })
+    .unwrap();
+}
+
+fn create_certs_if_local() -> () {
+  if ALAN_TECH_ENV.as_str() == "local" {
+    // Self signed certs for local dev
+    // openssl req -newkey rsa:2048 -nodes -keyout key.pem -x509 -days 365 -out certificate.pem -subj "/C=US/ST=California/O=Alan Technologies, Inc/CN=*.anycloudapp.com"
+    let subj = if cfg!(target_os = "windows") {
+      "//C=US\\ST=California\\O=Alan Technologies, Inc\\CN=*.anycloudapp.com"
```
dfellis, line 252: Are you sure this is necessary? This isn't a filesystem path.

dfellis (changes requested):
I'm afraid I have to block this because it's hardwiring things for `anycloud` that it shouldn't be.

anycloud/cli/src/common/mod.rs
cdmistman: unfortunately, it'd require <https://docs.rs/base64/0.13.0/base64/fn.encode.html> to be a `const fn`... we could have `build.rs` encode the `agz` file and output it to another file so we can do
```rust
const ANYCLOUD_AGZ: &'static str = include_str!("../../alan/anycloud.agz.b64");
```
Do we absolutely need to b64-encode, anyways?

dfellis: Yeah, we do, because it's passed over as a key-val pair in a JSON blob.

cdmistman: ahhhhh right... yeah, the best we can do for now is to do the encoding with `build.rs`. I don't think there's anything blocking a `const fn` variant or a proc-macro of `base64::encode` though, so I wonder if we might be able to contribute it or at least open a feature request?

avm/src/daemon/daemon.rs
aguillenv: Hmm it's here because of https://github.com/alantech/alan/pull/504/files#diff-583f2380a4d780cb04bd93e3492a6f2828052a50e23460f3a2c8d89cf3ade79fR371 but now that you said right now on `main` we are not able to do it?

avm/src/daemon/daemon.rs
aguillenv: hmm will check again, I do not remember why I've done it but remember to have read something about it.

avm/src/daemon/daemon.rs
dfellis: If we're spinning up an `alan daemon` locally (something like `ALAN_TECH_ENV=local alan daemon` it should accept whatever `.agz` file you want, which can be `anycloud.agz` or `my_alan_app.agz`. Perhaps this mode should still spin up the `4142` control port, but have a different endpoint enabled that lets you just pass in the `.agz` file path as a raw string? Eg `curl -k https://localhost:4142/startLocal -d "./my_alan_app.agz"`

anycloud/cli/src/common/mod.rs
dfellis: It was a passing comment to make this slightly more efficient, but not really needed.

I think opening a feature request for `const fn` versions of `encode` (and maybe `decode`?) is probably a good idea, though?

avm/src/daemon/daemon.rs
aguillenv: You're right, I missed that case. What I'll do is giving the possibility to pass an optional argument to `daemon` representing the relative path to the `.agz` file. This way we avoid adding complexity to the process.

avm/src/daemon/daemon.rs
```diff
@@ -232,13 +233,13 @@ async fn generate_token() -> String {
...
 
-async fn set_local_daemon_props() -> () {
+async fn set_local_daemon_props(agz_b64: Option<String>) -> () {
   let files_b64 = get_files_b64().await;
   create_app_tar(files_b64.get(&"app.tar.gz".to_string()));
   DAEMON_PROPS
     .set(DaemonProperties {
       clusterId: "daemon-local-cluster".to_string(),
-      agzB64: get_base_agz_b64(),
+      agzB64: agz_b64.unwrap(),
```
cdmistman, line 242: if we unconditionally `agz_b64.unwrap()` then it should just be a `String`, not `Option<String>`

aguillenv: I'm checking it here: https://github.com/alantech/alan/pull/504/files#diff-583f2380a4d780cb04bd93e3492a6f2828052a50e23460f3a2c8d89cf3ade79fR277-R283

cdmistman: sure, but that's not guaranteed at every call to `set_local_daemon_props`, so somebody who's contributing and calls that fn might pass in a `None` and then everything fails

avm/src/daemon/daemon.rs
```diff
@@ -293,12 +294,12 @@ async fn get_daemon_props() -> Option<&'static DaemonProperties> {
   None
 }
 
-pub async fn start() {
+pub async fn start(agz_b64: Option<String>) {
```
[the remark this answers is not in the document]
aguillenv: Hmm here it is necessary, but will be handle properly before it's use where necessary.

avm/src/daemon/daemon.rs
```diff
@@ -180,7 +195,95 @@ async fn run_agz_b64(agz_b64: &str) -> DaemonResult<()> {
...
+
+async fn set_local_daemon_props(local_agz_b64: Option<String>) -> () {
+  let files_b64 = get_files_b64().await;
+  create_app_tar(files_b64.get(&"app.tar.gz".to_string()));
+  let agz_b64 = if let Some(local_agz_b64) = local_agz_b64 {
+    local_agz_b64
+  } else {
+    eprintln!("No agz found");
+    std::process::exit(1);
+  };
```
aguillenv, line 244: @cdmistman  Now the option is properly handled before its use here.

avm/src/daemon/daemon.rs
aguillenv: Oh thats true, got it

avm/src/daemon/daemon.rs
cdmistman: can you do a more descriptive error, like `running the Alan Daemon in a local environment requires an argument specifying the .agz file to run`

aguillenv: sure makes sense

cdmistman (approved):

avm/src/daemon/daemon.rs
```diff
@@ -1,17 +1,20 @@
...
 use anycloud::deploy;
 use anycloud::{error, CLUSTER_ID};
 use base64;
 use byteorder::{LittleEndian, ReadBytesExt};
 use flate2::read::GzDecoder;
+use lazy_static::lazy_static;
 use once_cell::sync::OnceCell;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
+#[allow(unused_imports)]
```
dfellis, line 17: Why is this line needed? You use `Command` below in the Linux `get_private_ip` function. Is the warning showing up on OSX and Windows? @cdmistman would `#[cfg(...` guarding this allow work?

cdmistman: good point - yes, adding the `#[cfg(...` here would work but it might be better to just directly import in the function?
```rust
#[cfg(target_os = "linux")]
async fn get_private_ip() -> DaemonResult<String> {
  use tokio::process::Command;
  ...
}
```

avm/src/daemon/daemon.rs
```diff
@@ -180,7 +195,98 @@ async fn run_agz_b64(agz_b64: &str) -> DaemonResult<()> {
...
+async fn generate_token() -> String {
+  let body = json!({
+    "deployConfig": deploy::get_config().await,
+  });
+  post_v1("localDaemonToken", body).await
+}
+
+async fn set_local_daemon_props(local_agz_b64: Option<String>) -> () {
+  let files_b64 = get_files_b64().await;
+  create_app_tar(files_b64.get(&"app.tar.gz".to_string()));
```
dfellis, line 238: This should be conditional. It's not part of a regular Alan daemon application, only Anycloud.

avm/src/daemon/daemon.rs
```diff
@@ -180,7 +195,98 @@ async fn run_agz_b64(agz_b64: &str) -> DaemonResult<()> {
...
+      .arg("365")
+      .arg("-out")
+      .arg("certificate.pem")
+      .arg("-subj")
+      .arg("/C=US/ST=California/O=Alan Technologies, Inc/CN=*.anycloudapp.com")
+      .spawn()
+      .expect("Error generating self signed certificate");
+    open_ssl.wait().expect("Failed to wait on child");
+  }
+}
```
dfellis, line 283: Not a blocker, but I think we may want to just use this in general, not just when local? Create the cert and key files if they don't exist, period, and then we simplify the cluster startup script.

avm/src/daemon/dns.rs
```diff
@@ -45,6 +45,17 @@ impl VMMetadata {
...
+  fn get_local_metadata() -> DaemonResult<Vec<VMMetadata>> {
+    Ok(vec![VMMetadata {
+      schema_version: "v1".to_string(),
+      alan_version: option_env!("CARGO_PKG_VERSION").unwrap().to_string(),
+      cloud: "LOCAL".to_string(),
+      private_ip_addr: "127.0.0.1".to_string(),
+      region: "localhost".to_string(),
+      public_ip_addr: "127.0.0.1".to_string(),
+    }])
+  }
```
dfellis, line 58: Why is this a function when it's used a single time with a hardwired result? Probably better to just inline it where it's called *or* make it a global const OnceCell so it isn't reconstructed on every call to `get_vms` below?

avm/src/daemon/daemon.rs
aguillenv: Indeed this would simplify cluster startup script but not sure if we should do it. Having the command on the startup script give us flexibility to change it easier and faster. What I will do here is create them if they do not exists and not based on `ALAN_TECH_ENV`.

avm/src/daemon/daemon.rs
aguillenv: True, will work on this

avm/src/daemon/daemon.rs
```diff
@@ -180,7 +195,98 @@ async fn run_agz_b64(agz_b64: &str) -> DaemonResult<()> {
...
+  }
+  Ok(())
+}
+
+async fn get_files_b64() -> HashMap<String, String> {
+  let mut files_b64 = HashMap::new();
+  files_b64.insert("Dockerfile".to_string(), get_dockerfile_b64().await);
+  files_b64.insert("app.tar.gz".to_string(), get_app_tar_gz_b64().await);
+  files_b64
+}
```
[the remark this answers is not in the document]
aguillenv: True. For now I will look if `anycloud.json` exists, this will let us know we are testing an AnyCloud app so we fetch this files.

avm/src/daemon/daemon.rs
aguillenv: Yes, I was having the warning on OSX. I'll use `tokio::process::Command(...` directly.

avm/src/daemon/dns.rs
aguillenv: Agree, should be static

avm/src/main.rs
```diff
@@ -107,6 +108,8 @@ fn main() {
     .subcommand(SubCommand::with_name("daemon")
       .about("Run an .agz file in daemon mode. Used on deploy within cloud provider VMs.")
       .arg_from_usage("<CLUSTER_SECRET> -s, --cluster-secret=<CLUSTER_SECRET> 'A secret string to constrain access to the control port'")
+      .arg_from_usage("-f, --agz-file=[AGZ_FILE] 'Specifies an optional agz file relative path for local usage'")
+      .arg_from_usage("[ANYCLOUD_APP] -a, --anycloud-app 'Specifies an optional AnyCloud app flag for local usage'")
```
dfellis, line 112: I will hold my nose on this flag for now, but looking forward to removing it in the future! ;)

dfellis (approved):

avm/src/main.rs
aguillenv: ¯\_(ツ)_/¯

PR #508, opened 2021-05-04
[LNN] function selection

cdmistman:
Enables function selection in the new LNN frontend compiler


bdd/spec/007_comparator_spec_lnn.sh
```diff
@@ -0,0 +1,68 @@
+Include build_tools.sh
+
+Describe "Comparators"
+  Describe "Equals"
+    before() {
+      lnn_sourceToAll "
+        from @std/app import start, stdout, exit
+
+        on start {
+          const i8val = 0;
```
dfellis, line 10: Does this mean it tries to default to the smallest matching type instead of the largest? Or should this be fixed to `const i8val: int8 = 0;`?

compiler/src/ammtoaga/index.ts
```diff
@@ -20,51 +20,42 @@ const CLOSURE_ARG_MEM_START = BigInt(Math.pow(-2,63))
...
+    let val: string = rec.get('assignables').t.trim()
+    const typename = rec.get('fulltypename').t.trim()
+    if (typename === 'string') {
       let str: string
       try {
         // Will fail on strings with escape chars
-        str = JSON.parse(rec.get('assignables').t.trim())
+        str = JSON.parse(val)
       } catch (e) {
-        // Hackery to get these strings to work
```
dfellis, line 48: You don't agree that the regex is hackery? ;)

compiler/src/lnntoamm/Expr.ts
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  static fromAssignablesAst(ast: LPNode, metadata: MetaData): [Stmt[], Expr] {
+    // break it up so that we're only working on one base assignable list or operator at a time.
+    let operated = ast.getAll().map(work => {
+      work = work.get('withoperators');
+      if (work.has('baseassignablelist')) {
+        return Expr.fromBaseassignablelist(work.get('baseassignablelist'), metadata);
+      } else if (work.has('operators')) {
+        return work;
```
dfellis, line 101: Should there be a TODO here, too? I don't see how this branch can be right?

compiler/src/lnntoamm/Fn.ts
```diff
@@ -126,76 +120,64 @@ export default class Fn {
...
         }
-        break;
-      } else {
-        throw new Error(`did not expect to inline stmt: ${stmt}`);
       }
     }
     if (!isReturned) {
-      if (this.retTy.breakdown() !== opcodes().get('void')) {
+      if (!this.retTy.compatibleWithConstraint(opcodes().get('void'))) {
         throw new Error(`no return value for function`);
```
dfellis, line 155: This error message seems misleading? Handlers can't return non-void, and the check appears to be correct, but reading that message I would think I need to return some value for the compiler to be happy.

dfellis: Maybe a message like `Event handlers should not return a value`?

compiler/src/lnntoamm/Fn.ts
```diff
@@ -126,76 +120,64 @@ export default class Fn {
...
-  // TODO: this will have to change in order to call fns multiple times - maybe deep cloning?
-  inline(amm: Output, args: Ref[], assign: string, kind: 'const' | 'let' | '') {
-    let argNames = Object.keys(this.args);
-    if (argNames.length !== args.length) {
-      // this should be caught by Call, it's just a sanity check
-      throw new Error(`number of arguments off`);
+  inline(amm: Output, args: Ref[], kind: AssignKind, name: string, ty: Builtin) {
+    let paramDefs = Object.values(this.params);
+    if (args.length !== paramDefs.length) {
+      throw new Error(`function call argument mismatch`);
```
dfellis, line 164: This shouldn't ever happen in reality, correct?

compiler/src/lnntoamm/Fn.ts
```diff
@@ -227,11 +209,15 @@ export class OpcodeFn extends Fn {
     __opcodes.put(name, [this]);
   }
 
-  inline(amm: Output, args: Ref[], assign: string, kind: 'const' | 'let' | '') {
+  asHandler(_amm: Output, _event: string) {
+    TODO('opcodes as event listener???');
```
dfellis, line 213: I guess any opcode that returns a void type could work, but the zero or one input arguments requirement for handlers means that likely isn't ever possible without a wrapper function?

compiler/src/lnntoamm/Stmt.ts
```diff
@@ -0,0 +1,351 @@
...
+  constructor(
+    ast: LPNode,
+    upstream: VarDef,
+    expr: Expr,
+  ) {
+    super(ast);
+    this.upstream = upstream;
+    this.expr = expr;
+    if (this.upstream.immutable) {
+      throw new Error(`cannot reassign to a const value`);
```
dfellis, line 86: Nit: maybe include the name of the variable in this error message?

Though that can be a future task to add better error messages everywhere, ideally including file names and line numbers.

compiler/src/lnntoamm/Stmt.ts
```diff
@@ -0,0 +1,351 @@
...
+    this.ty = ty;
+  }
+
+  ref(): Ref {
+    return new Ref(this);
+  }
+}
+
+export class Dec extends VarDef {
+  private __ammName: string = '';
```
dfellis, line 140: A reason for both the `private` and the underscore prefix on this field? I think the underscore prefix convention comes from Python as a way to signal "this is a private field, but you can mess with it if you absolutely have to."

compiler/src/lnntoamm/Types.ts
```diff
@@ -1,32 +1,41 @@
...
+    const generics = node.get('opttypegenerics');
     genericTys.push(parseFulltypename(generics.get('fulltypename')));
-    generics.get('cdr').getAll().map(n => n.get('fulltypename'));
+    genericTys.push(...generics.get('cdr').getAll().map(n => n.get('fulltypename')).map(parseFulltypename));
   }
   return [name, genericTys];
 };
 
-export default abstract class Type {
+// TODO: figure out type aliases (i think it actually makes sense to make a new type?)
```
dfellis, line 20: You can't alias a base type, then? Because then `type int = int64` will make a new type and that won't match `add(int64, int64): Result<int64>` (for instance).

bdd/spec/007_comparator_spec_lnn.sh
cdmistman: no - it's a test on the constraint from `emit exit i8val` at the bottom

compiler/src/ammtoaga/index.ts
cdmistman: ah, I didn't actually refactor in-place so I forgot to carry it over

compiler/src/lnntoamm/Expr.ts
cdmistman: yes, this should be a TODO

compiler/src/lnntoamm/Fn.ts
cdmistman: yeah, I just check many times to make absolutely sure. This error should never be thrown, since `Call` filters out functions that have an argument mismatch

compiler/src/lnntoamm/Fn.ts
cdmistman: yeah, which is why I feel weird adding support for them... it's not possible right now but if it ever does become possible we should figure out if it's a scenario we want to handle or not

compiler/src/lnntoamm/Stmt.ts
cdmistman: I do intend for there to be a future task focused on improving error messages (eg, actually using the `ast` kept by all of the nodes), but this is simple enough

compiler/src/lnntoamm/Types.ts
cdmistman: Well, if there was a `class Alias extends Type` then `compatibleWithConstraint` would allow it to be compatible with the aliased type

bdd/spec/005_boolean_spec_lnn.sh
```diff
@@ -0,0 +1,48 @@
...
+
+  OUTPUT="true
+false
+true
+false
+true
+true
+false
+true
+false"
```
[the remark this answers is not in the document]
cdmistman: you're right... i wonder why the bdd test was passing??????

bdd/spec/007_comparator_spec_lnn.sh
dfellis: Ahh, got it. Can we add a comment saying that?

compiler/src/ammtoaga/index.ts
```diff
@@ -20,51 +20,43 @@ const CLOSURE_ARG_MEM_START = BigInt(Math.pow(-2,63))
...
-      break
-    case "float64":
-      val = rec.get('assignables').t.trim() + 'f64'
-      globalMem[`@${currentOffset}`] = val
-      addressMap[rec.get('decname').t] = currentOffset
-      currentOffset -= 8
-      break
-    case "string":
+    let offset: number = 8;
+    let val: string = rec.get('assignables').t.trim()
```
dfellis, line 38: Just curious: Typescript wasn't figuring these types out for you, or did you want to be more explicit here?

compiler/src/lnntoamm/Expr.ts
```diff
@@ -0,0 +1,316 @@
...
+    } else {
+      throw new Error(`unrecognized constants node: ${ast}`);
+    }
+    return [[], new Const(ast, val, detectedTy)];
+  }
+
+  inline(amm: Output, kind: AssignKind, name: string, ty: Builtin) {
+    const globalName = amm.global('const', ty, this.val);
+    let copyOp = 'copy';
+    switch (ty.ammName) {
```
dfellis, line 262: Instead of a switch statement why not use something like: https://github.com/alantech/alan/pull/508/files#diff-6f21a0275e18e6835d1c7f6e90fba00333ab84951fe8205eeb29bed63861cf6eR23

```js
const getSuffix = (typename) => typename[0] + typename.replace(/[a-z]*/, '')
copyOp += getSuffix(ty.ammName) || ''
```

compiler/src/lnntoamm/Fn.ts
```diff
@@ -112,90 +102,90 @@ export default class Fn {
...
-      } else {
-        throw new Error(`did not expect to inline stmt: ${stmt}`);
+      if (stmt instanceof Exit) {
+        if (ii !== this.body.length - 1) {
+          throw new Error(`got a return at a bad time (should've been caught already?)`);
+        }
+        let refCall = 'reff';
+        if (ty.eq(opcodes().get('string'))) {
+          refCall = 'refv';
+        }
```
dfellis, line 182: Maybe a `TODO` that this will eventually impact all other types `Array`, `HashMap`, `Result`, etc, and that this conditional should be inverted to something like:

```js
let refCall = 'refv';
if (['int8', 'int16', 'int32', 'int64', 'float32', 'float64', 'bool'].includes(ty.name)) {
  refCall = 'reff';
}
```

compiler/src/lnntoamm/Stmt.ts
```diff
@@ -0,0 +1,351 @@
...
+  ) {
+    this.ast = ast;
+  }
+
+  abstract inline(amm: Output): void;
+
+  static fromAst(ast: LPNode, metadata: MetaData): Stmt[] {
+    let stmts = [];
+    if (ast.has('assignables')) {
+      let [ss, expr] = Expr.fromAssignablesAst(ast.get('assignments'), metadata);
```
dfellis, line 54: Nit: maybe don't use `ss` as a variable name anywhere? It's never the Reich name for one.

compiler/src/lnntoamm/Expr.ts
```diff
@@ -0,0 +1,316 @@
...
+        break;
+      case 'float64':
+        copyOp += 'f64';
+        break;
+      case 'string':
+        copyOp += 'str';
+        break;
+      case 'bool':
+        copyOp += 'bool';
+        break;
```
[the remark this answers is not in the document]
cdmistman: this is for the name of the opcode, ie `copyi8` or `copybool`

compiler/src/lnntoamm/Fn.ts
cdmistman: i'll just go ahead and fix it now :)

compiler/src/lnntoamm/Expr.ts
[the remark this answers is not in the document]
dfellis: Oh, then my suggestion above won't work without some special-casing for `bool` and `string`

bdd/spec/007_comparator_lnn_spec.sh
```diff
@@ -83,11 +83,13 @@ true
 false"
 
     It "runs js"
+      Pending types-depending-on-each-other
```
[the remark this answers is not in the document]
dfellis: It would still be valuable to make sure the inference that currently exists doesn't regress.

bdd/spec/001_event_lnn_spec.sh
```diff
@@ -93,6 +93,7 @@ Describe "Events"
 
       on aNumber fn(num: int64) {
         print('I got a number! ' + num.toString());
+        wait(1000);
```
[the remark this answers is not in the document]
cdmistman: that would require interface types to work - `print` isn't currently in `lnn` (this test is `Pending`) - will update the pending message

dfellis (approved):

PR #512, opened 2021-05-05
Distributed Datastore

dfellis:
Still a WIP, but all of the necessary pieces have now been built. Planned perf improvements in a followup PR.

### DIFF
--- a/avm/src/daemon/ctrl.rs
+++ b/avm/src/daemon/ctrl.rs
@@ -120,9 +120,10 @@ impl LogRendezvousHash {
   // Runs a binary search for the record whose hash is closest to the key hash without
   // going over. If none are found, the *last* record in the list is returned as it wraps around.
   fn get_idx_for_key(&self, key: &str) -> usize {
+    let l = self.sorted_hashes.len();
     let mut key_hasher = XxHash64::with_seed(0xa1a2);
     key_hasher.write(key.as_bytes());
-    let key_hash = key_hasher.finish();
+    let key_hash = key_hasher.finish() % (l as u64);
     eprintln!("key: {} key_hash: {}", key, key_hash);
     let idx = match self
       .sorted_hashes
@@ -130,7 +131,7 @@ impl LogRendezvousHash {
     {
       Ok(res) => res,
       // All were too large, implies last (which wraps around) owns it
-      Err(_) => self.sorted_hashes.len() - 1,
+      Err(_) => l - 1,
     };
     idx
   }
--- a/avm/src/daemon/ctrl.rs
+++ b/avm/src/daemon/ctrl.rs
@@ -248,8 +248,7 @@ async fn dsgetf_inner(req: Request<Body>) -> DaemonResult<Arc<HandlerMemory>> {
   // TODO: For now assume this was directed at the right node, later on add auto-forwarding logic
   let bytes = body::to_bytes(req.into_body()).await?;
   let body: DSGet = serde_json::from_slice(&bytes).unwrap();
-  let ds = Arc::clone(&DS);
-  let maybe_hm = ds.get(&body.nskey);
+  let maybe_hm = DS.get(&body.nskey);
   let mut hand_mem = HandlerMemory::new(None, 1)?;
   hand_mem.init_fractal(0)?;
   hand_mem.push_fixed(0, if maybe_hm.is_some() { 1i64 } else { 0i64 })?;
@@ -282,8 +281,7 @@ async fn dsgetv_inner(req: Request<Body>) -> DaemonResult<Arc<HandlerMemory>> {
   // TODO: For now assume this was directed at the right node, later on add auto-forwarding logic
   let bytes = body::to_bytes(req.into_body()).await?;
   let body: DSGet = serde_json::from_slice(&bytes).unwrap();
-  let ds = Arc::clone(&DS);
-  let maybe_hm = ds.get(&body.nskey);
+  let maybe_hm = DS.get(&body.nskey);
   let mut hand_mem = HandlerMemory::new(None, 1)?;
   hand_mem.init_fractal(0)?;
   hand_mem.push_fixed(0, if maybe_hm.is_some() { 1i64 } else { 0i64 })?;
@@ -322,8 +320,7 @@ async fn dshas_inner(req: Request<Body>) -> DaemonResult<bool> {
   // TODO: For now assume this was directed at the right node, later on add auto-forwarding logic
   let bytes = body::to_bytes(req.into_body()).await?;
   let body: DSGet = serde_json::from_slice(&bytes).unwrap();
-  let ds = Arc::clone(&DS);
-  Ok(ds.contains_key(&body.nskey))
+  Ok(DS.contains_key(&body.nskey))
 }
 
 async fn handle_dsdel(req: Request<Body>) -> Result<Response<Body>, Infallible> {
@@ -346,8 +343,7 @@ async fn dsdel_inner(req: Request<Body>) -> DaemonResult<bool> {
   // TODO: For now assume this was directed at the right node, later on add auto-forwarding logic
   let bytes = body::to_bytes(req.into_body()).await?;
   let body: DSGet = serde_json::from_slice(&bytes).unwrap();
-  let ds = Arc::clone(&DS);
-  Ok(ds.remove(&body.nskey).is_some())
+  Ok(DS.remove(&body.nskey).is_some())
 }
 
 async fn handle_dssetf(req: Request<Body>) -> Result<Response<Body>, Infallible> {
@@ -370,8 +366,7 @@ async fn dssetf_inner(req: Request<Body>) -> DaemonResult<()> {
   let val = hand_mem.read_fixed(1)?;
   let mut hm = HandlerMemory::new(None, 1)?;
   hm.write_fixed(0, val)?;
-  let ds = Arc::clone(&DS);
-  ds.insert(nskey, hm);
+  DS.insert(nskey, hm);
   Ok(())
 }
 
@@ -394,8 +389,7 @@ async fn dssetv_inner(req: Request<Body>) -> DaemonResult<()> {
   let nskey = HandlerMemory::fractal_to_string(hand_mem.read_fractal(0)?)?;
   let mut hm = HandlerMemory::new(None, 1)?;
   HandlerMemory::transfer(&hand_mem, 1, &mut hm, 0)?;
-  let ds = Arc::clone(&DS);
-  ds.insert(nskey, hm);
+  DS.insert(nskey, hm);
   Ok(())
 }
 
--- a/avm/src/daemon/ctrl.rs
+++ b/avm/src/daemon/ctrl.rs
@@ -100,20 +100,17 @@ impl LogRendezvousHash {
   }
 
   pub fn get_primary_node_id(&self, key: &str) -> &str {
-    eprintln!("LRH: {:?}", self);
     self.get_assigned_nodes_id(key)[0]
   }
 
   pub fn get_assigned_nodes_id(&self, key: &str) -> Vec<&str> {
     let top = cmp::min(3, self.sorted_hashes.len());
     let mut idx = self.get_idx_for_key(key);
-    eprintln!("Main index: {}", idx);
     let mut ids: Vec<&str> = Vec::new();
     for _ in 0..top {
       ids.push(&self.sorted_hashes[idx].id);
       idx = (idx + 1) % self.sorted_hashes.len();
     }
-    eprintln!("Selected ids: {:?}", ids);
     ids
   }
 
@@ -124,7 +121,6 @@ impl LogRendezvousHash {
     let mut key_hasher = XxHash64::with_seed(0xa1a2);
     key_hasher.write(key.as_bytes());
     let key_hash = key_hasher.finish() % (l as u64);
-    eprintln!("key: {} key_hash: {}", key, key_hash);
     let idx = match self
       .sorted_hashes
       .binary_search_by(|a| a.hash.cmp(&key_hash))
--- a/avm/src/daemon/ctrl.rs
+++ b/avm/src/daemon/ctrl.rs
@@ -127,8 +127,6 @@ impl LogRendezvousHash {
       Ok(res) => res,
       Err(res) => res, // This is actually the last index less than the hash, which is what we want
     } % self.sorted_hashes.len();
-    eprintln!("key_hash {} idx {}", key_hash, idx);
-    eprintln!("sorted_hashes {:?}", self.sorted_hashes);
     idx
   }
 }
--- a/avm/src/daemon/ctrl.rs
+++ b/avm/src/daemon/ctrl.rs
@@ -32,7 +32,7 @@ use crate::vm::opcode::REGION_VMS;
 
 pub static NAIVE_CLIENT: OnceCell<Client<HttpsConnector<HttpConnector>>> = OnceCell::new();
 
-#[derive(Debug)]
+#[derive(Clone,Debug)]
 pub struct HashedId {
   pub id: String,
   pub hash: u64,
@@ -54,7 +54,7 @@ impl HashedId {
 // An algorithm based on the classic Rendezvous Hash but with changes to make the performance
 // closer to modern Consistent Hash algorithms while retaining Rendezvous Hash's coordination-free
 // routing.
-#[derive(Debug)]
+#[derive(Clone,Debug)]
 pub struct LogRendezvousHash {
   sorted_hashes: Vec<HashedId>,
 }
@@ -127,6 +127,7 @@ impl LogRendezvousHash {
   }
 }
 
+#[derive(Clone,Debug)]
 pub struct ControlPort {
   lrh: LogRendezvousHash,
   client: Client<HttpsConnector<HttpConnector>>,
--- a/avm/src/daemon/ctrl.rs
+++ b/avm/src/daemon/ctrl.rs
@@ -23,12 +23,15 @@ use once_cell::sync::OnceCell;
 use openssl::ssl::{SslConnector, SslMethod, SslVerifyMode};
 //use rustls::ClientConfig;
 use twox_hash::XxHash64;
+use serde::{Deserialize, Serialize};
+use protobuf::Message;
 
 use crate::daemon::daemon::{DaemonProperties, DaemonResult, CLUSTER_SECRET, DAEMON_PROPS};
 use crate::daemon::dns::VMMetadata;
 use crate::make_server;
 use crate::vm::http::{HttpType, HttpsConfig};
-use crate::vm::opcode::REGION_VMS;
+use crate::vm::memory::{CLOSURE_ARG_MEM_START, HandlerMemory};
+use crate::vm::opcode::{DS, REGION_VMS};
 
 pub static NAIVE_CLIENT: OnceCell<Client<HttpsConnector<HttpConnector>>> = OnceCell::new();
 
@@ -151,6 +154,12 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible> 
     "/health" => Ok(Response::builder().status(200).body("ok".into()).unwrap()),
     "/clusterHealth" => handle_cluster_health(),
     "/start" => handle_start(req).await,
+    "/datastore/getf" => handle_dsgetf(req).await, // TODO: How to better organize the datastore stuff?
+    "/datastore/getv" => handle_dsgetv(req).await,
+    "/datastore/has" => handle_dshas(req).await,
+    "/datastore/del" => handle_dsdel(req).await,
+    "/datastore/setf" => handle_dssetf(req).await,
+    "/datastore/setv" => handle_dssetv(req).await,
     _ => Ok(Response::builder().status(404).body("fail".into()).unwrap()),
   }
 }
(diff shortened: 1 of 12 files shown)


avm/src/daemon/ctrl.rs
```diff
@@ -112,21 +117,23 @@ impl LogRendezvousHash {
   // Runs a binary search for the record whose hash is closest to the key hash without
   // going over. If none are found, the *last* record in the list is returned as it wraps around.
   fn get_idx_for_key(&self, key: &str) -> usize {
+    let l = self.sorted_hashes.len();
     let mut key_hasher = XxHash64::with_seed(0xa1a2);
     key_hasher.write(key.as_bytes());
-    let key_hash = key_hasher.finish();
+    let key_hash = key_hasher.finish() % (l as u64);
```
aguillenv, line 123: why does need to be done?

avm/src/daemon/ctrl.rs
```diff
@@ -213,6 +226,181 @@ fn write_b64_file(pwd: &PathBuf, file_name: &str, content: &str) -> io::Result<(
...
+    Err(err) => {
+      // TODO: What error message here? Also should this also be a valid HM out of here?
+      eprintln!("{:?}", err);
+      Ok(Response::builder().status(500).body("fail".into()).unwrap())
+    }
+  }
+}
+
+async fn dsgetf_inner(req: Request<Body>) -> DaemonResult<Arc<HandlerMemory>> {
+  // For now assume this was directed at the right node, later on add some auto-forwarding logic
```
aguillenv, line 250: Should this comment be a `TODO`?

avm/src/daemon/ctrl.rs
```diff
@@ -213,6 +226,181 @@ fn write_b64_file(pwd: &PathBuf, file_name: &str, content: &str) -> io::Result<(
...
+    Err(err) => {
+      // TODO: What error message here? Also should this also be a valid HM out of here?
+      eprintln!("{:?}", err);
+      Ok(Response::builder().status(500).body("fail".into()).unwrap())
+    }
+  }
+}
+
+async fn dsgetv_inner(req: Request<Body>) -> DaemonResult<Arc<HandlerMemory>> {
+  // For now assume this was directed at the right node, later on add some auto-forwarding logic
```
aguillenv, line 284: Same as above

avm/src/daemon/ctrl.rs
```diff
@@ -150,6 +157,12 @@ async fn control_port(req: Request<Body>) -> Result<Response<Body>, Infallible>
     "/health" => Ok(Response::builder().status(200).body("ok".into()).unwrap()),
     "/clusterHealth" => handle_cluster_health(),
     "/start" => handle_start(req).await,
+    "/datastore/getf" => handle_dsgetf(req).await, // TODO: How to better organize the datastore stuff?
+    "/datastore/getv" => handle_dsgetv(req).await,
```
aguillenv, line 161: What does the `f` and the `v` stand for?

aguillenv (approved):

avm/src/daemon/ctrl.rs
dfellis: It was my fast workaround for a bug below; it's not selecting the index based on the `key_hash`, it was accidentally treating the `key_hash` as the index, so I just made sure it was within the range of the indexes.

Basically, the range of indexes is just `0` and `1` in my test, while on average a hash of type `u64` would have the value of `2^63` and the odds of actually ever selecting `0` in the code were `1 : 2^64` which is way below the threshold of likelihood physicists use for quantum mechanics experiments.

I will undo this after I figure out how to make the rest of the function do what it should be doing.

avm/src/daemon/ctrl.rs
dfellis: Yep

avm/src/daemon/ctrl.rs
dfellis: `f`ixed and `v`ariable. Basically is the value (stored as) a single 64-bit integer, or is it an array of some arbitrary size and shape?

avm/src/daemon/ctrl.rs
dfellis: I think this is why the code was not working: https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by

> If the value is found then Result::Ok is returned, containing the index of the matching element. If there are multiple matches, then any one of the matches could be returned. If the value is not found then Result::Err is returned, containing the index where a matching element could be inserted while maintaining sorted order.

We're almost never going to find the value. The `Err` is not actually an error here, but a weird flag indicating which index we could insert this value into the vector (which indicates which node we want it to go to). I *think* I have an idea on how to resolve this, going to test.

avm/src/vm/opcode.rs
```diff
@@ -3658,8 +3672,21 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+        Some(ctrl_port) => Some(ctrl_port.borrow().clone()), // TODO: Use thread-local storage
+        None => None,
+      };
+      let is_key_owner = match ctrl_port {
+        Some(ref ctrl_port) => ctrl_port.is_key_owner(&nskey),
+        None => true,
+      };
+      if is_key_owner {
+        let ds = Arc::clone(&DS);
+        ds.insert(nskey, hm);
```
cdmistman, line 3686: is there a reason for this instead of `DS.insert(nskey, hm);`?

avm/src/vm/opcode.rs
```diff
@@ -3658,8 +3672,21 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
       let ns = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[0])?)?;
       let key = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[1])?)?;
       let nskey = format!("{}:{}", ns, key);
-      let ds = Arc::clone(&DS);
-      ds.insert(nskey, hm);
+      let ctrl_port = CONTROL_PORT_CHANNEL.get();
+      let ctrl_port = match ctrl_port {
+        Some(ctrl_port) => Some(ctrl_port.borrow().clone()), // TODO: Use thread-local storage
```
cdmistman, line 3677: i can see this cloning being very expensive... is it necessary or can we just `borrow()` it? If we can get rid of it, this whole reassignment can be:
```rust
let ctrl_port = ctrl_port.map(watch::Receiver::borrow);
```

avm/src/vm/opcode.rs
```diff
@@ -3658,8 +3672,21 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      };
+      let is_key_owner = match ctrl_port {
+        Some(ref ctrl_port) => ctrl_port.is_key_owner(&nskey),
+        None => true,
+      };
+      if is_key_owner {
+        let ds = Arc::clone(&DS);
+        ds.insert(nskey, hm);
+      } else {
+        ctrl_port.unwrap().dssetv(&nskey, &hm).await;
```
cdmistman, line 3688: shouldn't we handle the `None` case?

avm/src/vm/opcode.rs
```diff
@@ -3668,8 +3695,21 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
       let ns = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[0])?)?;
       let key = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[1])?)?;
       let nskey = format!("{}:{}", ns, key);
-      let ds = Arc::clone(&DS);
-      let has = ds.contains_key(&nskey);
+      let ctrl_port = CONTROL_PORT_CHANNEL.get();
+      let ctrl_port = match ctrl_port {
+        Some(ctrl_port) => Some(ctrl_port.borrow().clone()), // TODO: Use thread-local storage
+        None => None,
+      };
```
cdmistman, line 3702: same as above - this can be cheaper (and reduced to 1 line)

avm/src/vm/opcode.rs
```diff
@@ -3668,8 +3695,21 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      let is_key_owner = match ctrl_port {
+        Some(ref ctrl_port) => ctrl_port.is_key_owner(&nskey),
+        None => true,
+      };
+      let has = if is_key_owner {
+        let ds = Arc::clone(&DS);
```
cdmistman, line 3708: same as above - we don't need this atomic op

avm/src/vm/opcode.rs
```diff
@@ -3668,8 +3695,21 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+        ds.contains_key(&nskey)
+      } else {
+        ctrl_port.unwrap().dshas(&nskey).await
```
cdmistman, line 3711: same as above - we should probably handle the `None` case

avm/src/vm/opcode.rs
```diff
@@ -3679,8 +3719,23 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
       let key = HandlerMemory::fractal_to_string(hand_mem.read_fractal(args[1])?)?;
       let nskey = format!("{}:{}", ns, key);
+      // If it exists locally, remove it here, too
       let ds = Arc::clone(&DS);
       let removed = ds.remove(&nskey).is_some();
+      let ctrl_port = CONTROL_PORT_CHANNEL.get();
+      let ctrl_port = match ctrl_port {
+        Some(ctrl_port) => Some(ctrl_port.borrow().clone()), // TODO: Use thread-local storage
+        None => None,
+      };
```
cdmistman, line 3729: same as above - can be cheaper + reduced to 1 line

avm/src/vm/opcode.rs
```diff
@@ -3679,8 +3719,23 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      let is_key_owner = match ctrl_port {
+        Some(ref ctrl_port) => ctrl_port.is_key_owner(&nskey),
+        None => true,
+      };
+      let removed = if is_key_owner {
+        removed
+      } else {
+        ctrl_port.unwrap().dsdel(&nskey).await
```
cdmistman, line 3737: same as above - `None` will cause a panic

avm/src/vm/opcode.rs
```diff
@@ -3691,13 +3746,41 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-      hand_mem.push_fixed(args[2], if maybe_hm.is_some() { 1i64 } else { 0i64 })?;
-      match maybe_hm {
-        Some(hm) => hand_mem.push_fixed(args[2], hm.read_fixed(0)?),
-        None => hand_mem.push_fractal(args[2], HandlerMemory::str_to_fractal("namespace-key pair not found")),
-      }?;
+      let ctrl_port = CONTROL_PORT_CHANNEL.get();
+      let ctrl_port = match ctrl_port {
+        Some(ctrl_port) => Some(ctrl_port.borrow().clone()), // TODO: Use thread-local storage
+        None => None,
+      };
```
cdmistman, line 3753: same expensive op

avm/src/vm/opcode.rs
```diff
@@ -3691,13 +3746,41 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+            hand_mem.push_fixed(args[2], 1i64)?;
+            hand_mem.push_fixed(args[2], hm.read_fixed(0)?)?;
+          },
+          None => {
+            hand_mem.push_fixed(args[2], 0i64)?;
+            hand_mem.push_fractal(args[2], HandlerMemory::str_to_fractal("namespace-key pair not found"))?;
+          },
+        }
+      } else {
+        let maybe_hm = ctrl_port.unwrap().dsgetf(&nskey).await;
```
cdmistman, line 3772: `None`

avm/src/vm/opcode.rs
```diff
@@ -3707,18 +3790,41 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
-        },
-        None => {
-          hand_mem.push_fixed(args[2], 0i64)?;
-          hand_mem.push_fractal(args[2], HandlerMemory::str_to_fractal("namespace-key pair not found"))?;
-        },
+      let ctrl_port = CONTROL_PORT_CHANNEL.get();
+      let ctrl_port = match ctrl_port {
+        Some(ctrl_port) => Some(ctrl_port.borrow().clone()), // TODO: Use thread-local storage
+        None => None,
+      };
```
cdmistman, line 3797: same expensive op

avm/src/vm/opcode.rs
```diff
@@ -3707,18 +3790,41 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+            HandlerMemory::transfer(&hm, 0, &mut hand_mem, CLOSURE_ARG_MEM_START)?;
+            hand_mem.push_register(args[2], CLOSURE_ARG_MEM_START)?;
+          },
+          None => {
+            hand_mem.push_fixed(args[2], 0i64)?;
+            hand_mem.push_fractal(args[2], HandlerMemory::str_to_fractal("namespace-key pair not found"))?;
+          },
+        }
+      } else {
+        let maybe_hm = ctrl_port.unwrap().dsgetv(&nskey).await;
```
cdmistman, line 3817: `unwrap`

avm/src/daemon/ctrl.rs
```diff
@@ -213,6 +224,181 @@ fn write_b64_file(pwd: &PathBuf, file_name: &str, content: &str) -> io::Result<(
...
+async fn dssetf_inner(req: Request<Body>) -> DaemonResult<()> {
+  // TODO: For now assume this was directed at the right node, later on add auto-forwarding logic
+  let bytes = body::to_bytes(req.into_body()).await?;
+  let pb = protos::HandlerMemory::HandlerMemory::parse_from_bytes(&bytes)?;
+  let hand_mem = HandlerMemory::from_pb(&pb)?;
+  let nskey = HandlerMemory::fractal_to_string(hand_mem.read_fractal(0)?)?;
+  let val = hand_mem.read_fixed(1)?;
+  let mut hm = HandlerMemory::new(None, 1)?;
+  hm.write_fixed(0, val)?;
+  let ds = Arc::clone(&DS);
```
cdmistman, line 373: we can act directly on `DS`

avm/src/daemon/ctrl.rs
```diff
@@ -213,6 +224,181 @@ fn write_b64_file(pwd: &PathBuf, file_name: &str, content: &str) -> io::Result<(
...
+
+async fn dssetv_inner(req: Request<Body>) -> DaemonResult<()> {
+  // TODO: For now assume this was directed at the right node, later on add auto-forwarding logic
+  let bytes = body::to_bytes(req.into_body()).await?;
+  let pb = protos::HandlerMemory::HandlerMemory::parse_from_bytes(&bytes)?;
+  let hand_mem = HandlerMemory::from_pb(&pb)?;
+  let nskey = HandlerMemory::fractal_to_string(hand_mem.read_fractal(0)?)?;
+  let mut hm = HandlerMemory::new(None, 1)?;
+  HandlerMemory::transfer(&hand_mem, 1, &mut hm, 0)?;
+  let ds = Arc::clone(&DS);
```
cdmistman, line 397: we can act directly on `DS`

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
       }
     }
   }
+
+  pub fn get_vm_for_key(self: &ControlPort, key: &str) -> &VMMetadata {
+    &self.vms[self.lrh.get_primary_node_id(key)]
+  }
```
cdmistman, line 592: should we have this return `Option<&VMMetadata>` and do `self.vms.get(self.lrh.get_primary_node_id(key)`?

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+      Some(my) => self.get_vm_for_key(key).private_ip_addr == my.private_ip_addr,
+      None => false,
+    }
+  }
+
+  pub async fn dsgetf(self: &ControlPort, key: &str) -> Option<Arc<HandlerMemory>> {
+    match self.dsgetf_inner(key).await {
+      Ok(hm) => Some(hm),
+      Err(_) => None,
+    }
```
cdmistman, line 605: nit: can be
```suggestion
    self.dsgetf_inner(key).await.ok()
```

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+  }
+
+  async fn dsgetf_inner(self: &ControlPort, key: &str) -> DaemonResult<Arc<HandlerMemory>> {
+    let vm = self.get_vm_for_key(key);
+    let url = format!("https://{}:4142/datastore/getf", vm.public_ip_addr);
+    let req = Request::builder().method("POST").uri(url);
+    let cluster_secret = CLUSTER_SECRET.get().unwrap().as_ref().unwrap().clone();
```
cdmistman, line 612: 2 nits:
1. should we handle the error case?
2. could be `.get().unwrap().clone().unwrap();`

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+    let bytes = hyper::body::to_bytes(res.body_mut()).await?;
+    let pb = protos::HandlerMemory::HandlerMemory::parse_from_bytes(&bytes)?;
+    Ok(HandlerMemory::from_pb(&pb)?)
+  }
+
+  pub async fn dsgetv(self: &ControlPort, key: &str) -> Option<Arc<HandlerMemory>> {
+    match self.dsgetv_inner(key).await {
+      Ok(hm) => Some(hm),
+      Err(_) => None,
+    }
```
cdmistman, line 630: nit:
```suggestion
    self.dsgetv_inner(key).await.ok()
```

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+  }
+
+  async fn dsgetv_inner(self: &ControlPort, key: &str) -> DaemonResult<Arc<HandlerMemory>> {
+    let vm = self.get_vm_for_key(key);
+    let url = format!("https://{}:4142/datastore/getv", vm.public_ip_addr);
+    let req = Request::builder().method("POST").uri(url);
+    let cluster_secret = CLUSTER_SECRET.get().unwrap().as_ref().unwrap().clone();
```
cdmistman, line 637: same as above

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+    let bytes = hyper::body::to_bytes(res.body_mut()).await?;
+    let pb = protos::HandlerMemory::HandlerMemory::parse_from_bytes(&bytes)?;
+    Ok(HandlerMemory::from_pb(&pb)?)
+  }
+
+  pub async fn dshas(self: &ControlPort, key: &str) -> bool {
+    match self.dshas_inner(key).await {
+      Ok(has) => has,
+      Err(_) => false,
+    }
```
cdmistman, line 655: nit:
```suggestion
    self.dshas_inner(key).await.unwrap_or(false)
```

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+  }
+
+  async fn dshas_inner(self: &ControlPort, key: &str) -> DaemonResult<bool> {
+    let vm = self.get_vm_for_key(key);
+    let url = format!("https://{}:4142/datastore/has", vm.public_ip_addr);
+    let req = Request::builder().method("POST").uri(url);
+    let cluster_secret = CLUSTER_SECRET.get().unwrap().as_ref().unwrap().clone();
```
cdmistman, line 662: nit: same as above

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+      Ok(has) => has,
+      Err(_) => false,
+    }
+  }
+
+  async fn dsdel_inner(self: &ControlPort, key: &str) -> DaemonResult<bool> {
+    let vm = self.get_vm_for_key(key);
+    let url = format!("https://{}:4142/datastore/del", vm.public_ip_addr);
+    let req = Request::builder().method("POST").uri(url);
+    let cluster_secret = CLUSTER_SECRET.get().unwrap().as_ref().unwrap().clone();
```
cdmistman, line 686: nit: same as above

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+    let mut res = self.client.request(req_obj).await?;
+    let bytes = hyper::body::to_bytes(res.body_mut()).await?;
+    Ok(std::str::from_utf8(&bytes)? == "true")
+  }
+
+  pub async fn dssetf(self: &ControlPort, key: &str, val: &Arc<HandlerMemory>) -> bool {
+    match self.dssetf_inner(key, val).await {
+      Ok(set) => set,
+      Err(_) => false,
+    }
```
cdmistman, line 703: nit: `unwrap_or`

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+  ) -> DaemonResult<bool> {
+    let vm = self.get_vm_for_key(key);
+    let url = format!("https://{}:4142/datastore/setf", vm.public_ip_addr);
+    let mut hm = HandlerMemory::new(None, 1)?;
+    hm.write_fractal(0, &HandlerMemory::str_to_fractal(key))?;
+    HandlerMemory::transfer(val, 0, &mut hm, 1)?;
+    let mut out = vec![];
+    hm.to_pb().write_to_vec(&mut out).unwrap();
+    let req = Request::builder().method("POST").uri(url);
+    let cluster_secret = CLUSTER_SECRET.get().unwrap().as_ref().unwrap().clone();
```
cdmistman, line 719: same as above

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+    let mut res = self.client.request(req_obj).await?;
+    let bytes = hyper::body::to_bytes(res.body_mut()).await?;
+    Ok(std::str::from_utf8(&bytes)? == "true")
+  }
+
+  pub async fn dssetv(self: &ControlPort, key: &str, val: &Arc<HandlerMemory>) -> bool {
+    match self.dssetv_inner(key, val).await {
+      Ok(set) => set,
+      Err(_) => false,
+    }
```
cdmistman, line 731: nit: `unwrap_or`

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+  ) -> DaemonResult<bool> {
+    let vm = self.get_vm_for_key(key);
+    let url = format!("https://{}:4142/datastore/setv", vm.public_ip_addr);
+    let mut hm = HandlerMemory::new(None, 1)?;
+    hm.write_fractal(0, &HandlerMemory::str_to_fractal(key))?;
+    HandlerMemory::transfer(val, 0, &mut hm, 1)?;
+    let mut out = vec![];
+    hm.to_pb().write_to_vec(&mut out).unwrap();
+    let req = Request::builder().method("POST").uri(url);
+    let cluster_secret = CLUSTER_SECRET.get().unwrap().as_ref().unwrap().clone();
```
cdmistman, line 747: same as above

avm/src/vm/opcode.rs
dfellis: I couldn't get it to compile with just `borrow`ing it. But as the `TODO` mentions I want to follow this up with a thread-local storage for the control port that is only updated when there's actually a change to it, so that cost all goes completely away *and* there's no lock contention on the borrow (the `clone` may be expensive, but once it's done, the lock is freed, so other datastore opcodes can also run.

avm/src/vm/opcode.rs
dfellis: So I want to make this effectively infallible, as well -- what these functions are *supposed* to do (and will eventually) is mark a node as down and then try the next node that's supposed to own the data. Worst case it decides the entire cluster is down besides itself and it should just fall back to the `ds.insert(nskey, hm)` logic.

That's not there yet, but I was hoping to make these improvements over time, so we can unblock the work on figuring out how to expose distributed datastore to the anycloud client applications in parallel.

avm/src/vm/opcode.rs
dfellis: I couldn't get this to compile.

avm/src/vm/opcode.rs
dfellis: So this is copy-pasted from [the current implementation](https://github.com/alantech/alan/blob/main/avm/src/vm/opcode.rs#L3693-L3694). I can try and see if it will work as you say, but I suspect I did it this way because it wasn't working otherwise.

avm/src/vm/opcode.rs
dfellis: Same as above, the None case should be nonexistent in the future.

avm/src/vm/opcode.rs
cdmistman: gotcha, maybe in the `match` above change it to
```rust
let ctrl_port = match ctrl_port {
  Some(ctrl_port) => ctrl_port.borrow().clone(),
  None => todo!("poll the rest of the cluster, fallback on local"),
};
```
or something like that?

dfellis: Wait, your issue is on the `unwrap` of the `ctrl_port`, not the return value of the `dssetv` call?

That's an impossible scenario. `is_key_owner` can only ever be `false` if `Some(ref ctrl_port)` was the path taken in generating that boolean. Therefore it is 100% guaranteed that `ctrl_port` is a `Some` value and I can `unwrap` it just fine.

avm/src/vm/opcode.rs
dfellis: Apparently it does compile just fine without this, so refactoring all of them.

avm/src/daemon/ctrl.rs
dfellis: 1. The `daemon` refuses to spin up without a cluster secret. And this path can only be called in `daemon` mode because the control port is not created otherwise.
2. I'll try it and see.

avm/src/daemon/daemon.rs
```diff
@@ -26,6 +27,7 @@ pub type DaemonResult<T> = std::result::Result<T, Box<dyn Error + Send + Sync>>;
 
 pub static CLUSTER_SECRET: OnceCell<Option<String>> = OnceCell::new();
 pub static DAEMON_PROPS: OnceCell<DaemonProperties> = OnceCell::new();
+pub static CONTROL_PORT_CHANNEL: OnceCell<Receiver<ControlPort>> = OnceCell::new();
```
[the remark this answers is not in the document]
dfellis: Yes. First once every `n` minutes when it reupdates its DNS records with the expected cluster configuration. Then every time it detects a down host. All part of follow-up work.

avm/src/daemon/ctrl.rs
dfellis: This should be an infallible path, though. `get_primary_node_id` should be returning an array index within the range of the `vms` since that `vms` array is literally what the LogRendezvousHash logic is indexing.

avm/src/daemon/ctrl.rs
```diff
@@ -400,4 +586,169 @@ impl ControlPort {
...
+    let mut res = self.client.request(req_obj).await?;
+    let bytes = hyper::body::to_bytes(res.body_mut()).await?;
+    Ok(std::str::from_utf8(&bytes)? == "true")
+  }
+
+  pub async fn dsdel(self: &ControlPort, key: &str) -> bool {
+    match self.dsdel_inner(key).await {
+      Ok(has) => has,
+      Err(_) => false,
+    }
```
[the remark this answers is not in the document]
dfellis:
```suggestion
    self.dsdel_inner(key).await.unwrap_or(false)
```

avm/src/daemon/ctrl.rs
dfellis:
```suggestion
    self.dssetf_inner(key, val).await.unwrap_or(false)
```

avm/src/daemon/ctrl.rs
dfellis:
```suggestion
    self.dssetv_inner(key, val).await.unwrap_or(false)
```

avm/src/daemon/ctrl.rs
cdmistman: i forgot this was in the daemon for a second

cdmistman (approved):

depombo (approved):

PR #513, opened 2021-05-06
Refactor dialoguer

aguillenv:


anycloud/cli/src/deploy/anycloud_dialoguer.rs
```diff
@@ -0,0 +1,61 @@
...
+
+// pub fn input_with_default_and_validation<F>(prompt: &str, default: String, validation_fn: F) -> String
+// where F: FnMut(&String) -> Result<(), &str> {
+//   Input::with_theme(&ColorfulTheme::default())
+//     .with_prompt(prompt)
+//     .validate_with(validation_fn)
+//     .default(default)
+//     .interact_text()
+//     .unwrap()
+// }
```
aguillenv, line 20: @cdmistman probably you could help me here. I'm trying to pass a closure to this function so the `.validate_with()` can receive it. As is the code I'm having this error message from the `.validate_with` method:
```
implementation of `std::ops::FnOnce` is not general enough

implementation of `std::ops::FnOnce` is not general enough

note: `std::ops::FnOnce<(&'0 std::string::String,)>` would have to be implemented for the type `F`, for some specific lifetime `'0`...
note: ...but `std::ops::FnOnce<(&std::string::String,)>` is actually implemented for the type `F`rustc 
```
This is the [link](https://docs.rs/dialoguer/0.8.0/dialoguer/struct.Input.html#method.validate_with) to `validate_with` docs.

I've tried different approaches without success.

anycloud/cli/src/deploy/mod.rs
```diff
@@ -213,48 +215,38 @@ fn get_aws_cli_creds() -> Result<AWSCLICredentialsFile, String> {
...
-    .with_prompt("Pick cloud provider for the new Credential")
-    .items(&clouds)
-    .default(0)
-    .interact()
-    .unwrap();
-  let cloud = clouds[selection];
-  let cred_name = Input::with_theme(&ColorfulTheme::default())
-    .with_prompt("Name for new Credential")
-    .validate_with(|input: &String| -> Result<(), &str> {
+  let clouds = vec!["AWS".to_string(), "GCP".to_string(), "Azure".to_string()];
```
cdmistman, line 218: i understand *why* there's a need for both `&str` and `String` but it never fails to disappoint me whenever things look like this because it could be so much easier if there was a way to unify them :(

cdmistman (approved):

anycloud/cli/src/deploy/mod.rs
aguillenv: agree

dfellis (approved):
I love net-negative PRs :) Greater clarity at the abstraction layer you're working at.

depombo (approved):
Amazing!

PR #515, opened 2021-05-10
Fix config and credentials override

aguillenv:


anycloud/cli/src/deploy/mod.rs
```diff
@@ -222,17 +222,21 @@ pub async fn add_cred() -> String {
...
+    } else {
+      Ok(())
+    }
+  };
+  let cred_name;
+  if credentials.contains_key(&default) {
+    cred_name = anycloud_dialoguer::input_with_validation(prompt, validator);
+  } else {
+    cred_name = anycloud_dialoguer::input_with_default_and_validation(prompt, default, validator);
+  }
```
cdmistman, line 239: Nit: can you turn this into `let cred_name = if credentials...`

anycloud/cli/src/deploy/mod.rs
```diff
@@ -501,17 +505,21 @@ pub async fn list_creds() {
...
+    } else {
+      Ok(())
+    }
+  };
+  let name;
+  if deploy_configs.contains_key(&default) {
+    name = anycloud_dialoguer::input_with_validation(prompt, validator);
+  } else {
+    name = anycloud_dialoguer::input_with_default_and_validation(prompt, default, validator);
+  }
```
cdmistman, line 522: Same here: `let name = if ...`

dfellis:
```suggestion
  let name = if deploy_configs.contains_key(&default) {
    anycloud_dialoguer::input_with_validation(prompt, validator)
  } else {
    anycloud_dialoguer::input_with_default_and_validation(prompt, default, validator)
  };
```

dfellis (approved):

PR #517, opened 2021-05-11
Control Port Extensions RFC

dfellis:


rfcs/018 - Control Port Extensions RFC.md
```diff
@@ -0,0 +1,121 @@
...
+3. Namespace both with `/avm/*` and `/app/*`
+
+The absolute safest path is option 3, as we can also add more namespaces in the future for other sources of control port functionality, however the split between "control port endpoints maintained by the AVM developers" and "control port endpoints maintained by AVM users" is a pretty solid split that is *unlikely* to run into issues in the future.
+
+Option 1 provides the least annoyance to application developers as they don't have `/app/` prefixed on every custom endpoint, while Option 2 is backwards compatible with how the AVM cluster is managed and coordinates with itself right now. Since this functionality is only for true power users of the language, and is not intended to be a publicly-facing endpoint, but glue for advanced integration needs with their own backend services, I am leaning towards Option 2.
+
+In the end, a service that establishes an HTTPS server and custom control port functionality would simply have two handlers, one for each:
+
+```ln
+from @std/httpserver import connection, controlPort, Connection, status, body, send
```
cdmistman, line 66: Since the control port will only be used with daemon mode, but not necessarily for every AVM programming using `@std/httpserver`, does it make sense to introduce a new std module? I think introducing `@std/avm/daemon` gives us plenty of flexibility in our runtime API while also being explicit that the module must only be used when running the AVM in daemon mode.

cdmistman: I still think it's the best solution - I think it keeps the API much cleaner and keeps us consistent in our past decisions on when and where to include built-in events in std modules. Maybe we can do `@std/avmdaemon` as a compromise instead, and once we start working on std caching we'll resolve the other issues? That'll also allow us to do `@std/http` and `@std/http/server` and the tcp equivalents

rfcs/018 - Control Port Extensions RFC.md
```diff
@@ -0,0 +1,122 @@
...
+
+However, this issue can be eliminated by recognizing that the endpoints are paths, and a path *prefix* can be used to namespace the two. There are three approaches:
+
+1. Namespace the AVM control port endpoints with `/avm/*`
+2. Namespace the Alan application control port endpoints with `/app/*`
+3. Namespace both with `/avm/*` and `/app/*`
+
+The absolute safest path is option 3, as we can also add more namespaces in the future for other sources of control port functionality, however the split between "control port endpoints maintained by the AVM developers" and "control port endpoints maintained by AVM users" is a pretty solid split that is *unlikely* to run into issues in the future.
+
+Option 1 provides the least annoyance to application developers as they don't have `/app/` prefixed on every custom endpoint, while Option 2 is backwards compatible with how the AVM cluster is managed and coordinates with itself right now. Since this functionality is only for true power users of the language, and is not intended to be a publicly-facing endpoint, but glue for advanced integration needs with their own backend services, I am leaning towards Option 2.
```
depombo, line 61: 100% agree with this

rfcs/018 - Control Port Extensions RFC.md
depombo: > I still think it's the best solution - I think it keeps the API much cleaner and keeps us consistent in our past decisions on when and where to include built-in events in std modules. Maybe we can do `@std/avmdaemon` as a compromise instead, and once we start working on std caching we'll resolve the other issues? That'll also allow us to do `@std/http` and `@std/http/server` and the tcp equivalents

That sounds good! We can even just do `@std/daemon` for now perhaps?

aguillenv (approved):

PR #518, opened 2021-05-12
Add anycloud npm module to version updates

depombo:


anycloud/node/README.md
```diff
@@ -1,48 +1,6 @@
 # AnyCloud
 
-AnyCloud is a Lambda alternative for Node.js that works with multiple cloud providers. We’re building the features of Lambda while fixing the problems we have run into while using it. Namely:
```
dfellis, line 3: This is the file that is rendered by NPM itself. Are you sure you want to gut it?

dfellis (approved):
Not a blocking change, but I think the README should remain useful.

anycloud/node/README.md
depombo: Yes. I prefer to centralize the documentation and information so we don't have to be constantly changing this. People won't really land on the npmjs site without reading about it somewhere else. Also things will remain just one click away.

aguillenv (approved):

PR #519, opened 2021-05-12
Build typescript for anycloud npm when releasing a new version

depombo:


.version.sh
```diff
@@ -46,5 +46,6 @@ cd -
 # Publish the anycloud/node to NPM with the new version
 cd anycloud/node
 rm -rf node_modules/ bin
+npm run build
```
dfellis, line 49: I think the order needs to be `npm install; npm run build; rm -rf node_modules/ bin; npm publish`

.version.sh
```diff
@@ -45,6 +45,9 @@ cd -
 
 # Publish the anycloud/node to NPM with the new version
 cd anycloud/node
+# make sure tsc/typescript is installed which is required for the build script
+yarn install
+yarn build
 rm -rf node_modules/ bin
```
dfellis, line 51: Maybe also add a comment here that we delete these so they don't accidentally end up in the published tarball since `bin` is populated based on your OS.

dfellis (approved):

PR #521, opened 2021-05-13
improvements to how-to aws psql prisma guide

depombo:
For easier review https://github.com/alantech/alan/blob/psql/anycloud/docs/how-to/aws-postgres.md

### DIFF
--- a/anycloud/docs/how-to/aws-postgres.md
+++ b/anycloud/docs/how-to/aws-postgres.md
@@ -4,6 +4,108 @@ The guide covers how to deploy a Node.js REST HTTP API to AWS using AnyCloud. Th
 
 All the code from this guide can be found in this [template repository](https://github.com/alantech/anycloud-node-aws-psql) which you can use to [create a new repository](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-from-a-template) as a starting point for your project.
 
+## Setup your project locally
+
+1) First download Prisma's example and install dependencies
+
+```
+curl https://codeload.github.com/prisma/prisma-examples/tar.gz/latest | tar -xz --strip=2 prisma-examples-latest/typescript/rest-express
+mv rest-express anycloud-node-aws-psql
+cd anycloud-node-aws-psql
+npm install
+```
+
+2) Install and start a local PostgreSQL instance by following Prisma's [guide](https://www.prisma.io/dataguide/postgresql/setting-up-a-local-postgresql-database).
+
+3) Now point your project's Node.js REST Prisma ORM to the local PostgreSQL instance you just created by setting the [datasource](https://www.prisma.io/docs/concepts/database-connectors/postgresql/) in the Prisma schema file at `anycloud-node-aws-psql/prisma/schema.prisma` to the environment variable `DATABASE_URL`.
+
+```
+datasource db {
+  provider = "postgresql"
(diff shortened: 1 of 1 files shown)


aguillenv (approved):

anycloud/docs/how-to/aws-postgres.md
```diff
@@ -65,15 +70,75 @@ RUN npm run build
...
 
-5) Use the AnyCloud CLI to create an `anycloud.json` file in the project directory and define a `Deploy Config`. You will need to pick a name or alias for the `Deploy Config`. The default value will be `staging`. You will also need to associate `Credentials` to this `Deploy Config`.
+8) Change HTTP server port in `src/index.ts` from `3000` to `8088` which is what what AnyCloud expects.
+
+```
+...
+
+const server = app.listen(8088, () =>
+  console.log(`
+🚀 Server ready at: http://localhost:8088
```
cdmistman, line 88: this looks weird just because it's not closed out...

dfellis (approved):
This looks complete and correct to me now.

I do also think this is also a wake-up call of sorts. There's a *lot* of extra steps due to RDS/Postgres being involved. It doesn't matter that we let people use their credits unlike some of the competition like Railway if their instructions are 1/4th the size of this :/

We have a good MVP, though, and for some the extra work with RDS isn't "that hard" because they've done it before, and the savings with our approach versus Kubernetes is worth it, but I think we need to PM up a roadmap for AnyCloud that tackles this developer experience deficit exposed by this blogpost.

depombo:
> This looks complete and correct to me now.
> 
> I do also think this is also a wake-up call of sorts. There's a _lot_ of extra steps due to RDS/Postgres being involved. It doesn't matter that we let people use their credits unlike some of the competition like Railway if their instructions are 1/4th the size of this :/
> 
> We have a good MVP, though, and for some the extra work with RDS isn't "that hard" because they've done it before, and the savings with our approach versus Kubernetes is worth it, but I think we need to PM up a roadmap for AnyCloud that tackles this developer experience deficit exposed by this blogpost.

100% agree and had the same thought myself. I'm still thinking about a proposal, but already on it. I see it as a spectrum of possibilities going from least involved to most involved:
- This guide in it's current state
- Making a script that users can run which spins up the Postgres Instance and connects it to one of their anycloud apps
- Offering something like ["add-ons" as Heroku does](https://elements.heroku.com/addons) so that the user can add Postgres in each cloud provider through the CLI and we move the code from the script described above into the `deploy` service
- We have also discussed offering a hosted Postgres ourselves. We spin up a Postgres in every VM and have a write main and multiple read replicas that can auto-promote if the main goes down. This is the most involved, but we can eventually offer a data migration tool to really squash vendor portability without having to use our datastore.

anycloud/docs/how-to/aws-postgres.md
```diff
@@ -65,15 +70,78 @@ RUN npm run build
...
+⭐️ See sample requests: http://pris.ly/e/ts/rest-express#3-using-the-rest-api`,
+  ),
+)
+
+...
+```
+
+9) Make sure everything works locally
+
+```javascript
```
cdmistman, line 98: should this maybe be bash instead?

depombo: yes!

PR #522, opened 2021-05-13
Write to (and delete from) all assigned nodes for a key in distributed datastore

dfellis:


dfellis:
I spun up a 4 node cluster, inserted a key, then took the cluster secret to manually craft control port queries to get the protobuf-encoded `/datastore/getv` payload on each of them (since it doesn't yet have auto-forwarding when requesting a key not assigned to it) to see if I get the value or an error as a way to verify these changes:

Screenshot from 2021-05-13 16-05-35
Screenshot from 2021-05-13 16-05-23
Screenshot from 2021-05-13 16-05-09
Screenshot from 2021-05-13 16-04-57


@aguillenv I was confused at first, but I realized that we're accidentally including the global memory when we're protobuf-encoding the HandlerMemory records. We should not do that as it is guaranteed to be identical across all nodes and is making the payload a lot larger than expected, so that seems like another task to tackle.

depombo (approved):

depombo:
> @aguillenv I was confused at first, but I realized that we're accidentally including the global memory when we're protobuf-encoding the HandlerMemory records. We should not do that as it is guaranteed to be identical across all nodes and is making the payload a lot larger than expected, so that seems like another task to tackle.

https://github.com/alantech/meta/issues/192

PR #526, opened 2021-05-18
[LNN] Operators

cdmistman:


compiler/src/aga.ts
```diff
@@ -33,7 +33,10 @@ const natural = OneOrMore.build(base10)
 const integer = And.build([ZeroOrOne.build(negate), natural])
 const real = And.build([integer, ZeroOrOne.build(And.build([dot, natural]))])
 const i8 = And.build([integer, Token.build('i8')])
+const i16 = And.build([integer, Token.build('i16')])
+const i32 = And.build([integer, Token.build('i32')])
 const i64 = And.build([integer, Token.build('i64')])
+const f32 = And.build([real, Token.build('f32')])
```
dfellis, line 39: I'm glad it was as simple as this to fix on the parsing side.

compiler/src/agatoagc.ts
```diff
@@ -31,9 +31,20 @@ const loadGlobalMem = (globalMemAst: LPNode): bigint[] => {
     if (value.has('i64')) {
       const val = BigInt(value.t.replace(/i64$/, ''))
       globalMem.push(val)
+    } else if (value.has('i32')) {
+      const val = BigInt(value.t.replace(/i32$/, ''))
+      globalMem.push(val)
+    } else if (value.has('i16')) {
+      const val = BigInt(value.t.replace(/i16$/, ''))
+      globalMem.push(val)
```
dfellis, line 39: Same here. Was afraid this'd be in three different places copy-pasted.

compiler/src/lnntoamm/Expr.ts
```diff
@@ -92,28 +93,202 @@ export default abstract class Expr {
...
+        // Will have to iterate through all of the interfaces in-scope and collect
+        // the applicable types as well
+        const op = work.get('operators').t.trim();
+        let operators = metadata.scope.get(op) as Operator[];
+        if (operators === null) {
+          console.log(metadata.scope);
+          throw new Error(`can't find operator ${op}`);
+        } else if (!isOpArray(operators)) {
+          // sanity check
+          throw new Error(`somehow ${op} isn't an operator?`);
```
dfellis, line 113: It may make sense to dump `operators` here, as well if this ever actually hits, because knowing what weird thing got in here would help a lot when tracing down the error.

compiler/src/lnntoamm/Expr.ts
```diff
@@ -92,28 +93,202 @@ export default abstract class Expr {
...
         throw new Error(`variables can't be assigned to operators`);
       }
-    } else {
-      return TODO('operators');
+      return operated[0] as [Stmt[], Expr];
+    }
+    // now we have to resolve operators - start by filtering out operators if they
+    // are in a position that must be prefix or infix
+    // since there are no suffix operators, this is relatively easy - operators
+    // immediately following an expression must be infix, while all others must be
```
dfellis, line 131:
```ln
const x = 5 + ++y;
```

(We don't have an increment operator, but we do need to make sure this kind of thing works.)

Maybe a more likely example:

```ln
const shouldRun = inValidState && !inUse;
```

compiler/src/lnntoamm/Expr.ts
```diff
@@ -200,20 +374,50 @@ class Call extends Expr {
...
+  */
   inline(amm: Output, kind: AssignKind, name: string, ty: Builtin) {
     const argTys = this.args.map(arg => arg.ty.instance());
     const selected = this.fns.reverse().find(fn => fn.acceptsTypes(argTys)) || null;
     // console.log('!!!!!!!!!!', this.ast.t.trim(), selected);
     if (selected === null) {
-      throw new Error(`no function selected: ${this.ast.t.trim()}`);
+      // TODO: to get better error reporting, we need to pass an ast when using
+      // operators. i'm not worried about error reporting yet, though :)
+      throw new Error(`no function selected`);
```
dfellis, line 420: Could be useful for now to also dump the `this.fns`, `this.args`, `kind` and `ty` variables?

dfellis (approved):

compiler/src/lnntoamm/Expr.ts
cdmistman: in these cases, the prefix operator is higher precedence so it gets resolved first. You're right that that'll trigger an operator ambiguity error though - I'll fix it

depombo (approved):

compiler/src/lnntoamm/Expr.ts
```diff
@@ -189,6 +189,75 @@ export default abstract class Expr {
...
+        if (val instanceof Expr) {
+          throw new Error(`uh, how?`)
+        }
+        // ensure that none of the operators disagree on fixity
+        const mode = prefixModeOf(val.get(prec));
+        if (mode === null) {
+          TODO('operator is both prefix and infix - how to determine?');
+        }
+      });
+      // first, prefix operators - we need to mutate idxs so no `.forEach`
```
dfellis, line 216: So precedence should still be the first order selection mechanism and then the `expr infix prefix expr` fix is that `prefix` wins out on ties, I think?

Eg if we had an increment prefix operator, and its predecence was below multiplication (and equal addition), then `6 + ++ 5 * 3` should produce `22`, not `24`. (`6 + ++ 5 * 3 -> 6 + ++ 15 -> 6 + 16 -> 22`)

dfellis: (if prefixes are always first regardless of operator predecence, we'd get `6 + ++ 5 * 3 -> 6 + 6 * 3 -> 6 + 18 -> 24`)

cdmistman: correct - we don't recalculate `prec` after completing the prefix operator handling so we're still at the same precedence level when we reach the infix operator loop. I just handle all of the prefix operators first to avoid having to repeatedly track forwards and backwards in the same loop

dfellis (approved):

PR #528, opened 2021-05-20
[LNN] method-style calling and paren groups

cdmistman:


bdd/spec/001_event_lnn_spec.sh
```diff
@@ -111,13 +111,13 @@ Describe "Events"
     INTOUTPUT="I got a number! 5"
 
     It "runs js"
-      Pending methods-interfaces-and-operators
+      Pending generics
```
dfellis, line 114: Why would this test still be pending? I don't see any generics in here?

compiler/src/lnntoamm/Expr.ts
```diff
@@ -84,8 +101,37 @@ export default abstract class Expr {
...
+            and,
+            (work as NamedAnd).filename,
+            work.line,
+            work.char,
+          );
+          TODO('closures/HOFs');
+        }
       } else {
-        // TODO: don't lump in HOF and chains
+          console.log(asts);
```
dfellis, line 134: Nit: `console.error`

compiler/src/lnntoamm/Expr.ts
```diff
@@ -115,6 +161,8 @@ export default abstract class Expr {
         }
         return operators;
       } else {
+        console.log(work)
```
dfellis, line 164: Nit 1: missing semicolon
Nit 2: `console.error`

compiler/src/lnntoamm/Expr.ts
```diff
@@ -84,8 +101,37 @@ export default abstract class Expr {
...
+          if (!assignableList.has('assignables') || assignableList.get('cdr').has(0)) {
+            console.log(assignableList, assignableList.has('assignables'), assignableList.get('cdr').has(0));
+            throw new Error(`unexpected token: found ${work.t.trim()} but it's not applied to a function`);
+          }
+          const [intermediates, res] = Expr.fromAssignablesAst(assignableList.get('assignables'), metadata);
+          generated.push(...intermediates);
+          expr = res;
+          skipDotIfNext();
+        } else {
+          // it's probably an HOF
```
dfellis, line 118: Not a higher order function, you mean a method call, right?

aguillenv (approved):

bdd/spec/001_event_lnn_spec.sh
cdmistman: ah, i was thinking there was some math in the test for some strange reason. I'll enable the test

compiler/src/lnntoamm/Expr.ts
cdmistman: it's most likely not a method call - that would get triggered under `work.has('variable')`. If that's *not* triggered, then that might mean that the ast we're looking at is `doSomething(1)(2)`, which is calling the fn that was returned from an HOF

dfellis: Ah, got it

dfellis (approved):

PR #539, opened 2021-05-21
Implement Control Port Extensions

dfellis:


dfellis:
Deployed the following code to https://ctrl-kv.alandeploy.com

Screenshot from 2021-05-21 15-03-13

It provides a read-only access to the key-val storage in public and read-write storage if you have access to the cluster secret:

Screenshot from 2021-05-21 15-02-53

Demonstrating that control port extensions work. :)

dfellis:
(The cluster has been terminated so that cluster secret is no longer real.)

aguillenv (approved):

cdmistman (approved):

compiler/browser/Std.js
```diff
@@ -19,18 +19,25 @@ module.exports = {
...
+            moduleAst.name,
+            moduleAst.ast,
+            rootModule.exportScope,
+            true
+          )
+          Module.getAllModules()[moduleAst.name] = stdModule
+        } catch (e) { // Failed to load, throw it back on the list to try again
+          moduleAst.name = currName;
+          stdAsts.push(moduleAst);
+        }
```
cdmistman, line 39: this can loop indefinitely, but I'm not worried about it since we'll be ensuring it doesn't loop in node

dfellis: Yep. That loop would also be caught by the `compiler-browser-check` in the CI, so I'm not worried about it, we'll catch it.

I would catch it!

depombo (approved):

PR #543, opened 2021-05-25
Expose cluster secret to AnyCloud container

aguillenv:
Expose cluster secreto to AnyCloud container

image

image


anycloud/cli/alan/anycloud.ln
```diff
@@ -19,9 +20,18 @@ on start {
     emit exit dockerRes.exitCode;
   }
 
+  const clusterSecret = getClusterSecret();
```
dfellis, line 23: What about putting this line in `@std/avmdaemon` as `export const clusterSecret = getClusterSecret();` and then just `from @std/avmdaemon import clusterSecret` here? You'd still need the `isNone()` check, but it might seem a little less awkward?

We're supposed to be able to export constants, and module-level constants can be defined by functions. Under the hood it's inlining the function call into whatever function you're actually using the constant in, so it should result in exactly the same AMM, but it might look cleaner?

dfellis: (My only concern about that statement, we never tested exported global constants that depends on a function that wasn't imported, so there might be a compiler bug with that that would also need addressing. In that case I would say to ignore this comment.)

aguillenv: Sounds good, will try it out and let you know.

avm/src/vm/opcode.rs
```diff
@@ -3824,6 +3824,22 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+  // cluster secret for avmdaemon
+  cpu!(getcs => fn(args, hand_mem) {
+    match CLUSTER_SECRET.get().unwrap() {
+      Some(cluster_secret) => {
+        hand_mem.init_fractal(args[2])?;
+        hand_mem.push_fixed(args[2], 1i64)?;
+        hand_mem.push_fractal(args[2], HandlerMemory::str_to_fractal(cluster_secret))?;
+      },
+      None => {
+        hand_mem.init_fractal(args[2])?;
```
cdmistman, line 3836: nit: can you put the `init_fractal` before the `match`?

anycloud/cli/alan/anycloud.ln
aguillenv: Exporting the const is giving me this error: `Cannot read property 'typename' of undefined`

dfellis (approved):

anycloud/cli/alan/anycloud.ln
cdmistman: Just realized that this doesn't work anyways. In order for that `const` to work, we'd have to allow it to be set at runtime, but I thought code can only be run within the context of event handling?

cdmistman (approved):

anycloud/cli/alan/anycloud.ln
dfellis: @cdmistman take a look at the BDD tests. Yo ncan have global consts and they can be based on functions.

What happens is any handler that *uses* them has the constant inserted at the point it is first used. Then if it is an actual static constant it will be hoisted away later,  and if not,  the copy stays there.

The issue here istthe function this constant is using is not getting imported and that trick fails because it can't be called in the other scope.

depombo (approved):

anycloud/cli/alan/anycloud.ln
dfellis: For clarity: https://github.com/alantech/alan/blob/v0.1.41/bdd/spec/020_module_constants_spec.sh

PR #544, opened 2021-05-25
Fix a bug in the first-time usage of 'anycloud' introduced with authentication

dfellis:
Thanks @cdmistman for spotting this.

### DIFF
--- a/anycloud/cli/src/oauth/mod.rs
+++ b/anycloud/cli/src/oauth/mod.rs
@@ -1,4 +1,4 @@
-use std::fs::{read_to_string, remove_file, File};
+use std::fs::{create_dir, read_to_string, remove_file, File};
 use std::io::prelude::*;
 use std::path::Path;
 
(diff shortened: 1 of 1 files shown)


anycloud/cli/src/oauth/mod.rs
```diff
@@ -119,6 +120,11 @@ async fn generate_token() {
     let json: Value = serde_json::from_str(&data_str).expect(ERR);
     if let Some(token) = json["access_token"].as_str() {
       let home = std::env::var("HOME").unwrap();
+      let dir_name = &format!("{}/{}", home, ANYCLOUD_DIR);
+      let path = Path::new(dir_name);
+      if !path.exists() {
+        create_dir(path).expect(ERR);
```
cdmistman, line 126: `create_dir` works like `mdir`, so [`create_dir` will fail if the path already exists](https://doc.rust-lang.org/std/fs/fn.create_dir.html#errors)

cdmistman: nevermind, there's the `exists` check

aguillenv (approved):

anycloud/cli/src/oauth/mod.rs
depombo: It would be more concise to move this file to use `OpenOptions` which does this for you when you set the `create` flag to `true` and it is how we handle the credentials and config file being created when it doesn't exist.

depombo (approved):

PR #546, opened 2021-05-25
Cluster status

cdmistman:


anycloud/cli/src/logger/mod.rs
```diff
@@ -44,15 +44,15 @@ macro_rules! error {
...
     eprintln!($($message)+);
     $crate::deploy::client_error(err_type, &format!($($message)+), "warn").await;
-  }};
-  (metadata: $metadata:tt, $errCode:ident, $($message:tt)+) => {async{
+  };
+  (metadata: $metadata:tt, $errCode:ident, $($message:tt)+) => {
     let err_type = $crate::logger::ErrorType::$errCode;
     let value = json!($metadata);
     eprintln!($($message)+);
     $crate::deploy::client_error(err_type, &format!($($message)+), "warn").await;
```
dfellis, line 56: How does this work if you `await in a non-async function?

cdmistman: You'd have to use one of `tokio`'s functions directly to achieve this. Assuming the non-async fn is called by an async fn in a tokio context, the code would look like this:
```rust
use tokio::runtime::Handle;
// before my change
Handle::current().block_on(warn!(...));
// after my change
Handle::current().block_on(async { warn!(...) });
```
neither are very convenient, but with my macro change there will be a compile error for not having an async block, whereas without it there's no compile error for not `.await`ing the future from the macro

anycloud/cli/src/deploy/mod.rs
```diff
@@ -1013,14 +1040,21 @@ pub async fn upgrade(
...
+    println!("No Apps deployed");
+    std::process::exit(0);
+  }
+  let (ids, sizes): (Vec<String>, Vec<usize>) = apps.into_iter().map(|a| (a.id, a.size)).unzip();
   let selection = anycloud_dialoguer::select_with_default("Pick App to upgrade", &ids, 0);
   let cluster_id = &ids[selection];
   CLUSTER_ID.set(cluster_id.to_string()).unwrap();
   let styled_cluster_id = style(cluster_id).bold();
   let config = get_config().await;
-  let sp = ProgressBar::new_spinner();
```
aguillenv, line 1023: If you `finish_and_clear` on line 1047 shouldn't you need to call `new_spinner` again?

anycloud/cli/src/deploy/mod.rs
```diff
@@ -993,7 +1008,19 @@ pub async fn new(
   }
   let resp = post_v1("new", body).await;
   let res = match resp {
-    Ok(res) => format!("Created App {} successfully!", style(res).bold()),
+    Ok(res) => {
+      // idc if it's been set before, I'm setting it now!!!
+      let _ = CLUSTER_ID.set(res);
```
aguillenv, line 1013: Does the `let _ =` is necessary?

aguillenv: And, the `set` could be call without unwrapping? didn't know that.

anycloud/cli/src/deploy/mod.rs
cdmistman: yes, good catch - fixing

anycloud/cli/src/deploy/mod.rs
cdmistman: the `let _ =` is to prevent rust warnings on not using the `Result` - basically, if you `unwrap` the result of `set` and there was already a value, the cli will crash. if the `Result` is assigned to `_`, then it won't crash and won't cause a compiler warning :)

dfellis (approved):

aguillenv (approved):

PR #549, opened 2021-05-26
Homebrew update on release

aguillenv:


.github/workflows/release.yml
```diff
@@ -138,3 +138,18 @@ jobs:
         tag: ${{ github.ref }}
         file: /tmp/anycloud-macos.tar.gz
         asset_name: anycloud-macos.tar.gz
+
+    - name: Update homebrew
+      # TODO: should we create a new token or reuse the AUTOPUBLISH one?
+      run: |
+        alanversion=$(echo ${{github.ref}} | sed "s/refs\/tags\///")
+        [[ $alanversion != *'beta'* ]] && \
```
dfellis, line 146: Is this syntax right? The stars on the outside of the single quotes is so weird.

.github/workflows/release.yml
```diff
@@ -138,3 +138,18 @@ jobs:
         tag: ${{ github.ref }}
         file: /tmp/anycloud-macos.tar.gz
         asset_name: anycloud-macos.tar.gz
+
+    - name: Update homebrew
+      # TODO: should we create a new token or reuse the AUTOPUBLISH one?
```
dfellis, line 143: Reusing the autopublish token should be fine, I think.

.github/workflows/release.yml
```diff
@@ -138,3 +138,18 @@ jobs:
...
+        alanversion=$(echo ${{github.ref}} | sed "s/refs\/tags\///")
+        [[ $alanversion != *'beta'* ]] && \
+        curl \
+        --fail \
+        -XPOST \
+        -H "Accept: application/vnd.github.everest-preview+json" \
+        -H "Content-Type: application/json" \
+        -H "authorization: token ${{secrets.AUTOPUBLISH}}" \
+        https://api.github.com/repos/alantech/homebrew-core/actions/workflows/update.yml/dispatches \
+        -d '{"ref": "auto-update", "inputs": {"alan_version": "'"$alanversion"'"}}' \
```
dfellis, line 154: Has this been tested? I'm pretty sure bash single quotes do not allow `$var` substitution. I also don't know what's going on with the `"'"$...` part.

I think this needs to be something like:

```bash
-d "{\"ref\": \"auto-update\", \"inputs\": {\"alan_version\": \"$alanversion\"}}" \
```

.github/workflows/release.yml
aguillenv: Ok, will remove the TODO

.github/workflows/release.yml
aguillenv: Yes, it is bash partial matching

dfellis:
I'm more concerned with this one; testing it is hard. But once it's figured out, we should be able to use it as a pattern on the Arch Linux repo; we push it to Github as well, add an action there to auto-update itself with the SHAs of the alan and anycloud binaries, and then push to Arch's git server (after uploading the account access as a secret in github, as well).

.github/workflows/release.yml
aguillenv: Yes, I have tested both, release and prerelease and it works. This give me some war but how I have it it works

aguillenv: I have to update the `ref` to `main`, though

dfellis (approved):

.github/workflows/release.yml
aguillenv: The `"'"$...` it is because it is a JSON, in that way I concatenate strings

cdmistman (approved):

.github/workflows/release.yml
dfellis: Oh! That's clever!

depombo (approved):

PR #559, opened 2021-06-01
Npm datastore

aguillenv:
Tested publishing my fork of the repo. 

Users code could look like the following:

image

Example result:

image


anycloud/node/package.json
```diff
@@ -30,6 +30,7 @@
     "typescript": "^4.1.5"
   },
   "dependencies": {
-    "@types/node": "^14.14.28"
+    "@types/node": "^14.14.28",
+    "alan-js-runtime": "file:../../js-runtime"
```
aguillenv, line 34: I'm doing this for local test

anycloud/node/src/datastore.ts
```diff
@@ -1,32 +1,67 @@
...
+  private clusterSecret?: string;
+  private isLocal: boolean;
+  private headers?: any;
+  private ctrlPortUrl: string;
+
+  constructor() {
+    this.localDS = {};
+    this.ns = 'kv';
+    this.clusterSecret = process.env.CLUSTER_SECRET;
+    this.isLocal = this.clusterSecret ? false : true;
```
dfellis, line 15: Nit: this could just be `this.isLocal = !this.clusterSecret;`

anycloud/node/src/datastore.ts
```diff
@@ -1,32 +1,67 @@
...
+  }
+
+  private async request(method: string, url: string, body?: any): Promise<Response> {
+    console.log('localpath', this.isLocal);
+    return fetch(url, {
+      method: method,
+      headers: this.headers,
+      body,
+    });
+  }
```
dfellis, line 27: I would `await` on the `fetch` call in here and return the promise for the body.

Otherwise:
1. This doesn't need to be an `async` function at all
2. You're requiring users of the functions below to `await (await ds.get('foo'))` which is awkward.

anycloud/node/src/datastore.ts
```diff
@@ -1,32 +1,67 @@
...
+  private localDS: any;
+  private ns: string;
+  private clusterSecret?: string;
+  private isLocal: boolean;
+  private headers?: any;
+  private ctrlPortUrl: string;
+
+  constructor() {
+    this.localDS = {};
+    this.ns = 'kv';
```
dfellis, line 13: Nit: Maybe this field should be an optional constructor argument? Not necessary for now, but maybe in the future when we want to power other things with Datastore, too.

anycloud/node/src/datastore.ts
```diff
@@ -1,67 +1,77 @@
...
     this.localDS = {};
-    this.ns = 'kv';
     this.clusterSecret = process.env.CLUSTER_SECRET;
     this.isLocal = this.clusterSecret ? false : true;
-    this.headers = this.clusterSecret ? {[this.clusterSecret]: 'true'} : null;
-    this.ctrlPortUrl = 'https://localhost:4142';
+    this.headers = this.clusterSecret ? { [this.clusterSecret]: 'true' } : null;
+    this.ctrlPortUrl = 'https://localhost:4142/app/';
+    // To avoid failure due to self signed certs
+    https.globalAgent.options.rejectUnauthorized = false;
```
dfellis, line 18: We are gonna want to constrain that to our library and not mutate the global agent or this library becomes a security vulnerability. We can do this only because we are directly connecting to localhost, but this change would affect all domains.

aguillenv: Yes, I was going to ask about this. This change the global request agent and that's not ok, but the `cross-fetch` package does not support pass a custom `agent`. I've seen [`node-fetch`](https://www.npmjs.com/package/node-fetch) and apparently we can instantiate a new agent and pass it through the fetch. Should we change library?

anycloud/node/src/datastore.ts
aguillenv: I've remove it since we handle this right now on `anycloud.ln` event handler. I think it would make sense to add it once we need it?

anycloud/node/src/datastore.ts
aguillenv: I've change the return values of all functions and did some changes before reading this comment. But I'm still returning the fetch. I guess a could `await` here as you suggest and just return the response to be handled by each method.

anycloud/node/src/datastore.ts
dfellis: That's probably even better; put as much of that into the control port logic on the anycloud.ln file so users can't accidentally break it.

anycloud/node/src/datastore.ts
dfellis: Yeah; thinking about it this library is never going to be used in a browser context so polyfilling isn't necessary.

anycloud/node/package-lock.json
```diff
@@ -0,0 +1,54 @@
+{
```
dfellis, line 1: We should probably `.gitignore` this file.

anycloud/node/package.json
```diff
@@ -3,13 +3,16 @@
...
   "bin": {
     "anycloud": "./bin/anycloud"
   },
   "scripts": {
     "build": "tsc",
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "preinstall": "node ./get-anycloud.js"
+    "test": "jest",
+    "preinstall": "node ./get-anycloud.js",
+    "beforepublish": "jest && rm -rf bin && tsc",
```
dfellis, line 14: I don't think this one exists. You want [`prepublishOnly`](https://docs.npmjs.com/cli/v6/using-npm/scripts).

anycloud/node/src/datastore.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  async del(dsKey: string): Promise<boolean> {
+    if (!dsKey) return false;
+    if (this.isLocal) {
+      return dsKey in this.localDS ? delete this.localDS[dsKey] : false;
+    }
+    try {
+      return await this.request('GET', `${this.ctrlPortUrl}del/${dsKey.toString()}`) === 'true';
```
cdmistman, line 73: whyyyyyy does `delete <symbol>` return `true` not only if the key already existed, but if the key didn't exist?

cdmistman: why does it return `true` at all????

anycloud/node/src/datastore.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (this.isLocal) {
+      this.localDS[dsKey] = dsValue;
+      return 'ok';
+    }
+    try {
+      return await this.request('POST', `${this.ctrlPortUrl}set/${dsKey.toString()}`, dsValue);
+    } catch (_) {
+      return 'fail';
+    }
+  }
```
dfellis, line 65: Why not always `JSON.stringify`ing the `dsValue` here and always `JSON.parse`ing it in the `get` method above? That would eliminate the weird behavior for local-mode get versus the rest and that everything is actually a string is hidden from the users of this.

anycloud/node/src/datastore.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (this.isLocal) {
+      return dsKey in this.localDS;
+    }
+    try {
+      return await this.request('GET', `${this.ctrlPortUrl}has/${dsKey.toString()}`) === 'true';
+    } catch (_) {
+      return false;
+    }
+  }
+}
```
dfellis, line 90: I'm also curious about using this class to power a proxy object, instead of exposing it directly. It's true you can't check if a key is present or not, but maybe that's fine since they're remote values anyways so you wouldn't want the network calls to be inserted into a `for-in` loop?

cdmistman (approved):

anycloud/node/package.json
aguillenv: Yes, it does not, just add it to run it manually temporally.

anycloud/node/src/datastore.ts
dfellis: Screenshot from 2021-06-03 11-13-28

lolwat.

Though I think I know why.

Screenshot from 2021-06-03 11-15-56

You can define undeleteable properties on objects, so the boolean indicates "does this property no longer exist".

anycloud/node/src/datastore.ts
dfellis: Maybe arrogant, but we can just always assume our code works. ;)

anycloud/node/src/datastore.ts
cdmistman: interesting... still a very weird decision and i don't like it, nope, not one bit

anycloud/node/src/datastore.ts
```diff
@@ -34,31 +34,35 @@ export class DataStore {
...
     if (this.isLocal) {
       if (dsKey in this.localDS) {
-        const val = this.localDS[dsKey]; 
-        if (typeof(val) === 'object') {
-          return JSON.stringify(val);
-        } else {
-          return `${val}`;
-        }
+        const val = this.localDS[dsKey];
+        return JSON.stringify(val);
```
dfellis: This should be `JSON.parse(val)`, right?

anycloud/node/src/datastore.ts
```diff
@@ -34,31 +34,35 @@ export class DataStore {
...
       } else {
         return undefined;
       }
     }
     try {
-      return await this.request('GET', `${this.ctrlPortUrl}get/${dsKey.toString()}`);
+      return JSON.stringify(await this.request('GET', `${this.ctrlPortUrl}get/${dsKey.toString()}`));
```
dfellis: Same here.

anycloud/node/src/datastore.ts
```diff
@@ -34,31 +34,35 @@ export class DataStore {
...
     } catch (e) {
       return e;
     }
   }
 
   async set(dsKey: string, dsValue: any): Promise<string> {
     if (!dsKey) return 'fail';
+    const parsedDSValue = (function () {
```
dfellis: This name is off. This is the `stringifiedDSValue`, right?

anycloud/node/src/datastore.ts
```diff
@@ -34,31 +34,35 @@ export class DataStore {
...
     }
     try {
-      return await this.request('GET', `${this.ctrlPortUrl}get/${dsKey.toString()}`);
+      return JSON.stringify(await this.request('GET', `${this.ctrlPortUrl}get/${dsKey.toString()}`));
     } catch (e) {
       return e;
     }
   }
 
   async set(dsKey: string, dsValue: any): Promise<string> {
```
dfellis, line 56: While we use the `fail/ok` strings for the HTTP responses, this could be a `Promise<boolean>` instead and it's just `return false;` or `return true;`, which is probably going to be less confusing in JS. (The "boolean-equivalent" in the REST API was the response code of 200/400.)

anycloud/node/src/datastore.ts
dfellis: Also, absorbing the error like this may not be the best idea, because *why* the set failed is unclear.

I think having this be something like `Promise<Error | undefined>` might be a better choice? You'd still need to try-catch it, but now the end users can check if they got an error object, something like:

```js
async set(dsKey: string, dsValue: any): Promise<Error | undefined> {
  if (!dsKey) return new Error("No key was provided");
  let stringifiedDSValue;
  try {
    stringifiedDSValue = JSON.stringify(dsValue);
  } catch (e) {
    return e;
  }
  if (this.isLocal) { ...
```

anycloud/node/src/datastore.ts
aguillenv: Yes, I think you were reviewing an out of dated version, sorry.

anycloud/node/src/datastore.ts
aguillenv: same

anycloud/node/src/datastore.ts
aguillenv: I've remove the `null` assignment because it made no sense. What I have done is to save the value even though it is no JSON serializable, so in the catch just assign the incoming value. Because is is a `key:value` datastore, I think we should not make it only accept JSON serializable objects. Nevertheless, your suggestion of returning an `Error` object makes sense and we must do something like that. Not sure if `Promise<Error | undefined>`, I think maybe  Promise<Error | boolean> makes more sense following your other comment below.

anycloud/node/src/datastore.ts
```diff
@@ -30,65 +34,94 @@ export class DataStore {
...
-        return JSON.stringify(val);
+        return JSON.parse(val) || val;
       } else {
         return undefined;
       }
     }
     try {
-      return JSON.stringify(await this.request('GET', `${this.ctrlPortUrl}get/${dsKey.toString()}`));
+      const dsValue = await this.request('GET', `${this.ctrlPortUrl}get/${dsKey}`);
+      return JSON.parse(dsValue) || dsValue;
```
dfellis, line 50: This won't work like you want. `JSON.parse("false") || "false"` will return `"false"` instead of the actual boolean.

anycloud/node/src/datastore.ts
```diff
@@ -30,65 +34,94 @@ export class DataStore {
...
+      if (DataStore.isInvalidKey(dsKey)) {
+        return false;
+      }
+      ds.del(dsKey);
+      return true;
+    }
+  };
+};
+
+export const datastore = new Proxy({}, dsHandler());
```
dfellis, line 134: I wouldn't mind also exporting an instance of the `DataStore` class (Not sure about that name, though. It's technically 2 words in English, but "datastore" is one logical name in our API, I'd just call it the `Datastore` class.)

Maybe the proxy version gets the name `ds` and the class instance `datastore`?

anycloud/node/src/datastore.ts
aguillenv: Yes, I've added the try/catch

anycloud/node/src/index.ts
```diff
@@ -1,7 +1,7 @@
-import { DataStore } from './datastore';
+import { datastore } from './datastore';
 
 export const PORT = 8088; // http://img2.wikia.nocookie.net/__cb20080604102014/starwars/images/3/3b/8t88.png
-export const DS = new DataStore();
+export const DS = datastore;
```
dfellis, line 4: So the `PORT` is all caps because it is an immutable constant, but this thing isn't, so I'm not sure about the allcaps namin here.

anycloud/node/src/datastore.ts
```diff
@@ -30,65 +34,94 @@ export class DataStore {
     });
     return await response.text();
   }
+
   async get(dsKey: string): Promise<string> {
```
dfellis: This return type is still wrong, since you `return undefined` in some paths. It should be `Promise<string | undefined>`

anycloud/node/src/datastore.ts
```diff
@@ -48,7 +48,7 @@ class DataStore {
       try {
         dsValue = await this.request('GET', `${this.ctrlPortUrl}get/${dsKey}`);
       } catch (e) {
-        return Error(e);
+        return new Error(e);
```
dfellis, line 51: Isn't `e` already an Error object? Why the rewrapping?

anycloud/node/src/datastore.ts
aguillenv: I have struggling with the naming haha, because we define the name here but later we have to give another one for the export in the `index.ts`. 
So what do you think of this:
`datastore.ts`:
- proxy: `dsProxy`
- class: `DatastoreClass`

`index.ts`:
```
export const DataStore = DatastoreClass;
export const ds = dsProxy;
```

anycloud/node/src/datastore.ts
aguillenv: you are completely right, my bad.

anycloud/node/src/datastore.ts
dfellis: So that second export in `index.ts` doesn't need to be a rename at all. And we don't have to export the raw class, just the instance of the class. I'd just do:

`datastore.ts`
```ts
export const datastore = new Datastore();
export const ds = new Proxy({}, dsHandler());
```

`index.ts`
```ts
import { datastore, ds } from './datastore'
export datastore
export ds
```

dfellis: https://stackoverflow.com/a/30714301

dfellis (approved):

anycloud/node/src/index.ts
```diff
@@ -1,6 +1,6 @@
+export { datastore, ds } from './datastore';
 export const PORT = 8088; // http://img2.wikia.nocookie.net/__cb20080604102014/starwars/images/3/3b/8t88.png
 
 export default {
-  PORT,
+  PORT
```
cdmistman, line 5: nit: any reason to get rid of the trailing comma? Also, why do we `export const` the port and provide it in `default`? Doesn't that do the same thing?

dfellis: So this is one piece that I was a little confused about but didn't want to make a fuss for now.

But what this `export default { ... }` does is create an object that is what's actually imported if you just `import anycloud from 'anycloud';`

Ideally, this would also have `datastore` and `ds` properties in it so then you can do `anycloud.ds['foo'] = 'bar';` and that would also work.

The idea behind this is so you don't have to write `import * as anycloud from 'anycloud';`

Maybe we should just delete this because it doesn't seem to bring much value.

aguillenv: Yes, I have leave only the named exports

cdmistman (approved):

depombo (approved):

PR #563, opened 2021-06-07
Log anycloud.ln errors

aguillenv:
Fixes: https://github.com/alantech/meta/issues/191

@dfellis would like your input with this solution. It is not working, yet. But before going to the long test phase, would like to know your opinion on the idea.

----- UPDATE -----

For the moment it logs the error.
image

Probably in a follow up task in the future would be nice to add the structure that let communicate anycloud.ln and AVM. With this structure we could add more detail to the log, as the cluster id.


dfellis:
The first thing that's not quite right with this is setting a value in memory to share on a follow-up request, and then still having the `emit exit errCode` lines that will cause the AVM to terminate immediately afterwards. The deploy management service will still be unable to query that value out since the process will be down.

It would probably be better to just `post('https://deploy.anycloudapp.com/newEndpoint', 'the payload');` in that case. There's no special logic required in the AVM *at all* for this that way.

Alternatively, we replace `emit exit errCode;` with just `return;` to early terminate the event handler and set that value which can be pushed to the deploy service on the next status update, but we also need some logic for the non-leader nodes to emit the status in that situation.

But in general, this is also pointing towards an issue with AVM/Alan introspection/interop if we need a new opcode for each one. It might be worthwhile to come up with a more generic mechanism for the two to share data back and forth, perhaps some sort of AVM struct that has a matching type on the Alan side, though it'd be nice if we could somehow sidestep the memory model representation issues.

aguillenv:
> The first thing that's not quite right with this is setting a value in memory to share on a follow-up request, and then still having the `emit exit errCode` lines that will cause the AVM to terminate immediately afterwards. The deploy management service will still be unable to query that value out since the process will be down.
> 
> It would probably be better to just `post('https://deploy.anycloudapp.com/newEndpoint', 'the payload');` in that case. There's no special logic required in the AVM _at all_ for this that way.
> 
> Alternatively, we replace `emit exit errCode;` with just `return;` to early terminate the event handler and set that value which can be pushed to the deploy service on the next status update, but we also need some logic for the non-leader nodes to emit the status in that situation.
> 
> But in general, this is also pointing towards an issue with AVM/Alan introspection/interop if we need a new opcode for each one. It might be worthwhile to come up with a more generic mechanism for the two to share data back and forth, perhaps some sort of AVM struct that has a matching type on the Alan side, though it'd be nice if we could somehow sidestep the memory model representation issues.

Ok, good idea. I think for now the cleanest way might be send the error to the `/clientError` endpoint or a new one that log it. And we can start thinking on the structure that you comment.

cdmistman (approved):

anycloud/cli/alan/anycloud.ln
```diff
@@ -158,3 +133,29 @@ fn toString(reqBody: ServiceRequestBody): string {
   const messageStr = '"message": "' + reqBody.message + '"';
   return '{' + errorCodeStr + ', ' + alanVersionStr + ', ' + levelStr + ', ' + messageStr + '}';
 }
+
+fn getCleanError(error: string): string {
+  return error
+    .split("\n").join(" ")
+    .split("").map(fn(value: string): string {
+      return value == '"' ? "'" : value;
```
dfellis, line 141: Maybe replacing `"` with `\"` would be better?

aguillenv: Yes, actually would be thee proper json serialization of the string. will try it

anycloud/cli/alan/anycloud.ln
```diff
@@ -138,7 +138,7 @@ fn getCleanError(error: string): string {
   return error
     .split("\n").join(" ")
     .split("").map(fn(value: string): string {
-      return value == '"' ? "'" : value;
+      return value == '"' ? '\"' : value;
```
dfellis, line 141: You may need to write this as `'\\"'` but I don't remember if we parse escape characters in single-quote strings or not.

aguillenv: I've test it as is and it worked

dfellis: Then I guess we have a difference between single-quote and double-quote strings. Whether we want to keep that or not is something else, though.

cdmistman: iirc we decided not to change the behavior in the current front-end, but to change the behavior with the change to the new front-end (since there will be some other breaking changes, ie in `@std`)

dfellis: For future reference, there wasn't actually a difference. :)

dfellis (approved):

PR #564, opened 2021-06-09
Create a tcptun opcode to do direct TCP tunneling

dfellis:
Got js-runtime and avm in plain HTTP mode implemented, still need to
implement the automatic SSL unwrapping


aguillenv (approved):

dfellis:
Screenshot from 2021-06-10 09-01-52
Proof the load balancing is working. They both are burning the same amount of CPU with my load-balance test app, and the node that's getting the incoming requests is burning more network packets than the other as it has to forward ~50% of the incoming requests to its peer.

avm/src/vm/http.rs
```diff
@@ -120,3 +120,181 @@ macro_rules! make_server {
...
+                loop {
+                  let l = REGION_VMS.read().unwrap().len();
+                  let i = async move {
+                    let mut rng = thread_rng();
+                    rng.gen_range(0..=l)
+                  }
+                  .await;
+                  let accept = tcp.accept().await;
+                  if accept.is_err() { continue; }
+                  let (mut socket, source_addr) = accept.unwrap();
```
cdmistman, line 216: nit: can you make this a match? this should work
```suggestion
                  let (mut socket, source_addr) = match tcp.accept().await {
                    Ok(accepted) => accepted,
                    Err(_) => continue,
                  };
```

avm/src/vm/http.rs
```diff
@@ -120,3 +120,181 @@ macro_rules! make_server {
...
+                  if i != l {
+                    let region_vms = REGION_VMS.read().unwrap().clone();
+                    let source_ip = source_addr.ip().to_string();
+                    if region_vms.iter().any(|ip| *ip == source_ip) {
+                      let strm = tls_acceptor.accept(socket).into_failable();
+                      let strm_val = strm.await;
+                      if strm_val.is_err() { continue; }
+                      yield Ok(strm_val.unwrap());
```
cdmistman, line 224: nit: can you turn this into a `match`?

avm/src/vm/http.rs
```diff
@@ -120,3 +120,181 @@ macro_rules! make_server {
...
+                          region_vms[i], ee
+                        ),
+                      };
+                    });
+                    continue;
+                  }
+                  let strm = tls_acceptor.accept(socket).into_failable();
+                  let strm_val = strm.await;
+                  if strm_val.is_err() { continue; }
+                  yield Ok(strm_val.unwrap());
```
cdmistman, line 249: nit: can you turn this into a `match`

cdmistman (approved):

avm/src/vm/http.rs
dfellis: Are you sure that works? A `continue` inside of an assignment statement doesn't make any sense.

dfellis: Just compiled it and it worked, still seems bonkers to me.

PR #565, opened 2021-06-09
Non interactive CLI

aguillenv:
This will make easier the integration with CI/CD tools since it would not need user's interaction. It will be based on credentials as environment variables defined by the user a some flags. The new flags are `--non-interative` that will let us know that we need to work in a non interactive way, `––app-name <name>` and `--config-name <name>` will improve current CLI experience since we can give this information right away.

Github actions example:
image

image

image


dfellis:
Why environment variables instead of new flags in `clap`?

aguillenv:
> Why environment variables instead of new flags in `clap`?

I think that enabling all those "configuration" flags in `clap` will pollute the CLI experience.

dfellis:
Discussed offline. The key is the `-n` non-interactive flag and how it will provide an error message if a required environment variable for a secret is missing.

anycloud/cli/src/main.rs
```diff
@@ -17,6 +17,9 @@ pub async fn main() {
     .subcommand(SubCommand::with_name("new")
       .about("Deploys your repository to a new App with a Deploy Config from anycloud.json")
       .arg_from_usage("-e, --env-file=[ENV_FILE] 'Specifies an optional environment file'")
+      .arg_from_usage("[NON_INTERACTIVE] -n, --non-interactive 'Specifies an optional flag for non interactive CLI mode'")
```
dfellis, line 20: This reads a bit weird "specifies an optional flag".

Maybe:

```suggestion
      .arg_from_usage("[NON_INTERACTIVE] -n, --non-interactive 'Enables non interactive CLI mode useful for scripting'")
```

anycloud/cli/src/main.rs
```diff
@@ -17,6 +17,9 @@ pub async fn main() {
...
+      .arg_from_usage("-a, --app-name=[APP_NAME] 'Specifies an optional app name.'")
+      .arg_from_usage("-c, --config-name=[CONFIG_NAME] 'Specifies an optional config name.'")
```
dfellis, line 22: Also maybe:

```suggestion
      .arg_from_usage("-c, --config-name=[CONFIG_NAME] 'Specifies a config name, required only in non-interactive mode.'")
```

anycloud/cli/src/main.rs
```diff
@@ -27,6 +30,9 @@ pub async fn main() {
     .subcommand(SubCommand::with_name("upgrade")
       .about("Deploys your repository to an existing App hosted in one of the Deploy Configs from anycloud.json")
       .arg_from_usage("-e, --env-file=[ENV_FILE] 'Specifies an optional environment file relative path'")
+      .arg_from_usage("[NON_INTERACTIVE] -n, --non-interactive 'Specifies an optional flag for non interactive CLI mode'")
```
dfellis, line 33: Same suggestions here (in case you want to use them)

```suggestion
      .arg_from_usage("[NON_INTERACTIVE] -n, --non-interactive 'Enables non-interactive CLI mode useful for scripting'")
```

anycloud/cli/src/main.rs
```diff
@@ -27,6 +30,9 @@ pub async fn main() {
...
+      .arg_from_usage("-a, --app-name=[APP_NAME] 'Specifies an optional app name.'")
+      .arg_from_usage("-c, --config-name=[CONFIG_NAME] 'Specifies an optional config name.'")
```
dfellis, line 35:
```suggestion
      .arg_from_usage("-c, --config-name=[CONFIG_NAME] 'Specifies a config name, required only in non-interactive mode.'")
```

anycloud/cli/src/main.rs
```diff
@@ -61,40 +67,50 @@ pub async fn main() {
...
+        matches.value_of("env-file"),
+        matches.value_of("app-name"),
+        matches.value_of("config-name"),
+        non_interactive,
       )
       .await;
     }
-    ("terminate", _) => deploy::terminate().await,
+    ("terminate", _) => {
+      authenticate(false).await;
```
dfellis, line 89: Should we have a `TODO` comment here to allow non-interactive termination in the future?

A normal CI/CD usage wouldn't need it, but if you had a test that fully spun up and shut down a test service, it would be necessary.

anycloud/cli/src/main.rs
```diff
@@ -117,6 +134,57 @@ pub async fn main() {
...
+        Some((dockerfile_b64, app_tar_gz_b64)),
+        env_b64,
+        app_name,
+        config_name,
+        non_interactive,
+      )
+      .await
+    }
+    _ => {}
+  };
```
dfellis, line 189: Nit: Make this an enum instead of a string and you don't need the `_ => {}` bit at the end?

Alternatively, instead of `mode: &str` you pass the async function to call and you don't need the match at all? (That one is more clever but less extensible in the future if the arguments ever change between one or the other.)

avm/src/main.rs
```diff
@@ -61,6 +61,9 @@ fn main() {
       .subcommand(SubCommand::with_name("new")
         .about("Deploys an .agz file to a new app with one of the Deploy Configs from anycloud.json")
         .arg_from_usage("<AGZ_FILE> 'Specifies the .agz file to deploy'")
+        .arg_from_usage("[NON_INTERACTIVE] -n, --non-interactive 'Specifies an optional flag for non interactive CLI mode'")
```
dfellis, line 64:
```suggestion
        .arg_from_usage("[NON_INTERACTIVE] -n, --non-interactive 'Enables non interactive CLI mode useful for scripting'")
```

avm/src/main.rs
```diff
@@ -61,6 +61,9 @@ fn main() {
...
+        .arg_from_usage("-a, --app-name=[APP_NAME] 'Specifies an optional app name.'")
+        .arg_from_usage("-c, --config-name=[CONFIG_NAME] 'Specifies an optional config name.'")
```
dfellis, line 66:
```suggestion
        .arg_from_usage("-c, --config-name=[CONFIG_NAME] 'Specifies a config name, required only in non-interactive mode.'")
```

avm/src/main.rs
```diff
@@ -71,6 +74,9 @@ fn main() {
       .subcommand(SubCommand::with_name("upgrade")
         .about("Deploys your repository to an existing App hosted in one of the Deploy Configs from anycloud.json")
         .arg_from_usage("<AGZ_FILE> 'Specifies the .agz file to deploy'")
+        .arg_from_usage("[NON_INTERACTIVE] -n, --non-interactive 'Specifies an optional flag for non interactive CLI mode'")
```
dfellis, line 77:
```suggestion
        .arg_from_usage("[NON_INTERACTIVE] -n, --non-interactive 'Enables non interactive CLI mode useful for scripting'")
```

avm/src/main.rs
```diff
@@ -71,6 +74,9 @@ fn main() {
...
+        .arg_from_usage("-a, --app-name=[APP_NAME] 'Specifies an optional app name.'")
+        .arg_from_usage("-c, --config-name=[CONFIG_NAME] 'Specifies an optional config name.'")
```
dfellis, line 79:
```suggestion
        .arg_from_usage("-c, --config-name=[CONFIG_NAME] 'Specifies a config name, required only in non-interactive mode.'")
```

avm/src/main.rs
```diff
@@ -154,19 +160,67 @@ fn main() {
...
+              None,
+              None,
+              app_name,
+              config_name,
+              non_interactive,
+            )
+            .await;
+          }
+          ("terminate", _) => {
+            authenticate(false).await;
```
dfellis, line 190: Same `TODO` comment here.

anycloud/cli/src/main.rs
aguillenv: I'll go with the latest alternative. Do not see this two being too different in the future

anycloud/cli/src/main.rs
aguillenv: Oh yes, forgot to implement it actually, will add this implementation in this PR

avm/src/main.rs
aguillenv: Implemented

anycloud/cli/src/main.rs
```diff
@@ -61,40 +71,82 @@ pub async fn main() {
...
   match matches.subcommand() {
     ("new", Some(matches)) => {
-      let dockerfile_b64 = get_dockerfile_b64().await;
-      let app_tar_gz_b64 = get_app_tar_gz_b64(true).await;
-      let env_b64 = match matches.value_of("env-file") {
-        Some(env_file) => Some(get_env_file_b64(env_file.to_string()).await),
-        None => None,
+      let non_interactive: bool = match matches.values_of("NON_INTERACTIVE") {
+        Some(_) => true,
+        None => false,
```
cdmistman, line 79: nit: this can be
```suggestion
      let non_interactive = matches.values_of("NON_INTERACTIVE").is_ok();
```

anycloud/cli/src/main.rs
```diff
@@ -61,40 +71,82 @@ pub async fn main() {
...
       };
```
cdmistman, line 80: nit: can be this:
```suggestion
      let non_interactive: bool = matches.values_of("NON_INTERACTIVE").is_ok();
```

anycloud/cli/src/main.rs
```diff
@@ -61,40 +71,82 @@ pub async fn main() {
...
-    ("upgrade", Some(matches)) => {
-      let dockerfile_b64 = get_dockerfile_b64().await;
-      let app_tar_gz_b64 = get_app_tar_gz_b64(true).await;
-      let env_b64 = match matches.value_of("env-file") {
-        Some(env_file) => Some(get_env_file_b64(env_file.to_string()).await),
+    ("terminate", Some(matches)) => {
+      let non_interactive: bool = match matches.values_of("NON_INTERACTIVE") {
+        Some(_) => true,
+        None => false,
+      };
```
cdmistman, line 106: same as above

anycloud/cli/src/main.rs
```diff
@@ -61,40 +71,82 @@ pub async fn main() {
...
+      authenticate(non_interactive).await;
+      let app_name = match matches.value_of("app-name") {
+        Some(name) => Some(name.to_string()),
         None => None,
       };
-      deploy::upgrade(
+      let config_name = match matches.value_of("config-name") {
+        Some(name) => Some(name.to_string()),
+        None => None,
+      };
```
cdmistman, line 115: nit: these can be `matches.value_of(...).map(String::from)`

anycloud/cli/src/main.rs
```diff
@@ -61,40 +71,82 @@ pub async fn main() {
...
+      deploy::terminate(app_name, config_name, non_interactive).await
+    }
+    ("upgrade", Some(matches)) => {
+      let non_interactive: bool = match matches.values_of("NON_INTERACTIVE") {
+        Some(_) => true,
+        None => false,
+      };
```
cdmistman, line 122: same as above

anycloud/cli/src/main.rs
```diff
@@ -117,6 +170,51 @@ pub async fn main() {
...
+    None => None,
+  };
+  let app_name = match app_name {
+    Some(name) => Some(name.to_string()),
+    None => None,
+  };
+  let config_name = match config_name {
+    Some(name) => Some(name.to_string()),
+    None => None,
+  };
```
cdmistman, line 210: same `map(String::from)`

avm/src/main.rs
```diff
@@ -154,19 +160,79 @@ fn main() {
         }
       }
       ("deploy", Some(sub_matches)) => {
-        authenticate().await;
         match sub_matches.subcommand() {
           ("new", Some(matches)) => {
+            let non_interactive: bool = match matches.values_of("NON_INTERACTIVE") {
+              Some(_) => true,
+              None => false,
+            };
```
cdmistman, line 168: nit: `.is_ok()`

avm/src/main.rs
```diff
@@ -154,19 +160,79 @@ fn main() {
...
             let agz_file = matches.value_of("AGZ_FILE").unwrap();
-            deploy::new(get_agz_b64(agz_file), None, None).await;
+            let app_name = match matches.value_of("app-name") {
+              Some(name) => Some(name.to_string()),
+              None => None,
+            };
+            let config_name = match matches.value_of("config-name") {
+              Some(name) => Some(name.to_string()),
+              None => None,
+            };
```
cdmistman, line 178: nits: `map(String::from)`

avm/src/main.rs
```diff
@@ -154,19 +160,79 @@ fn main() {
...
+              config_name,
+              non_interactive,
+            )
+            .await;
+          }
+          ("terminate", Some(matches)) => {
+            let non_interactive: bool = match matches.values_of("NON_INTERACTIVE") {
+              Some(_) => true,
+              None => false,
+            };
```
cdmistman, line 193: nit: `is_ok`

avm/src/main.rs
```diff
@@ -154,19 +160,79 @@ fn main() {
...
+            authenticate(non_interactive).await;
+            let app_name = match matches.value_of("app-name") {
+              Some(name) => Some(name.to_string()),
+              None => None,
+            };
+            let config_name = match matches.value_of("config-name") {
+              Some(name) => Some(name.to_string()),
+              None => None,
+            };
```
cdmistman, line 202: nits: `map(String::from)`

avm/src/main.rs
```diff
@@ -154,19 +160,79 @@ fn main() {
...
+            deploy::terminate(app_name, config_name, non_interactive).await
           }
-          ("terminate", _) => deploy::terminate().await,
           ("upgrade", Some(matches)) => {
+            let non_interactive: bool = match matches.values_of("NON_INTERACTIVE") {
+              Some(_) => true,
+              None => false,
+            };
```
cdmistman, line 209: nit: `is_ok`

avm/src/main.rs
```diff
@@ -154,19 +160,79 @@ fn main() {
...
             let agz_file = matches.value_of("AGZ_FILE").unwrap();
-            deploy::upgrade(get_agz_b64(agz_file), None, None).await;
+            let app_name = match matches.value_of("app-name") {
+              Some(name) => Some(name.to_string()),
+              None => None,
+            };
+            let config_name = match matches.value_of("config-name") {
+              Some(name) => Some(name.to_string()),
+              None => None,
+            };
```
cdmistman, line 219: nits: `map(String::from)`

anycloud/cli/src/main.rs
aguillenv: Of course!!! It would be `is_some()`, though

cdmistman: ah, you're right - was still waking up :sweat_smile:

cdmistman (approved):

dfellis (approved):

PR #567, opened 2021-06-10
Micro instances

aguillenv:
AWS [free tier](https://aws.amazon.com/ec2/pricing/) - t2.micro - From new to container running ~3min

image

image

Azure [free tier](https://azure.microsoft.com/en-us/free/) - Standard_B1s - From new to container running ~6min

image

image

GCP [free tier](https://cloud.google.com/free/docs/gcp-free-tier/#compute) - From new to container running ~10min

image

image

(GCP has the smaller vm type free tier)


cdmistman (approved):

avm/src/daemon/ctrl.rs
```diff
@@ -705,6 +705,32 @@ impl ControlPort {
...
+        Err(_) => {
+          results.push(false);
+        }
+        Ok(res) => {
+          results.push(res.status().as_u16() == 200);
+        }
+      }
+    }
+    return results.iter().all(|e| *e);
+  }
```
dfellis, line 732: This is kinda going in the opposite direction that we're trying to go. The cluster should be able to use its own state tracking to determine if it's up or down instead of asking the deploy service this. Why can't you use the `self.vms_up` HashMap and just return true if all of the booleans are true?

aguillenv: I'm not asking to deploy service, I'm just checking `clusterHealth` through control port. I am not using `self.vms_up` since that vector is filled within the loop that I'm short circuiting to avoid unnecessary leader and `/stats` logic. But I guess I could add my check later after the `update_vms`method is called to use the `vms_up`.
https://github.com/alantech/alan/pull/567/files#diff-583f2380a4d780cb04bd93e3492a6f2828052a50e23460f3a2c8d89cf3ade79fR303-R311

dfellis: Oh, sorry. Still waking up. I guess I didn't even consider that since the `vms_up` hashmap already exists and does roughly the same thing.

But if you short-circuit out the code that fills that, you'll break other parts of the daemon (how it knows which nodes can have data written to it, or which nodes should be included in the load balancer, etc).

aguillenv: Oh yes I might break some of that logic, let me move this check then

avm/src/daemon/ctrl.rs
```diff
@@ -706,29 +706,7 @@ impl ControlPort {
...
-        Err(_) => {
-          results.push(false);
-        }
-        Ok(res) => {
-          results.push(res.status().as_u16() == 200);
-        }
-      }
-    }
-    return results.iter().all(|e| *e);
+    return self.vms_up.values().all(|s| *s);
```
dfellis, line 709: This function no longer needs to be `async` with this change.

avm/src/daemon/daemon.rs
```diff
@@ -339,7 +334,8 @@ pub async fn start(is_local_anycloud_app: bool, local_agz_b64: Option<String>) {
                 .unwrap_or("<None>".to_string())
             );
           }
-          if stats.len() >= 4 {
+          // Do not send stats until cluster is up. Otherwise, will have scaling issues.
+          if stats.len() >= 4 && control_port.is_cluster_up().await {
```
dfellis, line 338: And you no longer need to `await` here once it's no longer an `async` function.

cdmistman (approved):

avm/src/daemon/daemon.rs
```diff
@@ -318,10 +318,15 @@ pub async fn start(is_local_anycloud_app: bool, local_agz_b64: Option<String>) {
           if control_port.is_leader() {
             // TODO: Should we keep these leader announcements in the stdout logging?
             println!("I am leader!");
-            match get_v1_stats().await {
-              Ok(s) => stats.push(s),
-              Err(err) => error!(NoStats, "{}", err).await,
-            };
+            // Do not collect stats until cluster is up. Otherwise, will have scaling issues.
+            if control_port.is_cluster_up() {
```
dfellis, line 322: Thinking about this after some sleep. There is a bit of a problem with this logic. It's true that during initial start or an upgrade some of the nodes won't be up and this will properly limit stats collection during that time, but if a node goes down unexpectedly it will also block stats collection and that may be very bad for keeping a live system from failing due to high load (since stats about its load will be stopped).

We also don't yet have logic in the deploy service to detect and remove/redeploy dead nodes, so any cluster that got into this state wouldn't get out (this latter point should be solved soon, but this change here will exacerbate it).

Distinguishing these two is going to be tough, so I won't block this for now, but I think we need to make an issue to track solving this problem. Perhaps a simple flag put into the DNS TXT record on "first time startup" or not, but there would also still need to be some logic there to handle "never actually started" so I don't think that's it, either.

aguillenv: I'm reviewing the code now and the `stats` are only being collected by the leader for itself, right? Probably this check would make sense just ensuring that I'm the leader and collect stats if I'm up, does not matter if any other random node is down we do not affect this logic. What do you think?

dfellis: That's true, but also was intended to be changed in the future. We are relying on an assumption that `request volume >> load per request` that works for a lot of "webby" things, but not "hard compute" things, and we wanted to have the cluster also share their stats with the leader so a snapshot of the whole cluster is shoved up to the deploy service, but that hasn't been done yet.

dfellis: It might be the right choice for now, though, and we revisit when we have whole-cluster stats tracking?

aguillenv: Even if the `get_stats` method collect the data from all up nodes and not only from the leader, we would still be only interested in call this method from the leader, so we will interested on do it when the leader is up, so the logic might be the same. But well that just be true once we revisit the logic as you said. Will make the change for now.

depombo (approved):

PR #568, opened 2021-06-11
Add lines length check

aguillenv:
@cdmistman Please take a look and let me know if this change makes sense or do you think is related to another part of the flow.

The error is the following when executing `new` or `terminate`:
```rs
thread 'main' panicked at 'attempt to subtract with overflow', /Users/alejandro/Documents/projects/alantech/alan/anycloud/cli/src/deploy/mod.rs:1321:44
```

### DIFF
--- a/anycloud/cli/src/deploy/mod.rs
+++ b/anycloud/cli/src/deploy/mod.rs
@@ -1316,15 +1316,15 @@ where
     // update the spinner and lines above the spinner
     new_lines
       .into_iter()
-      .filter(|line| !line.is_empty())
-      .for_each(|line| {
+      .filter(|new_line| !new_line.is_empty())
+      .for_each(|new_line| {
         if lines.len() > 0 {
           if let Some(last_line) = lines.get(lines.len() - 1) {
             sp.println(last_line);
           }
         }
-        sp.set_message(line);
-        lines.push(line.to_string());
+        sp.set_message(new_line);
+        lines.push(new_line.to_string());
       });
     tokio::time::sleep(match sleep_override.take() {
       None => DEFAULT_SLEEP_DURATION,
--- a/anycloud/cli/src/deploy/mod.rs
+++ b/anycloud/cli/src/deploy/mod.rs
@@ -1296,7 +1296,7 @@ where
       Ok(logs) => logs,
       Err(err) => {
         if let Some(last_line) = lines.get(lines.len() - 1) {
-          sp.prinqtln(last_line);
+          sp.println(last_line);
         }
         return match err {
           PostV1Error::Timeout => REQUEST_TIMEOUT.to_string(),
--- a/anycloud/cli/src/deploy/mod.rs
+++ b/anycloud/cli/src/deploy/mod.rs
@@ -1296,7 +1296,7 @@ where
       Ok(logs) => logs,
       Err(err) => {
         if let Some(last_line) = lines.get(lines.len() - 1) {
-          sp.println(last_line);
+          sp.prinqtln(last_line);
         }
         return match err {
           PostV1Error::Timeout => REQUEST_TIMEOUT.to_string(),
@@ -1313,17 +1313,19 @@ where
     // it's ok to leave out the newline chars, since `sp.println` will insert
     // those for us
     let new_lines = logs.split("\n").skip(lines.len()).collect::<Vec<_>>();
-    // print latest line if any
-    if lines.len() > 0 {
-      if let Some(last_line) = lines.get(lines.len() - 1) {
-        sp.println(last_line);
-      }
-    }
     // update the spinner and lines above the spinner
     new_lines
       .into_iter()
       .filter(|new_line| !new_line.is_empty())
       .for_each(|new_line| {
+        // print latest line if any.
+        // Will not print multiple times the same line since here we are adding a new one
+        // If no new lines, this iter will not execute and will not duplicate lines as if we put this check outside
+        if lines.len() > 0 {
+          if let Some(last_line) = lines.get(lines.len() - 1) {
+            sp.println(last_line);
+          }
+        };
         sp.set_message(new_line);
         lines.push(new_line.to_string());
       });
--- a/anycloud/cli/src/deploy/mod.rs
+++ b/anycloud/cli/src/deploy/mod.rs
@@ -1313,16 +1313,17 @@ where
     // it's ok to leave out the newline chars, since `sp.println` will insert
     // those for us
     let new_lines = logs.split("\n").skip(lines.len()).collect::<Vec<_>>();
+    // print latest line if any
+    if lines.len() > 0 {
+      if let Some(last_line) = lines.get(lines.len() - 1) {
+        sp.println(last_line);
+      }
+    }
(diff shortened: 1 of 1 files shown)


cdmistman (approved):
HOW did this not trigger during my testing????

dfellis (changes requested):
This fixes the crash, but doesn't look like it behaves correctly?

If the old `lines.len()` is `0`, it should be printing *all* of the new lines?

dfellis:
> HOW did this not trigger during my testing????

I suspect because you were reusing the same service name all the time? If this bug was introduced near the end,  then you might not have had a totally blank input during your testing.

aguillenv:
> This fixes the crash, but doesn't look like it behaves correctly?
> 
> If the old `lines.len()` is `0`, it should be printing _all_ of the new lines?

maybe @cdmistman could confirm, but what I understand is that if `lines` has something in it it should print the last one, if it's empty no need to do anything. The new lines are still being printed.

dfellis:
> > This fixes the crash, but doesn't look like it behaves correctly?
> > If the old `lines.len()` is `0`, it should be printing _all_ of the new lines?
> 
> maybe @cdmistman could confirm, but what I understand is that if `lines` has something in it it should print the last one, if it's empty no need to do anything. The new lines are still being printed.

So `lines` is the *old* set of lines and `new_lines` is the new set. The iterator calls each of the lines from `new_lines` just `line` confusingly.

Basically I stand behind my block here.

cdmistman:
> > HOW did this not trigger during my testing????
> 
> I suspect because you were reusing the same service name all the time? If this bug was introduced near the end, then you might not have had a totally blank input during your testing.

I used the same service name for some testing but not every time I tested (some times I chose the same name, I also sometimes didn't come back until >10 mins later, when the logs expired).

Regardless, `lines` *always* starts with `len() == 0`, so the first time the `/logs` endpoint returns *a single, non-empty line* during the lifetime of the CLI app should trigger a panic.

> If the old `lines.len()` is `0`, it should be printing all of the new lines?

lines is the vec of all the lines we've received. the vec doesn't actually do much besides storing them so we can `println` them later instead of just leaving them as the message of the spinner. Honestly, `lines` could be replaced with a single `Option` that we `take()` to print and then reassign with `line` in the `for_each` closure

cdmistman:
> > > This fixes the crash, but doesn't look like it behaves correctly?
> > > If the old `lines.len()` is `0`, it should be printing _all_ of the new lines?
> > 
> > 
> > maybe @cdmistman could confirm, but what I understand is that if `lines` has something in it it should print the last one, if it's empty no need to do anything. The new lines are still being printed.
> 
> So `lines` is the _old_ set of lines and `new_lines` is the new set. The iterator calls each of the lines from `new_lines` just `line` confusingly.
> 
> Basically I stand behind my block here.

yes, `line` in the closure is `new_line`. the only behavior this pr changes is when `len` is 0, which is before `/logs` has returned anything.

anycloud/cli/src/deploy/mod.rs
```diff
@@ -1318,8 +1318,10 @@ where
...
       .filter(|line| !line.is_empty())
       .for_each(|line| {
-        if let Some(last_line) = lines.get(lines.len() - 1) {
-          sp.println(last_line);
+        if lines.len() > 0 {
+          if let Some(last_line) = lines.get(lines.len() - 1) {
+            sp.println(last_line);
+          }
         }
         sp.set_message(line);
```
dfellis, line 1326: If there are multiple new entries in `new_lines` this is gonna print `last_line` multiple times. I wouldn't mind if this fix as currently written was before the functional stuff instead of inside of it?

aguillenv: > > > This fixes the crash, but doesn't look like it behaves correctly?
> > > If the old `lines.len()` is `0`, it should be printing _all_ of the new lines?
> > 
> > 
> > maybe @cdmistman could confirm, but what I understand is that if `lines` has something in it it should print the last one, if it's empty no need to do anything. The new lines are still being printed.
> 
> So `lines` is the _old_ set of lines and `new_lines` is the new set. The iterator calls each of the lines from `new_lines` just `line` confusingly.
> 
> Basically I stand behind my block here.

I've renamed it for clarity.

cdmistman: it will print `last_line` multiple times, but `last_line` is going to be different every time since we push to `lines` at the end of the closure. it has to be here, inside the closure to ensure that we override the spinner's message properly

anycloud/cli/src/deploy/mod.rs
```diff
@@ -1316,13 +1316,15 @@ where
...
+        if lines.len() > 0 {
+          if let Some(last_line) = lines.get(lines.len() - 1) {
+            sp.println(last_line);
+          }
         }
-        sp.set_message(line);
-        lines.push(line.to_string());
+        sp.set_message(new_line);
+        lines.push(new_line.to_string());
       });
```
dfellis: The issue with this is if `new_line` has multiple new lines compared to the old `lines` it will be printed multiple times.

It would be better to break this out of the `for_each` or somehow make sure it's only run on the first iteration of that loop?

anycloud/cli/src/deploy/mod.rs
aguillenv: > If there are multiple new entries in `new_lines` this is gonna print `last_line` multiple times. I wouldn't mind if this fix as currently written was before the functional stuff instead of inside of it?

makes sense, I did not catch that. Just extracted it.

dfellis (approved):

anycloud/cli/src/deploy/mod.rs
```diff
@@ -1313,16 +1313,19 @@ where
...
-      .for_each(|line| {
-        if let Some(last_line) = lines.get(lines.len() - 1) {
-          sp.println(last_line);
-        }
-        sp.set_message(line);
-        lines.push(line.to_string());
+      .filter(|new_line| !new_line.is_empty())
+      .for_each(|new_line| {
+        sp.set_message(new_line);
+        lines.push(new_line.to_string());
```
cdmistman, line 1328: this changes the behavior and does not do what is expected.

Say `lines` is empty and `new_lines` contains this:
```
"hello"
"hey"
"hi"
```

the result after the `for_each` is done will be:
```
<spinner> hi
```
when it should be
```
hello
hey
<spinner> hi
```

aguillenv: True, with this change the behaviour is the following that is not expected:
image

Restored it

aguillenv:
Just understood it and tested it. I have restored it to my initial fix since is the expected behaviour. Nevertheless I've added this comment to clarify: https://github.com/alantech/alan/pull/568/files#diff-cdb2ed790a8a8d7ee75b3740ed4c96f3a9bd1c20df6f27a064c8062b3c8fb906R1321-R1323

cdmistman (approved):

anycloud/cli/src/deploy/mod.rs
aguillenv: With the initial fix works as expected:
image

PR #570, opened 2021-06-13
Datastore automatic key rebalancing

dfellis:
Confirmed to work with a [new loadtest service](https://github.com/alantech/loadtest/tree/kv-tester):

Screenshot from 2021-06-15 17-15-23

Data is not lost as the load test pushes the cluster size up and up. It used to drop data during the scaling:

Screenshot from 2021-06-15 17-26-00


avm/src/daemon/ctrl.rs
```diff
@@ -611,10 +627,21 @@ impl ControlPort {
   }
 
   pub async fn update_vms(self: &mut ControlPort, self_ip: &str, vms: Vec<VMMetadata>) {
-    let ips = vms
+    let ips: Vec<String> = vms
       .iter()
       .map(|vm| vm.private_ip_addr.to_string())
       .collect();
+    // Detect changes and exit early if nothing has changed
+    let changed = if ips.len() != self.vms.len() {
```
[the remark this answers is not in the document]
dfellis: That's what the `else if` branch is for :)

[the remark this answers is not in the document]
aguillenv: yes, I got that. I've just realized that the operator is `!=` and not `==`!! my bad

dfellis:
Oh, I still need to remove the `panic!`s. They aren't actually triggered, but we should be logging these unexpected paths instead of crashing the AVM on them. Will do that in the morning.

aguillenv (approved):

avm/src/daemon/ctrl.rs
```diff
@@ -517,6 +541,26 @@ async fn dssetv_inner(req: Request<Body>) -> DaemonResult<()> {
...
+    Ok(keys) => Ok(Response::builder().status(200).body(keys.into()).unwrap()),
+    Err(err) => {
+      // TODO: What error message here? Also should this also be a valid HM out of here?
+      eprintln!("{:?}", err);
+      Ok(Response::builder().status(500).body("fail".into()).unwrap())
+    }
+  }
+}
+
+async fn keys_inner() -> DaemonResult<String> {
```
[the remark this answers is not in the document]
dfellis: That's right. I don't even need to break this into another function.

avm/src/daemon/ctrl.rs
```diff
@@ -424,6 +425,33 @@ async fn dsgetv_inner(req: Request<Body>) -> DaemonResult<Arc<HandlerMemory>> {
...
+      // TODO: What error message here? Also should this also be a valid HM out of here?
+      eprintln!("{:?}", err);
+      Ok(Response::builder().status(500).body("fail".into()).unwrap())
+    }
+  }
+}
+
+async fn dsgetr_inner(req: Request<Body>) -> DaemonResult<Arc<HandlerMemory>> {
+  let bytes = body::to_bytes(req.into_body()).await?;
+  let body: DSGet = serde_json::from_slice(&bytes).unwrap();
```
cdmistman, line 445: i think it makes sense to not `unwrap` the deserialization and instead bubble up the error with `?` - a malformed JSON structure is a valid error that shouldn't cause a panic

avm/src/daemon/ctrl.rs
```diff
@@ -441,7 +469,6 @@ async fn handle_dshas(req: Request<Body>) -> Result<Response<Body>, Infallible>
 }
 
 async fn dshas_inner(req: Request<Body>) -> DaemonResult<bool> {
-  // TODO: For now assume this was directed at the right node, later on add auto-forwarding logic
   let bytes = body::to_bytes(req.into_body()).await?;
   let body: DSGet = serde_json::from_slice(&bytes).unwrap();
```
cdmistman, line 473: same issue as above with panicking on invalid JSON messages

avm/src/daemon/ctrl.rs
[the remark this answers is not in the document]
cdmistman: in that case, `handle_keys` doesn't need to be `async` too right?

avm/src/daemon/ctrl.rs
```diff
@@ -611,10 +655,21 @@ impl ControlPort {
...
       .map(|vm| vm.private_ip_addr.to_string())
       .collect();
+    // Detect changes and exit early if nothing has changed
+    let changed = if ips.len() != self.vms.len() {
+      true
+    } else if ips.iter().any(|ip| !self.vms.contains_key(ip)) {
+      true
+    } else {
+      false
+    };
```
cdmistman, line 669: nit: all of this can be
```suggestion
    let changed = ips.len() != self.vms.len() || ips.iter().any(|ip| !self.vms.contains_key(ip));
```

avm/src/daemon/ctrl.rs
```diff
@@ -902,4 +970,165 @@ impl ControlPort {
...
+    (
+      ip.to_string(),
+      self.dskeys_inner(ip).await.unwrap_or(Vec::new()),
+    )
+  }
+
+  async fn dskeys_inner(self: &ControlPort, ip: &str) -> DaemonResult<Vec<String>> {
+    let url = format!("https://{}:4142/datastore/keys", ip);
+    let req = Request::builder().method("GET").uri(url);
+    let cluster_secret = CLUSTER_SECRET.get().unwrap().clone().unwrap();
```
cdmistman, line 984: seeing so many of `CLUSTER_SECRET.get().unwrap().clone().unwrap()` makes me think maybe we should have convenience getters for our `OnceCell`/`Lazy` variables

avm/src/daemon/ctrl.rs
```diff
@@ -902,4 +970,165 @@ impl ControlPort {
...
+        Ok(res) => res,
+        Err(_) => panic!("TODO"),
+      };
+      let bytes = match hyper::body::to_bytes(res.body_mut()).await {
+        Ok(bytes) => bytes,
+        Err(_) => panic!("TODO"),
+      };
+      let pb = match protos::HandlerMemory::HandlerMemory::parse_from_bytes(&bytes) {
+        Ok(pb) => pb,
+        Err(_) => panic!("jelly"),
```
cdmistman, line 1102: nice

avm/src/daemon/ctrl.rs
```diff
@@ -902,4 +970,165 @@ impl ControlPort {
...
+      };
+      let hm = match HandlerMemory::from_pb(&pb) {
+        Ok(hm) => hm,
+        Err(_) => panic!("hot topic"),
```
cdmistman, line 1106: nice

avm/src/daemon/ctrl.rs
```diff
@@ -902,4 +970,165 @@ impl ControlPort {
...
+        })
+        .to_string(),
+      ));
+      let req_obj = match req_obj {
+        Ok(req_obj) => req_obj,
+        Err(_) => panic!("Should be impossible"),
+      };
+      let mut res = match self.client.request(req_obj).await {
+        Ok(res) => res,
+        Err(_) => panic!("TODO"),
```
cdmistman, line 1094: there's also `todo!()` which basically does this ;)

avm/src/daemon/ctrl.rs
```diff
@@ -902,4 +970,165 @@ impl ControlPort {
...
+      let req_obj = req.body(Body::from(
+        json!(DSGet {
+          nskey: key.to_string()
+        })
+        .to_string(),
+      ));
+      let req_obj = match req_obj {
+        Ok(req_obj) => req_obj,
+        Err(_) => panic!("Should be impossible"),
+      };
```
cdmistman, line 1124: nit: could be `.expect("Should be impossible")`

avm/src/daemon/ctrl.rs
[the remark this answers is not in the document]
dfellis: Also right!

avm/src/daemon/ctrl.rs
dfellis: That was copy-pasted, so I'll fix that in the others. You're right on this.

avm/src/daemon/ctrl.rs
dfellis: :facepalm:

aguillenv (approved):

cdmistman (approved):

PR #571, opened 2021-06-14
Handle polling on upgrade

aguillenv:


anycloud/cli/src/deploy/mod.rs
```diff
@@ -1106,16 +1106,26 @@ pub async fn upgrade(
   let resp = post_v1("upgrade", body).await;
   let res = match resp {
     Ok(_) => {
+      // Check every 10s over 5 min if app already start upgrading
```
cdmistman, line 1109: is 5 mins too long to wait? The deploy service should be initializing the upgrade pretty quickly right?

aguillenv: I mean, this would be the maximum amount of time we wait until we are sure we can start polling. Once we are sure, we `break`. Probably we could reduce it but to something considerable like 2 or 3 minutes. It is true that should be a quick action but it is not immediate, so I was having troubles since it was not polling for me since it checked, the size haven't change so it assumed it was ready and finish the `clap` interaction.

cdmistman: gotcha - I had assumed that the network latency would be enough :/

cdmistman (approved):

anycloud/cli/src/deploy/mod.rs
```diff
@@ -1106,16 +1106,26 @@ pub async fn upgrade(
...
+      let mut counter: u8 = 0;
+      while counter < 30 {
+        let is_upgrading = get_apps(true)
+          .await
+          .into_iter()
+          .find(|app| &app.id == cluster_id)
+          .map(|app| app.status == "down")
+          .unwrap_or(false);
```
dfellis, line 1117: If it fails to unwrap, wouldn't that mean it didn't even find the relevant app ID and it should be considered to be still deploying/upgrading?

That is, shouldn't the default be `true`?

aguillenv: Hmm I don't think so because we are trying to be sure that the app exists and it is in a "down" state (meaning it started upgrading) in order to start polling and get the corresponding logs. On `upgrade` your suggested scenario would not happen and on a deploying case we don't want to start polling for `upgrade` if the app is being deployed.

dfellis: What can cause the `unwrap` to fail and need a default? The only thing I can think of is the `.find` fails to find a result and then the `.map` similarly fails because there isn't anything to `map` at all. That means that it somehow didn't get a response from the server that makes sense.

It normally would be impossible, but what if the deploy service itself went down right when someone requested an upgrade and the Redis cache of managed clusters hasn't been re-updated yet and it really *can't* find the app ID in question? It still seems like you'd want to *wait* on that loop until it at least finds the cluster again?

aguillenv: Yes, exactly, I want to stay in the loop (up to 5 min) in order to give the opportunity to start an accurate polling. If it fails to `unwrap`, we default `is_upgrading = false` so I do not break the loop. I could add some message after the 5 min saying something like `Could not find information related to your cluster, try again later.` instead of start the polling.

dfellis: Ohh, I misunderstood this loop. I thought the process was going to exit when the loop is done, but now I see that the `poll` below is *outside* of this loop, and this is a pre-check for when the logs are ready to query.

aguillenv: Yes!

aguillenv: I've added the return of a custom message instead of polling if after the 5 min nothing changed.

dfellis (approved):

PR #573, opened 2021-06-14
[lnn] interfaces

cdmistman:


bdd/spec/006_string_lnn_spec.sh
```diff
@@ -4,31 +4,26 @@ Describe "Strings"
...
 
-          emit stdout concat(toString(matches('foobar', 'fo.*')), \"\n\");
-          // matches('foobar', 'fo.*').print();
-          // print('foobar' ~ 'fo.*');
+          matches('foobar', 'fo.*').print();
+          print('foobar' ~ 'fo.*');
 
           // index('foobar', 'ba').print();
-          // print('foobar' @ 'ba');
+          // print('foobar' @ 'ra');
```
dfellis, line 20: I know this is still commented out, but is there a reason for choosing an invalid string here? To confirm the error message in the future, I guess?

compiler/src/lnntoamm/Expr.ts
```diff
@@ -385,15 +389,18 @@ class Call extends Expr {
     fns: Fn[],
     maybeClosure: VarDef | null,
     args: Ref[],
+    scope: Scope,
   ) {
+    // console.log('~~~ generating call ', ast, fns, maybeClosure, args, scope);
```
dfellis, line 394: Left behind debug log. Do you need this often or should it be deleted.

compiler/src/lnntoamm/Expr.ts
```diff
@@ -493,9 +497,9 @@ class Call extends Expr {
   */
   inline(amm: Output, kind: AssignKind, name: string, ty: Builtin) {
     const argTys = this.args.map(arg => arg.ty.instance());
-    const selected = this.fns.reverse().find(fn => fn.acceptsTypes(argTys)) || null;
+    const selected = Fn.select(this.fns.reverse(), argTys, this.scope) || [];
```
dfellis, line 500: Should the `reverse` be part of calling `Fn.select` or should this API have this explicitly exposed? (Are there any situations where you wouldn't reverse the list, basically, and therefore the `select` function should be taking care of this to DRY it out?)

compiler/src/lnntoamm/Operator.ts
```diff
@@ -51,8 +52,8 @@ class Operator {
       }
       throw new Error(`nope`);
     }
-    const tys = [arg1.ty, ...(arg2 ? [arg2.ty] : [])];
-    return this.fns.filter(fn => fn.acceptsTypes(tys));
+    const tys = [arg1, ...(arg2 ? [arg2] : [])];
+    return Fn.select(this.fns, tys, scope);
```
dfellis, line 56: I see for operators it's using the list as originally ordered. Is that correct?

compiler/src/lnntoamm/Types.ts
```diff
@@ -85,54 +103,85 @@ export class Builtin extends Type {
...
-    } if (ty instanceof OneOf || ty instanceof Generated) {
-      return ty.compatibleWithConstraint(this);
+    } else if (ty instanceof OneOf || ty instanceof Generated) {
+      return ty.compatibleWithConstraint(this, scope);
+    } else if (ty instanceof HasOperator) {
+      return Has.operator(ty, scope, this).length !== 0;
+    } else if (ty instanceof HasMethod) {
+      return Has.method(ty, scope, this).length !== 0;
+    } else if (ty instanceof HasField) {
+      return false;
```
dfellis, line 116: Should there be a `//TODO: Once structs are implemented fix this` here?

bdd/spec/006_string_lnn_spec.sh
cdmistman: it's to test the behavior of `index` to make sure that it doesn't false-positive on this input (i didn't have to do it in this PR, but i thought of it during my refactors and thought it'd be good to do)

compiler/src/lnntoamm/Types.ts
```diff
@@ -85,54 +103,85 @@ export class Builtin extends Type {
...
+    } else if (ty instanceof Interface) {
+      return ty.fields.length === 0 &&
```
dfellis, line 118: Same comment on the `fields` check here looking very temporary but not having an accompanying `TODO` comment.

compiler/src/lnntoamm/Expr.ts
cdmistman: i end up using it quite often

compiler/src/lnntoamm/Types.ts
```diff
@@ -85,54 +103,85 @@ export class Builtin extends Type {
...
-    if (ty instanceof OneOf || ty instanceof Generated) {
-      ty.constrain(this);
-    } else if (!this.compatibleWithConstraint(ty)) {
+  constrain(ty: Type, scope: Scope) {
+    if (ty instanceof OneOf || ty instanceof Generated || ty instanceof Interface) {
+      ty.constrain(this, scope);
+    } else if (ty instanceof HasOperator) {
+      if (Has.operator(ty, scope, this).length === 0) {
+        throw new Error(`type ${this.name} does not have operator ${ty.name}`);
+      }
```
dfellis, line 132: Nit. You can combine the nested conditional up without impacting the logic.

```suggestion
    } else if (ty instanceof HasOperator && Has.operator(ty, scope, this).length === 0) {
      throw new Error(`type ${this.name} does not have operator ${ty.name}`);
```

compiler/src/lnntoamm/Types.ts
```diff
@@ -85,54 +103,85 @@ export class Builtin extends Type {
...
+    } else if (ty instanceof HasMethod) {
+      if (Has.method(ty, scope, this).length === 0) {
+        throw new Error(`type ${this.name} does not have method ${ty.name}(${ty.params.map(p => p === null ? this : p).map(ty => ty.name).join(', ')})`);
+      }
```
dfellis, line 136: Same here.

```suggestion
    } else if (ty instanceof HasMethod && Has.method(ty, scope, this).length === 0) {
      throw new Error(`type ${this.name} does not have method ${ty.name}(${ty.params.map(p => p === null ? this : p).map(ty => ty.name).join(', ')})`);
```

compiler/src/lnntoamm/Types.ts
```diff
@@ -85,54 +103,85 @@ export class Builtin extends Type {
...
 
   eq(that: Equalable): boolean {
-    return that instanceof Builtin && this.ammName === that.ammName;
+    if (that instanceof Builtin) {
+      return this.ammName === that.ammName;
+    } else if (that instanceof Interface) {
+      if (that instanceof Generated && that.delegate !== null) {
+        return this.eq(that.delegate);
+      }
+      return that.tempDelegate !== null && this.eq(that.tempDelegate);
```
dfellis, line 151: What exactly is the difference between `.delegate` and `.tempDelegate`?

compiler/src/lnntoamm/Expr.ts
cdmistman: actually, that `reverse` shouldn't be there - the last-defined functions are already at the end of the list, and we `pop` to get the function we use. The bdd tests probably didn't catch this since the only scenario where the reversed list has more than 1 element is when looking at the integer types (and all integers are compatible once compiled).

fixing!

compiler/src/lnntoamm/Operator.ts
cdmistman: yes - as mentioned above, the fn list shouldn't be reversed

compiler/src/lnntoamm/Types.ts
```diff
@@ -197,83 +252,426 @@ class Struct extends Type {
...
+    if (!isOpArray(ops)) {
+      return [];
+    }
+    // filter out ops that aren't the same fixity
+    ops = ops.filter(op => op.isPrefix === operator.isPrefix);
+    if (operator.isPrefix) {
+      return ops.filter(op => op.select(scope, operator.params[0] || ty) !== []);
+    } else {
+      return ops.filter(op => op.select(scope, operator.params[0] || ty, operator.params[1] || ty) !== []);
+	  }
```
dfellis, line 302: Nit: accidental tab character here, I think.

```suggestion
    }
```

compiler/src/lnntoamm/Types.ts
cdmistman: I actually can't collapse this once since `HasOperator extends HasMethod` and will get checked by `Has.method` on the next branch, but i can combine the `Has.method` conditional with `instanceof HasMethod`

compiler/src/lnntoamm/Types.ts
```diff
@@ -197,83 +252,426 @@ class Struct extends Type {
...
+    return ty.compatibleWithConstraint(this, scope);
+  }
+
+  // convenience for `Type.hasX(...).constrain(ty)`
+  constrain(to: Type, scope: Scope) {
+    to.constrain(this, scope);
+  }
+
+  eq(that: Equalable): boolean {
+    return that instanceof Has && that.name === this.name;
```
dfellis, line 320: What if there's a field and a method of the same name? Like `@std/http`'s `Request` `body`/`body(string)`?

compiler/src/lnntoamm/Types.ts
cdmistman: delegates are only set once, while `tempDelegates` can be reset. I think only `Interface` is necessary (and we can get rid of `Generated`) but I want to wait until untyped fn parameters are implemented before making that decision (since i'm not 100% certain that they can be represented with just an `Interface`)

compiler/src/lnntoamm/Types.ts
cdmistman: should get caught by the `HasField.eq` and `HasMethod.eq` implementations - this is only for a little bit of delegation (the `instanceof Has` is redundant but it lets TS recognize that `that.name` is a valid symbol)

compiler/src/lnntoamm/Types.ts
cdmistman: er, not here - builtins never have user-accessible fields right?

compiler/src/lnntoamm/Types.ts
```diff
@@ -197,83 +252,426 @@ class Struct extends Type {
...
+
+      // check methods
+      return this.methods.every(m => Has.method(m, scope, ty)) &&
+        this.operators.every(o => Has.operator(o, scope, ty));
+      // check operators
+    } else if (ty instanceof Interface) {
+      // ensure `ty ⊆ this`
+      return ty.fields.every(field => this.fields.find(f => f.eq(field))) &&
+        ty.methods.every(method => this.methods.find(m => m.eq(method))) &&
+        ty.operators.every(operator => this.operators.find(o => o.eq(operator)));
```
dfellis, line 532: Nonblocking: It may make sense to change these from arrays to hashmaps for better lookup performance (this is currently `O(n*m)`) Since you can get an iterable array of the values with [`Object.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values) cheaply, it doesn't seem like it would impact any of the other uses, either?

compiler/src/lnntoamm/Types.ts
```diff
@@ -197,83 +252,426 @@ class Struct extends Type {
...
   eq(that: Equalable): boolean {
-    // TODO: gets more complicated later
-    return that instanceof Interface && this === that;
+    if (that instanceof Generated) {
+      return that.eq(this);
+    } else if (that instanceof Interface) {
+      // FIXME: this is technically wrong, but there's no other way
+      // to get the current generic types working without depending
+      // on `eq` returning this. Ideally, we would be checking to
+      // make sure all of the constraints match
```
dfellis, line 575: I don't quite follow this comment because generics haven't been implemented yet?

compiler/src/lnntoamm/Types.ts
```diff
@@ -197,83 +252,426 @@ class Struct extends Type {
...
+    return dup;
   }
 }
 
-class Generated extends Type {
-  private delegate: Type | null
-  private tempDelegate: Type | null
+// technically, generated types are a kind of interface - we just get to build up
+// the interface through type constraints instead of through explicit requirements.
+// this'll make untyped fn parameters easier once they're implemented.
```
dfellis, line 622: # :100:

compiler/src/lnntoamm/Types.ts
```diff
@@ -197,83 +252,426 @@ class Struct extends Type {
...
   }
 
   breakdown(): Builtin {
-    if (this.tempDelegate !== null) {
-      return this.tempDelegate.breakdown();
-    } else if (this.delegate !== null) {
+    if (this.delegate !== null) {
       return this.delegate.breakdown();
-    } else {
+    } else try { // try? more like DRY amiright
```
dfellis, line 634: I had no idea this was possible.

compiler/src/lnntoamm/Types.ts
```diff
@@ -300,28 +721,36 @@ class Generated extends Type {
...
+    } else if (to instanceof Has) {
+      TODO("i'm not sure");
     } else {
       this.tempDelegate = to;
     }
   }
 
   resetTemp() {
-    if (this.tempDelegate !== null) {
+    if (super.tempDelegate !== null) {
```
dfellis, line 729: Is this change right? If so, why doesn't the `super` also get its `tempDelegate` removed?

compiler/src/lnntoamm/Types.ts
dfellis: Unfortunately no. `ExecRes`, `InitialReduce`, `KeyVal`, `InternalRequest`, and `InternalResponse` all have user-accessible fields but are also built-in types as some opcodes directly consume or produce them.

compiler/src/lnntoamm/Types.ts
dfellis: Ahh. Why does `HasOperator` extend `HasMethod`, btw? I wouldn't have expected there to be that much code reuse between the two?

compiler/src/lnntoamm/Types.ts
cdmistman: HA, SO

I found out that, apparently, JS's `super` only works on functions :) I don't know why, and MDN doesn't explicitly state this, but whenever you do `super.<field>`, it's actually `undefined`. This error just never triggered because it's impossible to have a `Generated`-typed `FnParam`

compiler/src/lnntoamm/Types.ts
dfellis: What do you think would be needed extra to represent them versus `Interface`s? It might make sense to just add that as a potentially explicit syntax to `Interface` as well to force them to be the same (and also make them both equally expressive, I would hate if certain code can only be written in explicit or implicit style only and not the other).

compiler/src/lnntoamm/Types.ts
cdmistman: I was actually banging my head against the wall for at least an hour because I was trying to access `super.tempDelegate` in the other methods and it was never working and I had no idea why and then I decided to try accessing with `this` and it worked and I just... *sigh*

compiler/src/lnntoamm/Types.ts
cdmistman: ah, the comment's wording is a little off. I mean the current way of doing generic fn parameters (all `Stringifiables` are the same)

compiler/src/lnntoamm/Types.ts
cdmistman: gotcha, i'll leave a TODO then

compiler/src/lnntoamm/Types.ts
cdmistman: for 2 reasons: first, that operators are a kind of methods and so it makes sense to me, but also because that way the operators inherit the `params` and `ret` fields from `HasMethod` and can inherit behavior around the 2. I'm open to changing it, but this is what made sense to me

compiler/src/lnntoamm/Types.ts
cdmistman: Well right now there would be issues with type inference - *at least* fn selection will break, so the PR for fixing that would be the earliest that I can consider getting rid of the `Generated` type. It does make sense, but I'll most likely have to generate constraints such as `variableType.constrain(Type.hasMethod(...))`, etc

compiler/src/lnntoamm/Types.ts
cdmistman: This is a relatively large refactor on the `Types.ts` file, so I'll leave it as a TODO and get to it later when optimizing the compiler

compiler/src/lnntoamm/Types.ts
dfellis: No, it's minor; I just probably wouldn't have done it that way, but I don't see any harm in it?

dfellis (approved):

aguillenv (approved):

PR #575, opened 2021-06-17
bdd test for bugs

cdmistman:


bdd/spec/016_type_spec.sh
```diff
@@ -61,4 +61,47 @@ Describe "Type detection"
...
+      "
+    }
+    BeforeAll before
+
+    after() {
+      cleanTemp
+    }
+    AfterAll after
+
+    OUTPUT="{\"success\":\"true\"}"
```
dfellis, line 93: This test doesn't actually output anything because it ignores the `Result<Response>`

cdmistman: ah right

dfellis (approved):

PR #578, opened 2021-06-21
[lnn] structs

cdmistman:


cdmistman:
WRT the last commit, it was discussed offline and we decided to implement alternative versions of `pusharr` that correct `pusharr`'s behavior, so the last commit will be undone by my next commit(s).

bdd/spec/017_custom_event_lnn_spec.sh
```diff
@@ -0,0 +1,148 @@
...
+          if val >= 10 {
+            emit exit 0;
+          } else {
+            emit loop val + 1 || 0;
+          }
+        }
+
+        on start {
+          emit loop 0;
+        }
```
dfellis, line 22: I am a little hesitant on replicating this test as-is. It is easy to produce an infinite loop with this pattern, so I was hoping to block it in a similar way the direct recursive function calling is detected and blocked.

bdd/spec/017_custom_event_lnn_spec.sh
```diff
@@ -0,0 +1,148 @@
...
+    End
+
+    It "runs agc"
+      Pending generics-and-arithmetic
+      When run test_agc
+      The output should eq "$OUTPUT"
+    End
+  End
+
+  # TODO: this
```
dfellis, line 57: Nit: delete this comment?

compiler/src/lnntoamm/Amm.ts
```diff
@@ -114,6 +114,18 @@ export default class Output {
...
   }
 
+  call(
+    opcodeName: string,
+    args: string[],
+  ) {
+    let line = this.indent;
+    line = line.concat(opcodeName, '(');
+    line = line.concat(args.join(', '));
+    line = line.concat(')');
```
dfellis, line 124: Wondering why this isn't like:

```suggestion
    const line = `${this.indent}${opcodeName}(${args.join(', ')})`;
```

bdd/spec/017_custom_event_lnn_spec.sh
cdmistman: but infinite loops using this pattern is safe right? It doesn't have the same issue as infinite loops because it doesn't block the runtime and it doesn't have the same issue as infinite recursion because it creates a new task for the event handler?

compiler/src/lnntoamm/Expr.ts
```diff
@@ -84,15 +87,31 @@ export default abstract class Expr {
...
+            expr = dec.ref();
+          }
+          // ensure that the value has the field
+          let fieldTy = Type.generate();
+          const hasField = Type.hasField(varName, fieldTy);
+          if (!expr.ty.compatibleWithConstraint(hasField, metadata.scope)) {
+            throw new Error(`cannot access ${varName} on type ${expr.ty.name} because it doesn't have that field`);
+          }
+          expr.ty.constrain(hasField, metadata.scope);
+          // TODO: better ast
```
dfellis, line 104: Can you explain what's wrong with the current AST here and what you'd like to change?

compiler/src/lnntoamm/Amm.ts
cdmistman: makes sense to use!

compiler/src/lnntoamm/Expr.ts
cdmistman: the current ast only provides the name of the field - it'd be better to also see what the field is accessing (so for example, `getThing().field` would result in the `AccessField`'s AST just being `field`, whereas it'd be nice to have the full part.) It'll take quite a bit more logic to get it working and I have to provide a better AST in operator calls anyways so it'll have to be a future PR. I'll update the comment with that info though

compiler/src/lnntoamm/Expr.ts
```diff
@@ -595,6 +643,94 @@ class Const extends Expr {
...
+  valTy: Type
+  fields: {[name: string]: Ref}
+
+  get ty(): Type {
+    return this.valTy;
+  }
+
+  /**
+   * NOTE: does NOT check to make sure that the fields
+   * are valid.
```
dfellis, line 656: This note doesn't seem valid? I see checking below? Unless you mean the raw constructor does not check, but `fromTypeLiteral` does, in which case maybe a matching comment for that one? I thought this was for the whole class at first glance.

cdmistman: Yeah, the checking is only done in `fromTypeLiteral` though - I was leaving this comment specifically for the constructor (I used JSDoc so that if somebody has to manually construct the `New`, they'll hopefully see it in their IDE as a reminder). I'll update the comment to note that

compiler/src/lnntoamm/Types.ts
```diff
@@ -89,9 +91,27 @@ export default abstract class Type implements Equalable {
...
+    try {
+      name = this.instance().name;
+    } catch (e) {
+      name = this.name;
+    }
+    if (name !== '') {
+      name = ` ${name}`;
+    }
+    throw new Error(`Type${name} does not have fields`);
+  }
```
dfellis, line 106: This is a little weird. Why does the name get a prefixed space on it? The error message could have been done by just adding a space before the dollar sign, for instance.

compiler/src/lnntoamm/Types.ts
```diff
@@ -123,6 +139,7 @@ export class Builtin extends Type {
             ty.methods.every(m => this.compatibleWithConstraint(m, scope)) &&
             ty.operators.every(o => this.compatibleWithConstraint(o, scope));
     } else {
+      console.log(ty);
```
dfellis, line 142: Delete the debug log? Or just keep it around since this `TODO` means anything hitting this path is going to hard fail, anyways?

compiler/src/lnntoamm/Types.ts
cdmistman: because otherwise there's a duplicated space, and that was bothering me in testing :sweat_smile:

compiler/src/lnntoamm/Types.ts
```diff
@@ -156,22 +173,61 @@ export class Builtin extends Type {
...
+  fieldIndices(): FieldIndices {
+    return TODO("determine if it's even worth keeping the Builtin class");
+  }
+
   instance(): Type {
     return this;
   }
 
+  isFixed(): boolean {
+    switch (this.name) {
```
dfellis, line 185: This *might* be an issue if type aliasing (eg the `type int = int64; type float = float64;` the current first stage compiler does) is still kept around.

compiler/src/lnntoamm/Types.ts
```diff
@@ -156,22 +173,61 @@ export class Builtin extends Type {
...
+    // TODO: should probably figure out non-opaque types but we'll see
+    // what happens with this class first (after fn selection fix)
+    switch (this.name) {
+      case 'void':
+        return 0;
+      default:
+        // FIXME: currently the AVM encodes every value (including 8-bit values)
+        // in 8 bytes, so we have to enforce that behavior here by assuming that
+        // the size is in factors of 8 bytes. Eventually, we'll have to fix this.
+        return 1;
```
dfellis, line 218: Or we just pretend 64-bits is sufficient for everything and don't. ;)

compiler/src/lnntoamm/Types.ts
cdmistman: I can delete it, but yeah `TODO` does throw an exception so the log is just to give some extra info

compiler/src/lnntoamm/Types.ts
```diff
@@ -182,6 +238,12 @@ class Struct extends Type {
     super(name, ast);
     this.args = args;
     this.fields = fields;
+    this.order = {};
+    let sizeTracker = 0;
+    for (let fieldName in this.fields) {
+      this.order[fieldName] = sizeTracker;
+      sizeTracker += this.fields[fieldName].size();
```
dfellis, line 245: Wouldn't this *technically* blow up on us if we had a `void` field?

```ln
type Foo
  bar: void,
  baz: string,
}
```

would probably do a *lot* of weird crap if you tried to access either field.

compiler/src/lnntoamm/Types.ts
cdmistman: yeah - I haven't implemented aliasing yet so I'm not worried about it yet. I think it can be done with a `class Alias extends Type` to avoid scoping issues and just delegates to an internal delegate on every call, but I want to figure out generics first

compiler/src/lnntoamm/Types.ts
cdmistman: true! will update comment

compiler/src/lnntoamm/Types.ts
cdmistman: I think it'll be UB to access `bar` but not `baz` - I'm not worried about managing this specific case yet since some things related to `void` might have to change once conditionals are implemented

compiler/src/lnntoamm/Types.ts
```diff
@@ -780,7 +887,11 @@ class OneOf extends Type {
   }
 
   instance(): Type {
-    return this.select().instance();
+    const selected =  this.select();
```
dfellis, line 890: Super nit, but:

```suggestion
    const selected = this.select();
```

compiler/src/lnntoamm/Types.ts
```diff
@@ -780,7 +887,11 @@ class OneOf extends Type {
...
+    if (selected === undefined) {
+      throw new Error(`uh whaaaaat`);
```
dfellis, line 892: More nit:

```suggestion
      throw new Error('uh whaaaaat');
```

js-runtime/index.js
```diff
@@ -740,6 +740,8 @@ module.exports = {
   // Array opcodes TODO more to come
   newarr:   size => new Array(), // Ignored because JS push doesn't behave as desired
   pusharr: (arr, val, size) => arr.push(val),
+  pushf: (arr, val, size) => arr.push(val),
+  pushv: (arr, val, size) => arr.push(val),
```
dfellis, line 744: Should these opcodes even have a size argument?

cdmistman: I'm not entirely sure, tbh. Won't `pushf` need it, at least? (to distinguish between the different-sized values, assuming we figure out associated problems). I can get rid of it in `pushv` though, I don't see a scenario where it's ever helpful

bdd/spec/017_custom_event_lnn_spec.sh
dfellis: That's why I first allowed it, but I'm still ambivalent about it. If you're trying to do a `setInterval` type thing, it would probably be better for us to add a language feature to properly expose `setInterval` and the Tokio equivalent instead of a quasi-hot-loop polling the current time and seeing if it should run or not?

Basically, it is similar to a `while` loop in that providing functions or language-level constructs to directly tackle the patterns the user wants instead gives the compiler and runtime the ability to be more efficient with what they're doing. This test was very much just a convenient way to test event emission from conditionals, but actually needing this pattern feels more like we're missing a better abstraction in the language that we should add, instead?

compiler/src/lnntoamm/Types.ts
dfellis: `throw new Error(`Type${name ? ` ${name}` : ''} does not have fields`);`

Would also work (you can nest template strings, fyi) though that doesn't buy you *too* much.

You could also have something like:

```js
const formalName = name === '' ? 'Type' : `Type ${name}`;
throw new Error(`${formalName} does not have fields`);
```

But this is all extremely nitpicky over mutating the name of the type with formatting data, y'know, in case we ever want to translate the error messages into a language that doesn't put spaces between words. ;)

compiler/src/lnntoamm/Types.ts
dfellis: Nah, just keep it.

bdd/spec/017_custom_event_lnn_spec.sh
cdmistman: Honestly, I think that this pattern for `setInterval` is the most intuitive for Alan since it's event-based. We *could* expose *yet more* opcodes for this, but this feels like a really natural pattern for event-based languages and should probably be considered idiomatic over eg `@std/seq`.

compiler/src/ammtoaga/depgraph.ts
```diff
@@ -557,7 +557,7 @@ export const opcodeParamMutabilities = {
   condfn: [false, null],
   pusharr: [true, false, false],
   pushf: [true, false, false],
```
dfellis, line 559: Shouldn't this one also be changed to 2 arg?

js-runtime/index.js
```diff
@@ -741,7 +741,7 @@ module.exports = {
   newarr:   size => new Array(), // Ignored because JS push doesn't behave as desired
   pusharr: (arr, val, size) => arr.push(val),
   pushf: (arr, val, size) => arr.push(val),
```
dfellis, line 743: Is there a reason `pushf` still needs the size?

bdd/spec/017_custom_event_lnn_spec.sh
dfellis: So I wasn't prescribing how the event registration should look (I don't think it should look like JS's `setInterval`, I was mostly pointing out that using a loop like this would either: 1) be very inefficient being a quasi-hot-loop or 2) would be very loose on when it actually runs by needing a `wait` call inside of the event-based loop.

Having this construct at the language level would probably be a safer bet, *especially* in the cluster-of-AVMs perspective when you might want the interval to only run *once* for the entire cluster every N minutes, while using this loop is going to have it run M times for the number of nodes in the cluster.

cdmistman: > 1) be very inefficient being a quasi-hot-loop or 2) would be very loose on when it actually runs by needing a wait call inside of the event-based loop.

While true, I don't think that's enough of a reason to prevent this pattern. There might be scenarios where a function needs to run *roughly* once every hour - this pattern would be perfect for that.

Or, if a user is trying to implement a state machine, this pattern would be quite handy. If we block access to this pattern, the user would have to use `@std/seq`, which intentionally requires loops to be hard-capped at a specific number. That means that a user writing a state machine that should be running for a long time will have to rely on a stable number of http requests *or* will have to restart their program at the same interval. This allows the user to write an infinite loop if they need it, in a safe and idiomatic manner.

> especially in the cluster-of-AVMs perspective when you might want the interval to only run once for the entire cluster every N minutes, while using this loop is going to have it run M times for the number of nodes in the cluster.

This would definitely be very useful and I think we should include it as well, but that feels like a different use-case to me

js-runtime/index.js
dfellis: Well, in that case we don't need the new opcodes? This was simply fixed vs variable sized and then we don't care about the 1, 2, 4, 8 differences.

bdd/spec/017_custom_event_lnn_spec.sh
dfellis: I think it would be the standard use case,  though? When the cluster autoscalers to who knows how many nodes, making sure something is run on each node is going to be far less common of a need than wanting the event to happen once at a fixed interval.

But I can see value in leaving the other form open for the rarer cases that it is desired. (Particularly around monitoring.)

dfellis (approved):

cdmistman:
The last commit fixed the test in both ln and lnn - I realized that the test was failing in lnn because the event handlers were listed in reverse order in the amm output of the lnn front-end. Since the JSR is single-threaded, it runs handlers in the order they're defined, and the reverse order was causing the handler with `emit exit 0` to run first, so the test would only fail for lnn running on JS. Since the order is supposed to be non-deterministic, we decided offline that the test should be fixed, not the compiler code.

aguillenv (approved):

PR #579, opened 2021-06-22
Change test suite to Node 14 and up

dfellis:
This unblocks the `??` and `?.` operators (amongst other features) that need at least Node 14. Since all major browsers support these operators now, it's safe to require and we can make the compiler a bit easier to read in certain circumstances.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,7 +60,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.20.1, 12.x, 14.x]
+        node-version: [14.x, 16.x]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


aguillenv (approved):

cdmistman:
does it also make sense to change the `.nvmrc` version to `v16`?

dfellis:
> does it also make sense to change the `.nvmrc` version to `v16`?

I'm not sure. Node 16 is so new features it supports aren't supported on all browsers, so it would be annoying if we used a feature that worked in local testing but then fails during the BDD testing.

I'm tempted to have it rolling in this way: 1 version is the version we actually use, and 1 version ahead for verifying that everything will still work when we migrate up a major release, with that gated on some new feature in the major release we want to use finally having major browser support?

cdmistman (approved):
sounds good!

PR #582, opened 2021-06-22
Eliminate the last flaky test

dfellis:
Realized that the test was failing occasionally because the last print statement wasn't being printed, and that was because of a race condition between the exit event running and stdout being flushed, so a 100ms wait should fix it.

### DIFF
--- a/bdd/spec/023_http_spec.sh
+++ b/bdd/spec/023_http_spec.sh
@@ -242,6 +242,7 @@ connection not found"
     End
 
     It "runs agc"
+      sleep 1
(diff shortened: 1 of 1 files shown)


depombo (approved):

dfellis:
This has actually caused the flakiness to *always* happen instead, but it doesn't do that locally?

I need to spend some more time figuring out what's happening here.

dfellis:
Argh! Why can't I reproduce that bug locally. I'm also on Linux. :(

cdmistman (approved):
fully reproducible bdd tests? 😄

PR #583, opened 2021-06-23
Warn when no vms are found

aguillenv:


avm/src/daemon/ctrl.rs
```diff
@@ -658,11 +658,10 @@ impl ControlPort {
       .filter(|vm| vm.private_ip_addr == self_ip)
       .collect();
     if self_vm_vec.len() == 0 {
-      error!(
+      warn!(
         NoDnsPrivateIp,
-        "Failed to find self in cluster. Maybe I am being shut down?"
-      )
-      .await;
+        "Failed to find self in cluster. Maybe I am being initialize?"
```
dfellis, line 663: But, we *are* deleting the DNS record for a Node before finally killing it, right?

aguillenv: My bad here, saw it outside the `Promise.all` and it is in the same, will update the message again.

aguillenv: To answer your question, we are running both promises at the same time.

dfellis: So it's a race condition right now. We want the cluster to be able to handle both situations, but it's *probably* better to trim the records first and get the cluster rebalancing data and no longer forwarding requests to the to-be-killed nodes before killing them to reduce the number of errors seen (and in the future, some sort of signal from the node-to-be-killed on when its activity is finished so zero errors ever show up on clients during rescaling of the cluster).

dfellis (approved):

cdmistman (approved):

PR #584, opened 2021-06-25
Create app.tar.gz using a temporary dir when necessary

aguillenv:


anycloud/cli/src/common/mod.rs
```diff
@@ -2,22 +2,40 @@ use std::env;
...
+        .output()
+        .unwrap();
+      println!("{:?}", output);
+      match read(format!("{}/{}", file_path, file_name)) {
+        Ok(file) => Ok(file),
+        Err(_) => Err(format!("No {} at {}", file_name, file_path).into()),
+      }
+    }
+    None => {
+      let pwd = env::current_dir();
```
dfellis, line 22: Why not just expect the full path name for a file and have the consumer code use this API directly?

anycloud/cli/src/common/mod.rs
```diff
@@ -2,22 +2,40 @@ use std::env;
...
+
+async fn get_file(file_name: &str, file_path: Option<&str>) -> Result<Vec<u8>, String> {
+  match file_path {
+    Some(file_path) => {
+      println!("{}", format!("{}/{}", file_path, file_name));
+      let output = Command::new("ls")
+        .arg(format!("{}/{}", file_path, file_name))
+        .output()
+        .unwrap();
+      println!("{:?}", output);
```
cdmistman, line 15: any reason for these prints or for the `ls`?

aguillenv: No reason at all, I used it for testing and forgot to remove it

anycloud/cli/src/common/mod.rs
aguillenv: I was trying to follow the same pattern as it was to avoid the `match` for the `pwd` every time we call it.

cdmistman (approved):

dfellis (approved):

Issue #580, opened 2021-06-22
On macOS :8088 it is not accessible when running anycloud app daemon locally


aguillenv:
- Docker does not map the ports directly with host ports on macOS unless we explicitly use the `-p` flag.
- We are not publishing the 8088 port but using the `--net=host` flag so the container is able to interact with host services.

PR #585, opened 2021-06-25
Fix docker connection to host

aguillenv:
Fixes: #580 

It works for me locally on macOS:
image

and everything keeps working: 
image


dfellis (approved):
Very nice! I didn't know that kind of thing existed!

aguillenv:
> Very nice! I didn't know that kind of thing existed!

It is quite new, if I'm not wrong it is from December 2020.

cdmistman (approved):
woah

PR #586, opened 2021-07-01
Use the filesB64 field instead of the hardwired file fields

dfellis:
Part of the clean up and unification of the two deployment paths. This changes only the way they interact with the deploy API, not how they are built on each other (that will come later)


dfellis:
Confirmation that arbitrary files can be uploaded via `alan deploy`:

Screenshot from 2021-07-01 13-16-32

Screenshot from 2021-07-01 13-16-47

dfellis:
Also confirmed that `anycloud new` also works, but there's nothing there to screenshot to tell the difference.

cdmistman (approved):

avm/src/main.rs
```diff
@@ -167,10 +170,17 @@ fn main() {
             let agz_file = matches.value_of("AGZ_FILE").unwrap();
             let app_name = matches.value_of("app-name").map(String::from);
             let config_name = matches.value_of("config-name").map(String::from);
+            let files = matches.value_of("files");
+            let mut files_b64 = HashMap::new();
+            if let Some(files) = files {
+              let names = files.split(",");
+              for name in names {
+                files_b64.insert(name.to_string(), get_agz_file_b64(name.to_string()).await);
```
aguillenv, line 178: Are you sure the `get_agz_file_b64` fn works for this?

avm/src/main.rs
```diff
@@ -167,10 +170,17 @@ fn main() {
             let agz_file = matches.value_of("AGZ_FILE").unwrap();
             let app_name = matches.value_of("app-name").map(String::from);
             let config_name = matches.value_of("config-name").map(String::from);
+            let files = matches.value_of("files");
+            let mut files_b64 = HashMap::new();
+            if let Some(files) = files {
+              let names = files.split(",");
```
aguillenv, line 176: I was going to suggest if a `trim` would be necessary, but one cannot add spaces on bash command, right?

depombo (approved):

avm/src/main.rs
dfellis: Yep! That's what my second screenshot was about: to show that it decoded just fine.

The name of that function isn't great, but it works perfectly.

avm/src/main.rs
dfellis: It's theoretically possible but you have to intentionally bypass bash's standard parsing with backslashes or quotes.

aguillenv (approved):

PR #587, opened 2021-07-02
019 - Datastore Compute RFC

dfellis:


rfcs/019 - Datastore Compute RFC.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+```ln
+const bar = 'bar';
+ns('someNamespace').mut('foo').with(bar).mutOnly(fn (bar: string, foo: string) {
+  foo = foo + bar;
+});
+```
+
+All eight potential remote compute possibilities are handled with this API. (Technically the ninth where it's "mutation only" but also mutates the outer scope could exist, but it's identical in performance to `ns().mut().closure()` and so it was elided.)
```
aguillenv, line 122: But how would we know is a "mutation only" that is mutating an outer scope variable if we use the same `ns().mut().closure()`? Because as I understand it, the `mutOnly` opcode will give us the opportunity to make the mutation necessary but do not stop outer node exec until this inner node finish

dfellis: If we mutate the outer scope then the first node needs to wait for the mutated scope to come back, so it would have to stop outer node execution. That's why I said that it's functionally equivalent.

aguillenv: Oh that's true. I got a case in mind but would be handled by the eight case

rfcs/019 - Datastore Compute RFC.md
```diff
@@ -0,0 +1,202 @@
...
+
+This would be a minor update if we were post-1.0 as it should have zero breaking impact on existing code.
+
+## Proposal
+
+When Alan has user-defined code passed into a built-in opcode, such as the `Array<T>`'s `map`, the inner function is compiled into AGA as a `closure for <uuid>` that temporarily takes over the outer scope's `HandlerMemory` and executes the inner function with the relevant arguments passed in on each run, then the output is placed into the `HandlerMemory` and it continues on. If `mapLin` was used, it runs sequentially and the inner function can mutate values in the outer scope safely (and is allowed to do so).
+
+There is no theoretical reason why `map` must do its work on the same machine in the cluster. It could copy the `HandlerMemory` to any other node, do its work there, and then copy it back at the end. This simply doesn't make any sense from a performance perspective since you must spend the time to serialize, transmit, deserialize, compute, serialize, transmit, and finally deserialize, while staying on the same machine only requires creating a special child `HanlderMemory` associated with the original one, almost no overhead versus the compute step.
```
cdmistman, line 42: typo: `HanlderMemory`

rfcs/019 - Datastore Compute RFC.md
```diff
@@ -0,0 +1,202 @@
...
+let bar = 'bar';
+const baz = ns('someNamespace').ref('foo').closure(fn (foo: string) {
+  bar = 'foobar: ' + foo + bar;
+  return foo.length();
+});
+```
+
+```ln
+const bar = 'bar';
+const baz = ns('someNamespace').ref('foo').with(bar).run(fn (bar: string, foo: string) = #bar + #foo);
```
cdmistman, line 82: nit: it's a bit more intuitive for the function parameters to be `fn (foo: string, bar: string)` since that's the order of declaration in the method chaining

rfcs/019 - Datastore Compute RFC.md
```diff
@@ -0,0 +1,202 @@
...
+```
```
aguillenv, line 83: I had mixed feelings having the distinction between `run` and `closure` but having it explicit define is easier to process and less ambiguous to the user so nothing here, just a comment :)

aguillenv (approved):

rfcs/019 - Datastore Compute RFC.md
dfellis:
```suggestion
There is no theoretical reason why `map` must do its work on the same machine in the cluster. It could copy the `HandlerMemory` to any other node, do its work there, and then copy it back at the end. This simply doesn't make any sense from a performance perspective since you must spend the time to serialize, transmit, deserialize, compute, serialize, transmit, and finally deserialize, while staying on the same machine only requires creating a special child `HandlerMemory` associated with the original one, almost no overhead versus the compute step.
```

rfcs/019 - Datastore Compute RFC.md
```diff
@@ -0,0 +1,201 @@
...
+let bar = 'bar';
+const baz = ns('someNamespace').ref('foo').closure(fn (foo: string) {
+  bar = 'foobar: ' + foo + bar;
+  return foo.length();
+});
+```
+
+```ln
+const bar = 'bar';
+const baz = ns('someNamespace').ref('foo').with(bar).run(fn (bar: string, foo: string) = #bar + #foo);
```
dfellis, line 82:
```suggestion
const baz = ns('someNamespace').ref('foo').with(bar).run(fn (foo: string, bar: string) = #bar + #foo);
```

rfcs/019 - Datastore Compute RFC.md
```diff
@@ -0,0 +1,201 @@
...
+const baz = ns('someNamespace').mut('foo').closure(fn (foo: string) {
+  foo = foo + bar;
+  bar = bar * foo.length();
+  return bar.length();
+});
+```
+
+```ln
+const bar = 'bar';
+const baz = ns('someNamespace').mut('foo').with(bar).run(fn (bar: string, foo: string) {
```
dfellis, line 103:
```suggestion
const baz = ns('someNamespace').mut('foo').with(bar).run(fn (foo: string, bar: string) {
```

rfcs/019 - Datastore Compute RFC.md
```diff
@@ -0,0 +1,201 @@
...
+
+```ln
+ns('someNamespace').mut('foo').mutOnly(fn (foo: string) {
+  foo = foo + foo;
+});
+```
+
+```ln
+const bar = 'bar';
+ns('someNamespace').mut('foo').with(bar).mutOnly(fn (bar: string, foo: string) {
```
dfellis, line 117:
```suggestion
ns('someNamespace').mut('foo').with(bar).mutOnly(fn (foo: string, bar: string) {
```

rfcs/019 - Datastore Compute RFC.md
```diff
@@ -0,0 +1,201 @@
...
+fn enqueue(queueName: string, payload: any) {
+  ns('queue').mut(queueName).with(payload).mutOnly(fn (payload: any, queue: Array<any>) {
+    queue = [payload] + queue;
+  });
+}
+
+fn dequeue(queueName: string, default: any): any = ns('queue')
+  .mut(queueName)
+  .with(default)
+  .run(fn (default: any, queue: Array<any>): any = queue.pop() || default) || default;
```
dfellis, line 145:
```suggestion
  .run(fn (queue: Array<any>, default: any): any = queue.pop() || default) || default;
```

rfcs/019 - Datastore Compute RFC.md
```diff
@@ -0,0 +1,201 @@
...
+With these primitives, we can easily build a queue API:
+
+```ln
+fn createQueue(queueName: string) {
+  // Taking advantage of the current type unsafety with datastore
+  ns('queue').set(queueName, new Array<int64> []);
+}
+
+fn enqueue(queueName: string, payload: any) {
+  ns('queue').mut(queueName).with(payload).mutOnly(fn (payload: any, queue: Array<any>) {
```
dfellis, line 137:
```suggestion
  ns('queue').mut(queueName).with(payload).mutOnly(fn (queue: Array<any>, payload: any) {
```

depombo (approved):

dfellis:
As a note before merging it: no one is satisfied with the method names here, but we also can't think of anything better.

If we come up with something different that we actually like we'll go with that, instead.

PR #590, opened 2021-07-08
Drop AnyCloud

dfellis:
Now that AnyCloud has been refactored into a simple cli wrapper around `alan deploy` and *most* of the special hackery for it has been eliminated, turning the runtime piece into a standard Alan application.

This PR fold the portions of the `anycloud` codebase that the `alan deploy` sub-command depends on into the AVM, with the rest either moved into [the (old) new `anycloud` repo](https://github.com/alantech/anycloud) or outright deleted. Follow-up work is to further simplify and excise code that isn't really necessary but still entangled, but kept the code refactoring in this PR to a minimum.


avm/src/main.rs
```diff
@@ -3,20 +3,21 @@ use std::env;
...
-use anycloud::deploy;
-use anycloud::oauth::authenticate;
 use base64;
 use clap::{crate_name, crate_version, App, AppSettings, SubCommand};
 use tokio::runtime::Builder;
 
 use crate::compile::compile::compile;
 use crate::daemon::daemon::{start, CLUSTER_SECRET};
+use crate::deploy::common::get_agz_file_b64;
+use crate::deploy::deploy as inner_deploy;
```
depombo, line 13: `deploy_mod`?

depombo (approved):

avm/src/main.rs
dfellis: Discussed offline, going for `use crate::cloud::deploy` instead.

depombo (approved):

PR #591, opened 2021-07-09
Stats logic

aguillenv:
Fixes: https://github.com/alantech/alan/issues/592


depombo (approved):

avm/src/daemon/stats.rs
```diff
@@ -206,3 +218,97 @@ pub async fn get_v1_stats() -> DaemonResult<VMStatsV1> {
...
+    Some(acc) => acc / samples.len() as f64,
+    None => 0 as f64,
+  }
+}
+
+fn get_max_proc_cpu_usage(samples: &Vec<CPUSample>) -> f64 {
+  match samples
+    .iter()
+    .map(|s| s.max_proc_cpu_usage)
+    .reduce(|a, b| a + b)
```
dfellis, line 309: Is this right? Wouldn't you want to take the `max(a, b)` instead of `a + b` in that reduce?

aguillenv: I've just checked the code here: https://github.com/alantech/deploy/pull/122/files#diff-f00db8b6a98c6dd57f17c00982110f875edf281162a2af23be0b075e059d377aL28-L54.

What I think is wrong is the name of the variable. It actually should be something like: `avg_max_proc_cpu_usage`. what do you think?

dfellis: That is more descriptive. I like it.

Whether or not that's what we *want* for this variable is something else, but I think we can stick with existing behavior for now and then revisit.

dfellis (approved):

PR #594, opened 2021-07-12
Non http deployments

aguillenv:
Fixes: https://github.com/alantech/alan/issues/593


dfellis:
Discussed this offline, we can simplify this and not have a special-cased exit code. Just having the exit opcode become an IO opcode that `await`s sending the termination message to the deploy service when in daemon mode and then still exits normally would be better.

dfellis (approved):

depombo (approved):

PR #596, opened 2021-07-15
Implement datastore distributed compute

dfellis:
- [x] Implement the stdlib and compiler changes
- [x] Implement the js-runtime opcodes
- [x] Implement the AVM opcodes
- [x] Implement BDD test


js-runtime/index.js
```diff
@@ -1051,6 +1051,53 @@ module.exports = {
...
+    const [out, newb] = await func(a, b);
+    ds[`${nskey[0]}:${nskey[1]}`] = newb;
+    return [true, out];
+  },
+  dsmonly: async (nskey, func) => {
+    let val = ds[`${nskey[0]}:${nskey[1]}`];
+    const newval = await func(val);
+    ds[`${nskey[0]}:${nskey[1]}`] = newval;
+  },
+  dswonly: async (wth, func) => {
```
aguillenv, line 1084: I'd rather call it `dsmwonly` to follow more or less the same pattern in all opcodes

dfellis: I can't remember if the opcode name limit is 7 or 8 characters, so I went with 7.

compiler/src/ammtoaga/index.ts
```diff
@@ -10,18 +10,15 @@ type AddressMap = {
 type EventDecs = {
   [name: string]: number;
 };
-type MemoryMap = {
-  [name: string]: number;
-};
 type HandlerMem = {
   memSize: number;
-  addressMap: MemoryMap;
+  addressMap: AddressMap;
```
dfellis, line 15: Some of the code referenced addresses as JS numbers, others as BigInt. Since the address space *is* 64-bit just use BigInt everywhere, which simplifies the code a lot.

compiler/src/ammtoaga/index.ts
```diff
@@ -493,8 +492,17 @@ const loadStatements = (
...
-      const resultAddress = localMem[asgn.get('decname').t.trim()];
-      localMemToLine[resultAddress] = line;
+      const asgnName = asgn.get('decname').t.trim();
+      let resultAddress;
+      if (localMem[asgnName]) {
+        resultAddress = localMem[asgnName];
+      } else if (hasClosureArgs && fnArgs.indexOf(asgnName) > -1) {
+        resultAddress = globalMem[asgnName + fnName];
+        mutatedArgs[asgnName + fnName] = true;
+      } else {
```
dfellis, line 502: Allows mutations to input arguments, and also flags it as an input arg that is mutated

compiler/src/ammtoaga/index.ts
```diff
@@ -672,13 +680,29 @@ const loadStatements = (
     vec.push(s);
     line += 1;
   }
+  fnArgs.forEach((arg, i) => {
+    if (globalMem.hasOwnProperty(arg + fnName) && mutatedArgs[arg + fnName]) {
```
dfellis, line 684: If the input argument was mutated, re-reference it back to the original argument offset at the end of the closure so it can be picked up properly by the calling opcode, if necessary

compiler/src/lntoamm/Microstatement.ts
```diff
@@ -1165,19 +1165,48 @@ ${letdeclarationAst.t} on line ${letdeclarationAst.line}:${
...
-        scope,
-        true,
-        val.outputName,
-        val.outputType,
-        [],
-        [],
-        letAlias,
-      ),
-    );
+    if (val.outputType === Type.builtinTypes.string) {
```
dfellis, line 1168: Basically creates code that looks like:

```ln
const someUUID = 'my string';
let actualVarName = someUUID.clone();
```

Then the string constant is hoisted away and the let variable clones the global memory version to allow mutation.

js-runtime/index.js
```diff
@@ -1065,14 +1065,14 @@ module.exports = {
     const nskey = wth[0];
     const a = wth[1];
     const b = ds[`${nskey[0]}:${nskey[1]}`];
-    return [true, await func(a, b)];
```
dfellis: I hurt myself in my confusion when I swapped the argument order like @cdmistman suggested (which makes a lot of sense! but my js was already written the other way around).

avm/src/vm/opcode.rs
```diff
@@ -3831,6 +3831,339 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+    })
+  });
+  unpred_cpu!(dsrwith => fn(args, mut hand_mem) {
+    Box::pin(async move {
+      let with = hand_mem.read_fractal(args[0])?;
+      let nsref = hand_mem.read_from_fractal(&with, 0).0;
+      let ns = HandlerMemory::fractal_to_string(hand_mem.read_from_fractal(&nsref, 0).0)?;
+      let key = HandlerMemory::fractal_to_string(hand_mem.read_from_fractal(&nsref, 1).0)?;
+      let nskey = format!("{}:{}", ns, key);
+      hand_mem.register_out(args[0], 1, CLOSURE_ARG_MEM_START + 2)?;
```
aguillenv, line 3926: What does this `register_out` is doing?

avm/src/vm/opcode.rs
```diff
@@ -3831,6 +3831,339 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
...
+      Ok(hand_mem)
+    })
+  });
+  unpred_cpu!(dswonly => fn(args, hand_mem) {
+    Box::pin(async move {
+      let with = hand_mem.read_fractal(args[0])?;
+      let nsref = hand_mem.read_from_fractal(&with, 0).0;
+      let ns = HandlerMemory::fractal_to_string(hand_mem.read_from_fractal(&nsref, 0).0)?;
+      let key = HandlerMemory::fractal_to_string(hand_mem.read_from_fractal(&nsref, 1).0)?;
+      let nskey = format!("{}:{}", ns, key);
```
aguillenv, line 4051: And why the `register_out` it is not necessary here?

avm/src/vm/opcode.rs
dfellis: So this is a "with" opcode, so the value that's being passed in as the second parameter is nested as the second element of the `With<NsRef, any>` generic. Since all of our opcodes operate on the top-level addresses and have no nested pathing, this memory call allows us to take the value in an array offset and make a pointer to it in a top-level address.

It is literally identical to the [`register` opcode](https://github.com/alantech/alan/blob/main/avm/src/vm/opcode.rs#L2592-L2599) (which still has that old naming scheme because we didn't want to break compilation / change the compiler, but maybe we should just do that). There are three of these `HandlerMemory` methods:

1. `register` just creates a pointer from address A to address B
2. `register_out` creates a pointer in an array in address A at offset X to address B
3. `register_in` creates a pointer from address A to an array in address B at offset X

I hope that helps!

avm/src/vm/opcode.rs
dfellis: I did it [here instead](https://github.com/alantech/alan/pull/596/files#diff-aa0e5a81bac19bdacb5f06f8d0cefe2516b53e02227ac03e058f2230d02f7ba2R4066)

It shouldn't affect things either way. They're very similar (since they all do nearly the same things with slight differences) but I wrote them separately and apparently put all of the `HandlerMemory` mutation down in the `if is_key_owner` block, which is probably the better way to do this?

aguillenv: Yes, my confusion was that in `dsrwith` for instance you `register_out` the original `hand_mem` (before the fork) and later on you `register_out` the forked `hm`, while in this opcode you only `register_out` the forked one.

js-runtime/index.js
cdmistman: I remember it actually being 7, but I can't find anywhere in the avm/compiler where it would get shortened to 7 (and they would need to agree in order for the opcode hashmap to work). It must've been fixed at some point

compiler/src/ammtoaga/index.ts
cdmistman: i think a future task will be to use amm as an SSA form and performing variable assignment optimization. it should have a sizeable impact on runtime, as it avoids new `Vec` allocations on each instruction in the avm. This will require the type information to be parsed at the amm level, but the logic can be shared between all `ammto*` stages - it should even result in higher performance with JSR until V8 JIT kicks in

cdmistman (approved):

dfellis:
This is now finally ready for review! As discussed we're going to leave the mutation race condition for a follow-up PR. I don't know why `dashmap` is not unlocking the key when I use the locking mechanism, and it's looking like I'll need to replace dashmap entirely in a followup PR.

That's probably also where I reorganize all of datastore to be less all-over-the-place.

avm/src/daemon/ctrl.rs
```diff
@@ -558,6 +567,413 @@ fn handle_keys() -> Result<Response<Body>, Infallible> {
...
+      HandlerMemory::transfer(&ds, 0, &mut hm, CLOSURE_ARG_MEM_START + 1)?;
+      let hm = subhandler.run(hm).await?;
+      // Also grab the mutation to the datastore value and re-insert it
+      let mut newds = HandlerMemory::new(None, 1)?;
+      HandlerMemory::transfer(&hm, CLOSURE_ARG_MEM_START + 1, &mut newds, 0)?;
+      drop(ds);
+      DS.insert(nskey, newds);
+    }
+    None => {
+      // Do nothing
```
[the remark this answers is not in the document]
dfellis: The `only` opcodes are side-effect-only functions. They don't return anything to the consuming side (eg incrementing a counter or pushing a value onto an array).

So since the return type is `void` there's literally nothing we can do in this case where the key you're trying to mutate doesn't even exist.

It should be possible to create `refOr` and `mutOr` functions in pure alan that first check for the key existing and if it doesn't exist it sets it with the provided value if this issue is a problem for them.

In general I don't think the `mutOnly` opcodes (`dsmonly` and `dswonly`) are gonna get much use, though.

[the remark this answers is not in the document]
aguillenv: Oh! I did not realized the returning type. Got it!

avm/src/daemon/ctrl.rs
```diff
@@ -558,6 +567,413 @@ fn handle_keys() -> Result<Response<Body>, Infallible> {
...
+    }
+    Err(err) => {
+      // TODO: What error message here? Also should this also be a valid HM out of here?
+      eprintln!("{:?}", err);
+      Ok(Response::builder().status(500).body("fail".into()).unwrap())
+    }
+  }
+}
+
+async fn dsrwith_inner(req: Request<Body>) -> DaemonResult<Arc<HandlerMemory>> {
```
cdmistman, line 693: nit: `dsrwith_inner` is the exact same as `dsrrun_inner` - is there a reason they're not collapsed into a single function? from what i can tell, all of the `_inner` functions can be collapsed into 1 function with an enum, which can be passed into a `handle_distributed` which collapses all of the `handle_` functions

cdmistman: i see the side-effect-only functions below - those can be handled with a `handle_distributed_side_effect` which would be nearly the same as `handle_distributed` except it would `.map(|_| ())` the `Result` from the collapsed `_inner` function

avm/src/daemon/ctrl.rs
```diff
@@ -996,6 +1412,385 @@ impl ControlPort {
     }
   }
 
+  pub async fn dsrrun(
```
cdmistman, line 1415: all of these top-level handlers except `dsmonly` and `dswonly` can also be collapsed, while those 2 can be collapsed with each other. likewise, these `_inner` handlers can be collapsed

compiler/src/ammtojs.ts
```diff
@@ -55,16 +66,28 @@ const functionbodyToJsText = (fnbody: LPNode, indent: string) => {
         : 'undefined';
       outText += `r.emit('${name}', ${arg})\n`;
     } else if (statement.has('exits')) {
-      outText += `${statement.get('exits').t.trim()}\n`;
+      if (typeof mutationDirective[1] === 'boolean') {
```
cdmistman, line 69: these `typeof` checks are kinda bleh - can `mutationDirective[1]` be an enum instead? Or check on `!== undefined` to be more explicit?

avm/src/vm/opcode.rs
```diff
@@ -3838,6 +3838,323 @@ pub static OPCODES: Lazy<HashMap<i64, ByteOpcode>> = Lazy::new(|| {
       Ok(hand_mem)
     })
   });
+  unpred_cpu!(dsrrun => fn(args, mut hand_mem) {
```
cdmistman, line 3841: these opcodes can just call the functions in `daemon::ctrl` right?

avm/src/daemon/ctrl.rs
dfellis: I didn't care for the duplication, but I couldn't see how to collapse them, since they each differ by calling a different inner function. And the inner functions all differ on the exact details of how they set up the handler memory that's serialized and which endpoint they call, so they can't really be collapsed.

avm/src/vm/opcode.rs
dfellis: No, because the control port itself doesn't exist when you `alan run foo.agz` instead of `alan deploy new foo.agz`.

aguillenv (approved):

Issue #534, opened 2021-05-20
Function types


cdmistman:
These can be implemented internally to enable things like `fn map`, etc. This will be necessary to fix fn selection in the compiler. While exposing complex function types to users (eg `Function<Arg0, void>` as proposed in [RFC 11](https://github.com/alantech/alan/blob/main/rfcs/011%20-%20Function%20Type%20RFC.md)) is blocked on #537, a workaround with the current Alan `function` type is possible by exposing a ducked `Type` in `Types.ts` that accepts any new internally-recognized `Function` type (or whatever the class is defined in `Types.ts` is). This will enable the PR for #537 to expose the new Alan `Function<P, R>` type

PR #600, opened 2021-07-20
[lnn] fn selection fix

### LINKED ISSUE #599: fix function selection

cdmistman:
fixes #599 

I kept some debug output *in case* something happens but I don't expect any of them to actually print. Other than that the documentation should be pretty thorough and should in the end be more accurate than the current selection system. This system should also determine all types of values except for those that depend on the parameters of the function, which means that functions are fully type-checked (instead of only partially type-checked) before inlining, even if they're on the cold path (functions not used by any handler).

### DIFF
--- a/compiler/src/lnntoamm/Expr.ts
+++ b/compiler/src/lnntoamm/Expr.ts
@@ -420,7 +420,7 @@ export default abstract class Expr {
           stmts.push(dec);
           right = dec.ref();
         }
-        let retTys: Type[] = [];
+        const retTys: Type[] = [];
         while (ops.length > 0) {
           const op = ops.pop();
           const selected = op.select(metadata.scope, left.ty, right.ty);
@@ -588,12 +588,17 @@ class Call extends Expr {
   }
 
   private fnSelect(): [Fn[], Type[]] {
-    return FunctionType
-      .matrixSelect(this.fns, this.args.map(a => a.ty), this.scope)
-      .reduce(
-        ([fns, tys], [fn, ty]) => [[...fns, fn], [...tys, ty]],
-        [new Array<Fn>(), new Array<Type>()],
-      );
+    return FunctionType.matrixSelect(
+      this.fns,
+      this.args.map((a) => a.ty),
+      this.scope,
+    ).reduce(
+      ([fns, tys], [fn, ty]) => [
+        [...fns, fn],
+        [...tys, ty],
+      ],
+      [new Array<Fn>(), new Array<Type>()],
+    );
   }
 
   cleanup() {
--- a/compiler/src/lnntoamm/Expr.ts
+++ b/compiler/src/lnntoamm/Expr.ts
@@ -363,7 +363,7 @@ export default abstract class Expr {
               retTys = [...retTys, ...selTys];
               return [fns, retTys];
             },
-            [new Array<Fn>(), new Array<Type>()] as [Fn[], Type[]],
+            [new Array<Fn>(), new Array<Type>()],
           );
           const retTy = Type.oneOf(retTys);
           precedences[applyIdx] = new Call(
@@ -566,7 +566,7 @@ class Call extends Expr {
     // first reduction
     const argTys = args.map((arg) => arg.ty);
     const selFns = FunctionType.matrixSelect(fns, argTys, metadata.scope);
-    fns = selFns.map(([fn, _ty]) => fn);
+    fns = selFns.map((selFn) => selFn[0]);
     const retTy = Type.oneOf(selFns.map(([_fn, ty]) => ty));
     // now, constrain all of the args to their possible types
     // makes it so that the type of the parameters in each position are in their own list
@@ -654,8 +654,12 @@ class Call extends Expr {
       console.log('expected output type:', ty);
       throw new Error(`no function selected`);
     }
-    // Fn.select should implement the matrix so that the most reasonable
-    // choice is last in the fn array.
+    // FunctionType.matrixSelect implements the matrix so that the most
+    // reasonable choice is last in the fn array. "Reasonableness" is computed
+    // with 2 factors: 1st is alignment with given OneOf types. If `add(1, 0)`
+    // is called, the literal types should prefer `int64` to `int32` etc. The
+    // other factor is order of declaration - Alan should always prefer using
+    // functions that are defined last.
     const fn = selFns.pop();
     fn.inline(amm, this.args, kind, name, ty);
   }
--- a/compiler/src/lnntoamm/Expr.ts
+++ b/compiler/src/lnntoamm/Expr.ts
@@ -448,6 +448,9 @@ export default abstract class Expr {
     return [stmts, precedences.pop() as Ref];
   }
 
+  /**
+   * @returns true if more cleanup might be required
+   */
   cleanup(): boolean {
     // most implementing Exprs don't have anything they need to do.
     // I just didn't want to expose any of the Expr classes except
@@ -595,7 +598,7 @@ class Call extends Expr {
 
   cleanup() {
     const [fns, tys] = this.fnSelect();
-    const isChanged = this.fns.length === fns.length;
+    const isChanged = this.fns.length !== fns.length;
     this.fns = fns;
     this.retTy.constrain(Type.oneOf(tys), this.scope);
     return isChanged;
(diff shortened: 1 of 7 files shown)


compiler/src/lnntoamm/Expr.ts
```diff
@@ -356,16 +356,23 @@ export default abstract class Expr {
...
-            (fns, op) => [...fns, ...op.select(metadata.scope, applyTo.ty)],
-            new Array<Fn>(),
+          const [fns, retTys] = operators.reduce(
+            ([fns, retTys], op) => {
+              const [selFns, selTys] = op.select(metadata.scope, applyTo.ty);
+              fns = [...fns, ...selFns];
+              retTys = [...retTys, ...selTys];
+              return [fns, retTys];
+            },
+            [new Array<Fn>(), new Array<Type>()] as [Fn[], Type[]],
```
dfellis, line 366: What a failure of TS's type inference here...

compiler/src/lnntoamm/Expr.ts
```diff
@@ -356,16 +356,23 @@ export default abstract class Expr {
...
           );
+          const retTy = Type.oneOf(retTys);
```
dfellis, line 368: Which one does `Type.oneOf` choose? Just the first, or something more complex?

compiler/src/lnntoamm/Expr.ts
```diff
@@ -543,7 +565,9 @@ class Call extends Expr {
     }
     // first reduction
     const argTys = args.map((arg) => arg.ty);
-    fns = Fn.select(fns, argTys, metadata.scope);
+    const selFns = FunctionType.matrixSelect(fns, argTys, metadata.scope);
+    fns = selFns.map(([fn, _ty]) => fn);
```
dfellis, line 569: Nit: couldn't this just be `fns = selFns.map(selFn => selFn[0]);`

Doing function argument-level destructuring and assignment to an unused variable seems like a performance penalty that will affect interpreter mode heavily and may still trip up the JIT on being able to fully optimize back down to the form I wrote above?

compiler/src/lnntoamm/Expr.ts
cdmistman: when `retTy.instance()` is called, it'll return the type furthest back in the list. that's to maintain the last-defined rule

compiler/src/lnntoamm/Expr.ts
```diff
@@ -607,7 +654,9 @@ class Call extends Expr {
       console.log('expected output type:', ty);
       throw new Error(`no function selected`);
     }
-    const fn = selected.pop();
+    // Fn.select should implement the matrix so that the most reasonable
+    // choice is last in the fn array.
+    const fn = selFns.pop();
```
dfellis, line 659: Perhaps a warning mode might be worthwhile to allow users to know that the type was ambiguous and it's choosing from a list? "most reasonable" might not be what they're actually after and letting them know they should annotate their desired type seems like a good idea?

compiler/src/lnntoamm/Fn.ts
```diff
@@ -41,6 +42,37 @@ export default class Fn {
     this.body = body;
     this.metadata =
       metadata !== null ? metadata : new MetaData(scope, this.retTy);
+    while (
+      this.body.reduce(
+        (carry, stmt) => stmt.cleanup(this.scope) || carry,
```
dfellis, line 47: So we are now having the min Node version be 14, so we could do this instead:

```suggestion
        (carry, stmt) => stmt.cleanup(this.scope) ?? carry,
```

compiler/src/lnntoamm/Expr.ts
cdmistman: that's a good idea, but that'd have to wait for a future PR that focuses on better error handling throughout the compiler in general. the "most reasonable" wording has 2 meanings here: first, it prefers functions that are most specialized for the given argument types (for example, it picks `fn add(int64, int64)` over `fn add(int32, int32)` given raw literals, even if the latter was declared after the former since we prefer `int64` over `int32` in alan - this really only means that it impacts such cases). The other meaning is that it still maintains order of definition, so the last-defined function is given higher precedence over the first-defined function, even if both have the same function signature.

The comment is slightly outdated (I made it in one of the earlier commits before I moved the matrix selection to `Types.ts`) so I'll update it and add clarity

compiler/src/lnntoamm/Types.ts
```diff
@@ -242,6 +246,162 @@ class Builtin extends Type {
...
+        }, new Array<[Fn, Type]>()),
+      );
+      fnsByWeight.set(weight, fnsForWeight);
+      if (
+        indices.every(
+          (idxInDim, dimIdx) => idxInDim === matrix[dimIdx].length - 1,
+        )
+      ) {
+        break;
+      }
```
dfellis, line 309: Is this conditional guaranteed to always eventually execute? Otherwise this for loop can potentially infinite loop.

compiler/src/lnntoamm/Fn.ts
cdmistman: the carry is a boolean though, so I think it's slightly more readable to keep it as `||`

compiler/src/lnntoamm/Types.ts
```diff
@@ -242,6 +246,162 @@ class Builtin extends Type {
...
+      weightFns = weightFns.filter(
+        ([weightedFn, _retTy]) =>
+          fns.findIndex(([fn, _retTy]) => fn === weightedFn) === -1,
+      );
+      return [...fns, ...weightFns];
+    }, new Array<[Fn, Type]>());
+    if (ret.length > originalLength) {
+      console.log('~~~ ERROR');
+      console.log('original:', originalLength);
+      console.log('retLengt:', ret.length);
```
dfellis, line 338:
```suggestion
      console.log('retLength:', ret.length);
```

compiler/src/lnntoamm/Types.ts
```diff
@@ -242,6 +246,162 @@ class Builtin extends Type {
...
+    } else if (ty instanceof OneOf || ty instanceof Generated) {
+      return ty.compatibleWithConstraint(this, scope);
+    } else {
+      return false;
+    }
+  }
+
+  constrain(to: Type, scope: Scope): void {
+    console.log(to);
+    TODO('figure out what it means to constrain a function type');
```
dfellis, line 366: Probably similar to how generic types would be constrained? You could treat function types like `Fn<A, R>` with A being the arg types and R being the return type, and then either also implement variadic generic types or just hardwire something like `Arg0`, `Arg1<A>`, `Arg2<A, B>`, ... to some large-ish maximum arg length.

Those simple generics could themselves be generated in the code and never exposed to the user so it wouldn't be *too* crazy to go up to something like `Arg99` if that is possible.

compiler/src/lnntoamm/Types.ts
cdmistman: yes. the index iteration is similar to how normal addition is performed. 

Say dimension 0 (using 0-indexing since that's how the `matrix` var works) has 2 elements, dimension 1 has 1 element, and dimension 2 has 2 elements. The first loop run would be `indices=[0,0,0]`. the `reduceRight` on the next line performs the addition-like iteration, resulting in `indices=[0,0,1]`. On the next loop, it's `indices=[1,0,0]` because of the carry skipping dimension 1. After this `if` check gets performed, it performs the final addition to get `indices=[1,0,1]` but must go to the beginning of the loop to calculate weight, etc before finally passing this `if` check and `break`ing

dfellis (approved):

compiler/src/lnntoamm/Types.ts
cdmistman: right, but I'm waiting until I tackle #534 before I figure this out, since there aren't any cases where `FunctionType.constrain` is called (unless there's some weird bug)

compiler/src/lnntoamm/Expr.ts
dfellis: Yeah, no problem, just pointing out that this is a good place for eventually adding a warning that the compiler could emit and also maybe a fail-on-warning mode check if we want that, too.

compiler/src/lnntoamm/Fn.ts
dfellis: Ahh, I thought you were replacing a null-ish result from `cleanup` when I first wrote this. Keeping it as `||` is actually correct and `??` would break it.

compiler/src/lnntoamm/Expr.ts
cdmistman: er, it does work without the `as` but it wasn't in a previous commit. removing the `as`

dfellis (approved):

depombo (approved):

PR #602, opened 2021-07-21
install nested dependencies

depombo:


dfellis (approved):

std/deps.ln
```diff
@@ -14,11 +14,12 @@ export fn add(remote: string) {
   const parts = remote.split('/');
   const repo = parts[length(parts) - 1] || '';
   const group = parts[parts.length() - 2] || '';
-  const dest = '/dependencies/' + group + '/' + repo;
-  const rm = exec('rm -rf .' + dest);
-  const git = exec('git clone ' + remote + ' .' + dest);
+  const dest = 'dependencies/' + group + '/' + repo;
```
aguillenv, line 17: Is it possible to have empty `group` and empty `repo`? If so, could be a weird `dest` path.

depombo: I don't think so? The three biggest in terms of git hosting all require it as far as I know (Github, Gitlab, Bitbucket)

aguillenv (approved):

PR #604, opened 2021-07-23
[lnn] generics

### LINKED ISSUE #537: Generic Types
Necessary for things like arithmetic which depends on having `Result<T>` available

cdmistman:
fixes #537


bdd/spec/001_event_lnn_spec.sh
```diff
@@ -91,7 +91,7 @@ Describe "Events"
       event aNumber: int64
 
       on aNumber fn(num: int64) {
-        ('I got a number! ' + num.toString()).print();
+        print('I got a number! ' + num.toString());
```
dfellis, line 94: Reason for this change?

PR #607, opened 2021-08-05
Add dependencies permissions RFC

aguillenv:


rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,104 @@
...
+
+This is the basic case when we want to use `new_dep_A` in our project but we do not want this library to have access to `std/cmd` for example. So, we we add this library we will specify the standard libraries blacklisted. This block would need to work recursively due to module resolution priorities. The flow will be:
+
+![Block at library level](./lib-lvl-v1.png)
+
+The `.dependencies.ln` file could look something like:
+
+Block `@std/cmd` but do not override if any mock exists:
+```ln
+add('https://github.com/org/new_dep_A', some(['cmd']), false);
```
dfellis, line 58: First comment: That should probably be `some(['@std/cmd'])`. It is possible that someone would want to restrict access to third-party dependencies that the library has declared a dependency on.

Second comment: This is an awkward API and the meaning of the arguments aren't clear while reading the code itself.

Some ideas on how to make this a nicer API:

### 1. Utilize multiple functions with the same name

```js
add('https://github.com/org/new_dep_A', ['@std/cmd']); // Your first example
add('https://github.com/org/new_dep_A', ['@std/cmd'], true); // Your second example
add('https://github.com/org/new_dep_A'); // Your third example
```

You can define `fn add(url: string), fn add(url: string, blockedLibs: Array<string>), fn add(url: string, blockedLibs: Array<string>, fullBlock: bool)` and it'll select the right one in the right situation.

This still doesn't explain *what* the args mean, though.

### 2. Make it a fluent API

```js
'https://github.com/org/new_dep_A'.block('@std/cmd').add(); // Your first example
'https://github.com/org/new_dep_A'.block('@std/cmd').block('@std/tcp').fullOverride().add(); // Your second example
'https://github.com/org/new_dep_A'.block(['@std/cmd', '@std/tcp']).fullOverride().add(); // Another form of your second example
'https://github.com/org/new_dep_A'.add() // Your third example
```

But even this has some issues (what if you want to override a mock for `@std/cmd` but not for `@std/tcp`?) So maybe this would be better:

```js
'https://github.com/org/new_dep_A'.block('@std/cmd').add(); // Your first example
'https://github.com/org/new_dep_A'.block('@std/cmd').fullBlock('@std/tcp').add(); // Your second example
'https://github.com/org/new_dep_A'.block(['@std/cmd', '@std/tcp']).add(); // Another form of your second example
'https://github.com/org/new_dep_A'.add() // Your third example
```

In either case, the implementation is something like this: `fn add(url: string)` for the basic situation, but for the rest there's a `Dependency` struct that tracks the metadata that you are building and the `fn add(dep: Dependency)` function parses that struct into the correct set of rules to build. `fn add(url: string)` could simply create one of these `Dependency` values and then call the other add on it to keep things DRY.

### 3. Expose that `Dependency` type directly

Something like:

```js
add(new Dependency {
  url: 'https://github.com/org/new_dep_A',
  block: ['@std/cmd'],
  fullBlock: [],
}); // First example

new Dependency {
  url: 'https://github.com/org/new_dep_A',
  block: [],
  fullBlock: ['@std/cmd'],
}.add(); // Second example

add(new Dependency {
  url: 'https://github.com/org/new_dep_A',
  block: [],
  fullBlock: [],
}); // Third example
```

This makes it very clear what's going on all the time, but requires the author to be aware of all of the installation options and to provide a value for each of them, which may or may not be a bad thing?

Finally, these three options are all compatible with each other (a multi-arg `add` could construct a `Dependency` value internally). I would personally only go with a combo of 2 and 3, though.

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,104 @@
...
+Do not block any standrad library:
+```ln
+add('https://github.com/org/new_dep_A', none(), false);
+```
+
+### Block at application level
+
+We might want to block any standard library we decide for every third party dependency or maybe have a custom behaviour for every third party library. This block would also need to work recursively due to module resolution priorities. The flow will be:
+
+![Block at app level](./app-lvl-v1.png)
```
dfellis, line 75: Why not put it in `./modules` directly? It should be blocking *its own* access to the standard libraries, right?

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,104 @@
...
+
+Block `@std/cmd` but do not override if any mock exists:
+```ln
+block('cmd', false);
+```
+
+Block `@std/cmd` and override mocks if any:
+```ln
+block('cmd', true);
+```
```
dfellis, line 88: I would invert this:

Libraries need to explicitly declare the `@std/something` dependencies they're using and *all others* get blacklisted.

Eg something like

```js
add('@std/http');
```

*Removes it* from a list of `@std` dependencies that will be mocked away.

dfellis: This way libraries and applications only get functionality they ask for by default so using a third party library doesn't require a massive `block` list every single time you install one to get a decently safe-by-default behavior.

rfcs/020 - Dependencies permissions RFC.md
aguillenv: On your first comment, yes, I had in mind only standard library blocks but yes there's no reason to limit it only to that.

On your second comment, completely agree on your suggestion, the API I present was quite confusing. It was a naive approach to express the flow and understand all the pieces.

I'll think about it but now I like the option of a `fn add(url: string): Dependency` and then calling `block` of `fullBlock` (not so sure about this name) over `Dependency`.

rfcs/020 - Dependencies permissions RFC.md
aguillenv: I think in terms of third party dependencies should exists inside `dependencies` directory. If the user wants to block its own access to the standard libraries it's another use case? That'd be something related to the behaviour of the app I'm writing instead of what I'm deciding the libraries I used have access to?

rfcs/020 - Dependencies permissions RFC.md
aguillenv: Hmm I think I get your idea, but taking into account that the blocks are based on mocks feels like apps would have a lot of "boilerplate code" even though they will be generated automatically for the user.

rfcs/020 - Dependencies permissions RFC.md
cdmistman: > So maybe this would be better:
> 
> ```js
> 'https://github.com/org/new_dep_A'.block('@std/cmd').add(); // Your first example
> 'https://github.com/org/new_dep_A'.block('@std/cmd').fullBlock('@std/tcp').add(); // Your second example
> 'https://github.com/org/new_dep_A'.block(['@std/cmd', '@std/tcp']).add(); // Another form of your second example
> 'https://github.com/org/new_dep_A'.add() // Your third example
> ```
> 

I personally think this is the best solution. The functions in `@std/deps` would look like:
```ts
fn add(url: string) = new Dependency {
  url: url,
  block: [],
  fullBlock: [],
}.add();
fn add(dep: Dependency) = adddep(dep);
fn block(dep: Dependency, block: string): Dependency {
  dep.block.push(block);
  return dep;
}
fn block(dep: Dependency, block: Array<string>): Dependency {
  dep.block = dep.block + block;
  return dep;
}
// repeat both `block` fns above but `s/block/fullBlock`
```
(btw, realizing that I don't believe I have proper detection on fields vs method calls in lnn right now, I'll have to fix that)

> 
> ### 3. Expose that `Dependency` type directly
> Something like:
> 
> ```js
> add(new Dependency {
>   url: 'https://github.com/org/new_dep_A',
>   block: ['@std/cmd'],
>   fullBlock: [],
> }); // First example
> 
> new Dependency {
>   url: 'https://github.com/org/new_dep_A',
>   block: [],
>   fullBlock: ['@std/cmd'],
> }.add(); // Second example
> 
> add(new Dependency {
>   url: 'https://github.com/org/new_dep_A',
>   block: [],
>   fullBlock: [],
> }); // Third example
> ```
> 
> This makes it very clear what's going on all the time, but requires the author to be aware of all of the installation options and to provide a value for each of them, which may or may not be a bad thing?

Most other dependency configuration tools don't require all fields to be populated, since the focus is on how easy it is to add dependencies. If users need more configuration options, then it's usually a minor change to add those configurations. For example, `Cargo.toml` values in the `dependencies` table can either be strings (indicating a semver-compatible string) or an object with required key `version` and optional keys `features`, `no-default-features`, `optional`, etc. A similar mechanism is present in npm, gemspec, and pubspec - some other popular package managers I looked at (cabal, stack, pip, opam) only support specifying the version (pip is weird and can do more, but not via table-types)

rfcs/020 - Dependencies permissions RFC.md
cdmistman: yes, I think it makes sense for users to explicitly mock their own `std` deps but for dependencies' mocks to be handled in `.dependencies.ln`

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -51,21 +51,70 @@ This is the basic case when we want to use `new_dep_A` in our project but we do
...
+
+fn fullBlock(dep: Dependency, block: string): Dependency {
+  dep.fullBlock.push(fullBlock);
+  return dep;
+}
+
+fn fullBlock(dep: Dependency, block: Array<string>): Dependency {
+  dep.fullBlock = dep.fullBlock + block;
+  return dep;
+}
```
cdmistman, line 93: maybe instead of having the empty `add()` which serves no use besides constructing the `Dependency`, we can have `fn block(block: string) = new Dependency { ... }` and the same for `fullBlock`?

rfcs/020 - Dependencies permissions RFC.md
cdmistman: agreed - I prefer opt-out of `@std` with the ability to override from a consuming project

rfcs/020 - Dependencies permissions RFC.md
aguillenv: Yeah probably thats more meaningful

cdmistman: likewise, there should be an equivalent `fn block(block: Array<string>)` etc

cdmistman (commented):
I think this is a very solid API

cdmistman (approved):
I think this is a very solid API

cdmistman:
er, ignore that comment, i thought i could delete it

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -51,21 +51,70 @@ This is the basic case when we want to use `new_dep_A` in our project but we do
...
+  url: url,
+  block: [],
+  fullBlock: [],
+};
+
+fn add() = new Dependency {
+  url: '',
+  block: [],
+  fullBlock: [],
+};
```
dfellis: This is missing the `add(dep: Dependency)` function that would act on the dependency object and actually do the work. Eg, if they want to list the things they're blocking first and then the library at the end.

rfcs/020 - Dependencies permissions RFC.md
dfellis: I don't agree with this. If a library can have access to the full standard library without explicitly declaring it, how will *consumers* of that library know which standard library dependencies are in use that they may want to block?

rfcs/020 - Dependencies permissions RFC.md
aguillenv: This is an outdated version, after Colton's comment on Friday.

rfcs/020 - Dependencies permissions RFC.md
aguillenv: Hmm thats true, the consumers might not be blocking what the library uses and they do not want. I've rephrased the proposal and add an extra argument taking this into account.

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,168 @@
...
+  // Download and install each dep
+  // Apply blocks defined for each dependency
+  // Block/Enable (based on what we finally decide) depenencies at application level, meaning that mocks will exists at /dependencies/modules/
+}
+```
+The `.dependencies.ln` file could look something like:
+
+```ts
+const using = [
+  use('@std/tcp');
```
dfellis, line 135:
```suggestion
  use('@std/tcp'),
```

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,168 @@
...
+  // Block `@std/cmd` and override mocks if any:
+  add('https://github.com/org/new_dep_B').fullBlock('@std/cmd'),
+  // Do not block any standrad library:
+  add('https://github.com/org/new_dep_C'),
+];
+const globalDepsBlock = [
+  block('@std/fs'),
+  fullBlock('@std/exec'),
+];
+commit(using, dependencies, globalDepsBlock);
```
dfellis, line 149: Another idea here:

```js
using([
  '@std/tcp',
]).dependencies([
  add('https://github.com/org/new_dep_A').block('@std/cmd'),
  add('https://github.com/org/new_dep_B').fullBlock('@std/cmd'),
  add('https://github.com/org/new_dep_C'),
]).globalDepsBlock([
  block('@std/fs'),
  fullBlock('@std/exec'),
]).commit();
```

Or maybe:

```js
module()
  .using('@std/tcp')
  .dependency('https://github.com/org/new_dep_A')
    .block('@std/cmd')
    .add()
  .dependency('https://github.com/org/new_dep_B')
    .fullBlock('@std/cmd')
    .add()
  .dependency('https://github.com/org/new_dep_C')
    .add()
  .globalBlock('@std/fs')
  .globalFullBlock('@std/exec')
  .commit();
```

dfellis: I don't know about the `global` naming, either. You could potentially stick to `block` and `fullBlock` and it applying to this library/module versus a dependency depends on whether it's attached to a dependency declaration or the library/module declaration.

aguillenv: Agree, add that name to distinguish function parameters but using an API ad the proposed one makes sense to just apply to the module or dependency the block/fullBlock functions.

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,193 @@
...
+  dependencies: Array<Dependency>,
+  block: Array<string>,
+  fullBlock: Array<string>,
+}
+
+type Dependency {
+  url: string, // Empty string or ignore it for global blocks at application level
+  block: Array<string>,
+  fullBlock: Array<string>,  // fullBlock, deepBlock, extensiveBlock, exhaustiveBlock?
+  package: Package,
```
dfellis, line 74: Why does this property exist?

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,193 @@
...
+  .use(['@std/tcp'])
+  // Block `@std/cmd` but do not override if any mock exists:
+  .dependency('https://github.com/org/new_dep_A')
+    .block('@std/cmd')
+    .add()
+  // Block `@std/cmd` and override mocks if any:
+  .dependency('https://github.com/org/new_dep_B')
+    .fullBlock('@std/cmd')
+    .add()
+  // Do not block any standrad library:
```
dfellis, line 169:
```suggestion
  // Do not block any standard library:
```

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,193 @@
...
+  .dependency('https://github.com/org/new_dep_C')
+    .add()
+  .block('@std/fs')
+  .fullBlock(['@std/exec'])
```
dfellis, line 173: Can we include a comment above this section indicating that these are the explicitly blocked standard library modules for all dependencies? (Since `use` effectively defines the inverse of modules to insert blocking mocks.)

rfcs/020 - Dependencies permissions RFC.md
aguillenv: In order to distinguish `block` from `Dependency` and `Package`. When returning `Dependency` we need a way to keep track of the `Package` object.

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,193 @@
...
+fn package(): Package {
+  return new Package {
+    std: [],
+    dependencies: [],
+    block: [],
+    fullBlock: [],
+  };
+}
+
+fn use(pkg: Package, std: [string]): Package {
```
cdmistman, line 88: nit: `Array<string>` not `[string]`

cdmistman (approved):

rfcs/020 - Dependencies permissions RFC.md
dfellis: Got it.

dfellis (approved):

rfcs/020 - Dependencies permissions RFC.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Current Status
+
+### Proposed
+
+2021-08-05
+
+### Accepted
+
+YYYY-MM-DD
```
dfellis, line 11: Be sure to update this date before you merge this PR

PR #610, opened 2021-08-12
Dependencies permissions

aguillenv:


std/deps.ln
```diff
@@ -35,9 +35,9 @@ type PackageDependency {
 export event install: RawPackage
 
 fn package(): RawPackage {
-  const emptyArrStr = new Array<string> [];
-  const emptyArrDep = new Array<Dependency> [];
-  const pkg = new RawPackage {
+  let emptyArrStr = new Array<string> [];
+  let emptyArrDep = new Array<Dependency> [];
+  let pkg = new RawPackage {
```
dfellis, line 40: .

std/deps.ln
```diff
@@ -1,34 +1,331 @@
...
+  block: Array<string>,
+  fullBlock: Array<string>,
+}
+
+type PackageDependency {
+  pkg: RawPackage,
+  dep: Dependency,
+}
+
+// Exporting type as alias since could not export the type and used it in this file at the same time
```
dfellis, line 31: That's a weird bug we should tackle. I'm sure it's because we do exports last so the inner functions wouldn't have seen the type.

Does `type Package { ... }` and then `export Package` after it not work?

std/deps.ln
```diff
@@ -1,34 +1,331 @@
...
   emit exit 0;
 }
 
+fn getBlacklist(std: Array<string>): Array<string> {
+  // TODO: add @std/deps  to this list and handle it on bdd tests
+  const stdList = [
+    '@std/app', '@std/avmdaemon', '@std/cmd', '@std/datastore', '@std/http', 
+    '@std/httpcommon', '@std/httpserver', '@std/json', '@std/seq', '@std/tcp',
+    '@std/tcpserver', '@std/trig',
+  ];
```
dfellis, line 159: Eventually we'll want more "precise" mocks that export the same functions and types but don't actually do anything. At that point, the listing of those mocks would probably be able to be this array, rather than being hardwired. :)

std/deps.ln
```diff
@@ -1,34 +1,331 @@
...
+  return stdList.filter(fn (a: string) = !(pkg.std.find(fn (b: string) = eq(a,b)).isOk()));
```
dfellis, line 160: Two questions:

1. Do you need the `()` around the `!(pkg.std.find(...` part? `!pkg.std.find(...` doesn't work? Or did you find that to look confusing?
2. `fn (b: string) = eq(a,b)` instead of `fn (b: string) = a == b` was due to preference or did the latter not work?

std/deps.ln
```diff
@@ -1,34 +1,331 @@
...
+    makeDirectory(dependencies);
+  }
+  const modulesPath = [dependencies, 'modules'].join('/');
+  if !directoryExists(modulesPath) {
+    makeDirectory(modulesPath);
+  }
+  const stdPath = [modulesPath, 'std'].join('/');
+  if !directoryExists(stdPath) {
+    makeDirectory(stdPath);
+  }
```
dfellis, line 175: This pattern is used enough and looks so much like `mkdir -p` that maybe it should be refactored to something like:

```
fn mkdirp(dir: string) {
  if !directoryExists(dir) {
    makeDirectory(dir);
  }
}

...

mkdirp('dependencies');
mkdirp('dependencies/modules');
mkdirp('dependencies/modules/std');
```

std/deps.ln
aguillenv: 1) I will check because I remember rewriting this line several times while testing it.
2) Preference at that time to have everything with fns, I think I didn't even try the `==`

std/deps.ln
aguillenv: Should I add that to the `TODO` comment?

std/deps.ln
aguillenv: definitely! forgot to refactor it

std/deps.ln
aguillenv: Update: both worked, so removed the `()` for the first one and updated to `==` the second one

std/deps.ln
dfellis: Maybe?

dfellis: Not gonna block on this.

std/deps.ln
```diff
@@ -1,34 +1,333 @@
...
+
+export fn fullBlock(pkgd: PackageDependency, block: string): PackageDependency {
+  pkgd.dep.fullBlock.push(block);
+  return pkgd;
+}
+
+export fn fullBlock(pkgd: PackageDependency, block: Array<string>): PackageDependency {
+  pkgd.dep.fullBlock = pkgd.dep.fullBlock + block;
+  return pkgd;
+}
```
depombo, line 132: Maybe instead of `block` and `fullBlock` we could do `blockWithMocks` and `block` respectively? It seems more evident with naming like that?

aguillenv: it is always with mocks, the difference is if it overrides/force them or not. Another suggestion with this context?

depombo: ah nvm, hmmm then maybe as-is works well 👍

dfellis (approved):

PR #611, opened 2021-08-13
Datastore Restructuring RFC

dfellis:


rfcs/021 - Datastore Restructuring RFC.md
```diff
@@ -0,0 +1,82 @@
...
+
+It may also make sense for the storage layer to get the execution context, as well; then no transferrance of the key in question between OS threads would be necessary and since blocking mutations like that prevent other mutations on the same key from happening simultaneously, that might be a better use of CPU time on these OS threads, as well. But mixing Tokio async/await with real OS threads may be difficult/impossible, or may still require spawning a task with an `async move` block inside, which would nullify the benefits while complicating the storage layer, so we should be careful with this. Likely we don't do something like this in the first pass to make the layers cleaner and more easily extensible, and we come back to it later as an optimization pass when we're more convinced that little will change with datastore's API and the increased maintenance burden is worth it.
```
cdmistman, line 61: > or may still require spawning a task with an `async move` block inside, which would nullify the benefits while complicating the storage layer

I'm not sure how the benefits would be nullified? In fact, I think it's better for performance overall without impacting the API at all - the fewer OS threads there are, the less likely any OS thread running the AVM will be unscheduled. Being able to schedule the tokio task on any currently-running OS thread while some other AVM task is running an IO-bound opcode also prevents other OS threads from hogging CPU resources, preventing the hashmap thread from getting run until some other thread yields CPU time.

If this were implemented using OS threads, it would look like:
```rs
thread::spawn(move {
  let mut data = HashMap::new();
  loop {
    if let Ok(req) = read_requests.try_recv() {
      ...
    }
    if let Ok(req) = write_requests.try_recv() {
      ...
    }
    thread::yield_now();
  }
}
```
whereas with tokio, you could do:
```rs
tokio::spawn(async move {
  let mut data = HashMap::new();
  loop {
    // note that `tokio::select` works similarly to rust's `match`, and the
    // requests can be handled directly inside of the `select` instead of
    // returning `(Option<ReadRequest>, Option<WriteRequest>)`
    let (read_req, write_req) = tokio::select! {
      req = read_requests.recv() => (Some(req.unwrap()), None),
      req = write_requests.recv() => (None, Some(req.unwrap())),
    };
    if let Some(read_req) = read_req {
      ...
    } else if let Some(write_req) = write_req {
      ...
    }
  }
});
```
Excluding the `thread::yield_now()` in the OS thread version would result in a hotloop that would infinitely consume CPU resources - this automatically gets handled when `.poll()`ing the `recv()` futures in the tokio version, and doesn't happen until tokio tries several times to receive from each of the channels.

dfellis: So most CPUs are [NUMA](https://en.wikipedia.org/wiki/Non-uniform_memory_access) at least at the CPU cache layer. If the data that was in that hashmap was in cache for the CPU core the OS thread was tied to, then the `async move` for a tokio task is gonna require whichever other thread on whichever other CPU core to instead reacquire the data from RAM where the addressing is uniform again.

If we can make sure that the particular compute task using that data actually executes in a blocking manner on the OS thread that houses the data, we *aren't* "hogging CPU resources", they're being used to actually do the relevant work and immediately stored the result back where it belongs, and then immediately polling the channels for work and executing it, or sleeping for 100ns or something and trying again would be *less* wasted time than using the Tokio framework here.

cdmistman: The `async move` only happens once per hashmap though. If you're worried about task affinity, <https://docs.rs/tokio/1.10.0/tokio/task/fn.spawn_local.html> can be used in conjunction with <https://docs.rs/tokio/1.10.0/tokio/runtime/struct.Builder.html#method.on_thread_start> to start a hashmap task *per OS thread*.

dfellis: I think we're talking over each other? I'm not referring to the `async move` on the initialization of the thread (though the task affinity piece is nice to know and would let us use tokio tasks here, too); I was referring to the idea that the datastore remote execution opcodes could run "in place" or if the data is copied into a tokio task and executed elsewhere and then brought back. if my `dsmrun` opcode call just happened in the same thread as where the cached data is located, and the inner opcode calls were all `cpu!` opcodes, then we skip a *ton* of scheduling overhead and memcpys.

rfcs/021 - Datastore Restructuring RFC.md
```diff
@@ -0,0 +1,82 @@
...
+
+![Datastore arch](./datastore_arch.jpg)
+
+The architecture and flow for the proposed datastore is diagrammed above. The routing layer (whether local mode or remote) is responsible for mutating the HandlerMemory for the given opcode and `opcode.rs` is simply a shell. It chooses the correct node to direct the work to and triggers its counterpart on the other node to request the relevant data from the storage layer, which then eventually returns it. The routing layer hands all of this to the execution layer on the same node to perform the requested work, which then returns it back to the routing layer on that node. The routing layer then stores the mutation, if any, back into the storage layer and then passes the result back to its counterpart on the original node, which then hands the results back to the original datastore opcode in the form of a new HandlerMemory object that replaces the old one (which may simply be the original HandlerMemory object in the case of `mutOnly` calls).
```
aguillenv, line 65: Do you see an eventual bottleneck with this architecture with the routing layer?

rfcs/021 - Datastore Restructuring RFC.md
```diff
@@ -0,0 +1,82 @@
...
+
+By putting the routing layer as the centerpiece, we can add on greater intelligence to datastore on where it places data in the future. One obvious optimization is to use the DNS regional load balancing that chooses which datacenter to route data to as a "strong hint" on where the primary node should be located to minimize latency for the end user, and we can also go from a fixed replica count to having a replica per datacenter to make eventually consistent reads fast whereever in the world it is located. The AVM gets enough metadata from its DNS-based configuration to do both of these, but in the current structure of the codebase it is very difficult to tackle.
+
+### Alternatives Considered
+
+The primary alternative considered is to just drop `Dashmap` and use a `RWLock`-wrapped `Hashmap` singleton. This would resolve the immediate problem of mutating opcodes not waiting for a proper lock before executing, but would also serialize all datastore accesses on a node, which would be a significant performance regression. It may still be worthwhile to do in the interim if prioritization of this RFC is lower.
```
aguillenv, line 71: In my opinion if we are going to the "distributed data" path we should tackle this with the proposed architecture right away. Otherwise would be difficult to handle any change.

depombo (approved):

rfcs/021 - Datastore Restructuring RFC.md
dfellis: This part shouldn't be the bottleneck. There is a potential bottleneck with the way the datastore KV internals are structured, but the routing layer shouldn't be a problem.

PR #612, opened 2021-08-13
[lnn] generics

cdmistman:


compiler/src/lnntoamm/Expr.ts
```diff
@@ -1,3 +1,4 @@
+import { stdout } from 'process';
```
dfellis, line 1: Nit: You don't need this (the relevant lines are commented out.

```suggestion
```

compiler/src/lnntoamm/Expr.ts
```diff
@@ -504,11 +524,16 @@ class Call extends Expr {
...
     super(ast);
     if (fns.length === 0 && maybeClosure === null) {
       throw new Error(`no function possibilities provided for ${ast}`);
     }
+    // if (this.dbg()) {
+    //   console.dir('-> created', this.ast.t.trim());
+    //   console.dir(fns, { depth: 4 });
+    //   console.dir(args, { depth: 4 });
+    //   console.dir(retTy, { depth: 4 });
+    // }
```
dfellis, line 536:
```suggestion
```

compiler/src/lnntoamm/Expr.ts
```diff
@@ -565,80 +590,62 @@ class Call extends Expr {
...
+    const [selFns, selPTys, selRetTys] = FunctionType.matrixSelect(
+      fns,
+      argTys,
+      retTy,
+      metadata.scope,
+    );
+    fns = selFns;
+    // console.log('initial select for', ast.t.trim(), '->', fns);
+    // stdout.write('-> args ');
+    // console.dir(args, { depth: 4 });
```
dfellis, line 603:
```suggestion
```

compiler/src/lnntoamm/Expr.ts
```diff
@@ -565,80 +590,62 @@ class Call extends Expr {
...
+    args.forEach((arg, ii) =>
+      arg.ty.constrain(Type.oneOf(selPTys[ii]), metadata.scope),
+    );
```
dfellis, line 606: Any reason why this isn't operating on the `argTys` array instead?

This seems equivalent and more like the prior code you were writing:

```suggestion
    argTys.forEach((ty, ii) =>
      ty.constrain(Type.oneOf(selPTys[ii]), metadata.scope)
    );
```

compiler/src/lnntoamm/Expr.ts
```diff
@@ -565,80 +590,62 @@ class Call extends Expr {
...
-      // console.log('constraining', ty, 'to', paramTys);
-      ty.constrain(Type.oneOf(paramTys), metadata.scope);
-      // console.log('constrained:', ty);
-    });
+    // argTys.forEach((ty, ii) => {
+    //   const paramTys = (fns as Fn[]).map((fn) => fn.params[ii].ty);
+    //   // console.log('constraining', ty, 'to', paramTys);
+    //   ty.constrain(Type.oneOf(paramTys), metadata.scope);
+    //   // console.log('constrained:', ty);
+    // });
```
dfellis, line 620:
```suggestion
```

compiler/src/lnntoamm/Expr.ts
```diff
@@ -565,80 +590,62 @@ class Call extends Expr {
...
+    );
+    fns = selFns;
+    // console.log('initial select for', ast.t.trim(), '->', fns);
+    // stdout.write('-> args ');
+    // console.dir(args, { depth: 4 });
+    args.forEach((arg, ii) =>
+      arg.ty.constrain(Type.oneOf(selPTys[ii]), metadata.scope),
+    );
+    retTy.constrain(Type.oneOf(selRetTys), metadata.scope);
     // now, constrain all of the args to their possible types
```
dfellis, line 608: Is this comment still relevant?

compiler/src/lnntoamm/Expr.ts
```diff
@@ -565,80 +590,62 @@ class Call extends Expr {
...
+      this.retTy,
       this.scope,
-    ).reduce(
-      ([fns, tys], [fn, ty]) => [
-        [...fns, fn],
-        [...tys, ty],
-      ],
-      [new Array<Fn>(), new Array<Type>()],
     );
+    // console.log('fnSelect for', this.ast.t.trim(), '->', ret[0]);
```
dfellis, line 634:
```suggestion
```

compiler/src/lnntoamm/Fn.ts
```diff
@@ -1,11 +1,12 @@
+import { stdout } from 'process';
```
dfellis, line 1:
```suggestion
```

compiler/src/lnntoamm/Fn.ts
```diff
@@ -76,25 +77,25 @@ export default class Fn {
...
-        // eg the generic identity function. Or just wait until generic
-        // fn type parameters.
-        throw new Error(`type erasure is illegal`);
-      }
+      // if (retTy.dup() !== null) {
+      //   // TODO: figure out how to prevent type erasure while allowing
+      //   // eg the generic identity function. Or just wait until generic
+      //   // fn type parameters.
+      //   throw new Error(`type erasure is illegal`);
+      // }
```
dfellis, line 94: Is this still relevant?

```suggestion
```

compiler/src/lnntoamm/Fn.ts
```diff
@@ -123,13 +126,25 @@ export default class Fn {
...
+    // if (name === 'none') {
+    //   console.log('~~~~~~~~ created', name);
+    //   stdout.write('params: ');
+    //   console.dir(params, { depth: 4 });
+    //   stdout.write('retTy: ');
+    //   console.dir(retTy, { depth: 4 });
+    //   stdout.write('body: ');
+    //   console.dir(body, { depth: 4 });
+    // }
+
```
dfellis, line 147:
```suggestion
```

compiler/src/lnntoamm/Fn.ts
```diff
@@ -219,22 +247,43 @@ export default class Fn {
...
+    let res: [Type[], Type] | null = null;
     try {
-      this.params.forEach((param, ii) =>
-        param.ty.tempConstrain(argTys[ii], scope),
-      );
-      res = this.retTy.instance();
-    } catch (_e) {
+      this.params.forEach((param, ii) => {
+        return param.ty.tempConstrain(argTys[ii], scope);
+      });
```
dfellis, line 263: Nit: Single-line functions like these have tended to go without the curly braces, so I am adding a suggested change to make it more consistent with the rest of the code in here.

```suggestion
      this.params.forEach((param, ii) => param.ty.tempConstrain(argTys[ii], scope));
```

compiler/src/lnntoamm/Fn.ts
```diff
@@ -219,22 +247,43 @@ export default class Fn {
...
       // we return null (`res` is already `null`) and we need to
       // ensure the param tys have `resetTemp` called on them.
+      // However, if the Error message starts with `TODO`, then
+      // print the error since it's for debugging purposes
+      const msg = e.message as string;
+      if (msg.startsWith('TODO')) {
+        console.log(
+          'warning: came across TODO from Types.ts when getting result ty for a function:',
+        );
+        console.group();
```
dfellis, line 280: Should this be above the other `console.log` to make sure they don't get split from each other?

compiler/src/lnntoamm/Types.ts
```diff
@@ -39,15 +66,67 @@ export default abstract class Type implements Equalable {
...
+  static getFromTypename(
+    name: LPNode | string,
+    scope: Scope,
+    dupOpts?: DupOpts,
+  ): Type {
+    if (typeof name === 'string') {
+      name = fulltypenameAstFromString(name);
+    }
+    const parsed = parseFulltypename(name);
+    const solidify = ([name, generics]: TypeName): Type => {
```
dfellis, line 85: Nit: This doesn't need to be a closure.

compiler/src/lnntoamm/Types.ts
```diff
@@ -123,127 +193,279 @@ export default abstract class Type implements Equalable {
...
-    } else if (ty instanceof Interface) {
-      // FIXME: once builtins can have fields
+  compatibleWithConstraint(that: Type, scope: Scope): boolean {
+    if (this.eq(that)) {
+      return true;
+    }
+    if (that instanceof Opaque) {
+      const thisGens = Object.values(this.generics);
+      const thatGens = Object.values(that.generics);
+      if (this.name !== that.name || thisGens.length !== thatGens.length) {
```
dfellis, line 230: Is the first half of this conditional true? Wouldn't `Array<any>` be compatible with `Array<int64>` but not have the same name?

compiler/src/lnntoamm/Types.ts
```diff
@@ -123,127 +193,279 @@ export default abstract class Type implements Equalable {
...
+            return true;
+          } else {
+            return thisGen.compatibleWithConstraint(thatGen, scope);
+          }
+        })
       );
+    } else if (that instanceof Interface || that instanceof OneOf) {
+      return that.compatibleWithConstraint(this, scope);
+    } else if (that instanceof HasField) {
+      return false;
```
dfellis, line 248: Is this one of the ones that should be labeled "TODO: Requires struct types"?

compiler/src/lnntoamm/Types.ts
```diff
@@ -123,127 +193,279 @@ export default abstract class Type implements Equalable {
...
+      })
+    );
   }
 
-  fieldIndices(): FieldIndices {
-    return TODO("determine if it's even worth keeping the Builtin class");
+  fnselectOptions(): Type[] {
+    const genOptions = Object.values(this.generics).map((g) =>
+      g === null ? [g] : g.fnselectOptions(),
+    );
```
dfellis, line 342:
```suggestion
    const genOptions = Object.values(this.generics).map((g) => g?.fnselectOptions() ?? [g]);
```

compiler/src/lnntoamm/Types.ts
```diff
@@ -123,127 +193,279 @@ export default abstract class Type implements Equalable {
...
+    const opts = new Array<Type>();
+    const getIndices = matrixIndices(genOptions);
+    const toSolidify = new Opaque(this.name, Object.keys(this.generics));
+    for (
+      let indicesRes = getIndices.next();
+      !indicesRes.done;
+      indicesRes = getIndices.next()
+    ) {
```
dfellis, line 350: This looks like an iterator, couldn't it be refactored to:

```suggestion
    for (let indicesRes of getIndices) {
```

compiler/src/lnntoamm/Types.ts
```diff
@@ -123,127 +193,279 @@ export default abstract class Type implements Equalable {
...
-      case 'float64':
-      case 'float32':
-      case 'bool':
-      case 'void':
-        return true;
-      default:
+      case 'string':
         return false;
+      default:
+        return true;
```
dfellis, line 386: Can you revert this change? It's why `Result<int64>` got `reff` instead of `refv`.

compiler/src/lnntoamm/Types.ts
```diff
@@ -260,16 +482,66 @@ export class FunctionType extends Type {
     this.retTy = retTy;
   }
 
-  static matrixSelect(fns: Fn[], args: Type[], scope: Scope): [Fn, Type][] {
-    const originalLength = fns.length;
+  /*
+  Original design comment (was written as a FIXME above `Expr.Call.inline`):
```
dfellis, line 486: Do we still need this comment?

Does it still depend on the order of `root.lnn`?

compiler/src/lnntoamm/Types.ts
```diff
@@ -260,16 +482,66 @@ export class FunctionType extends Type {
...
+    isDbg && stdout.write('callTy: ') && console.dir(callTy, { depth: 4 });
     fns = fns.filter((fn) => fn.ty.compatibleWithConstraint(callTy, scope));
+    isDbg && console.log('filtered:', fns);
     // if it's 0-arity then all we have to do is grab the retTy of the fn
     if (args.length === 0) {
+      isDbg && console.log('nothin');
       return fns.reduce(
-        (fns, fn) => [...fns, [fn, fn.retTy.instance()]],
-        new Array<[Fn, Type]>(),
+        // ([fns, _pTys, retTys], fn) => [[]...fns, [fn, fn.params.map((p) => p.ty.instance()), fn.retTy.instance()]],
```
dfellis, line 544:
```suggestion
```

compiler/src/lnntoamm/Types.ts
```diff
@@ -278,84 +550,127 @@ export class FunctionType extends Type {
...
-    const indices = matrix.map(() => 0);
+    const fnsByWeight = new Map<number, [Fn, Type[], Type][]>();
+    const getIndices = matrixIndices(matrix);
     // keep it as for instead of while for debugging reasons
-    for (let i = 0; ; i++) {
+    for (
+      let indicesRes = getIndices.next();
+      !indicesRes.done;
+      indicesRes = getIndices.next()
+    ) {
```
dfellis, line 578: Same comment here, this looks iterable, and I think this would work:

```suggestion
    for (let indicesRes of getIndices) {
```

compiler/src/lnntoamm/Expr.ts
cdmistman: this is useful for debugging so I was planning on keeping it

compiler/src/lnntoamm/Types.ts
```diff
@@ -278,84 +550,127 @@ export class FunctionType extends Type {
...
+    //   } else {
+    //     fns.push([fn, pTys, retTy]);
+    //   }
+    // });
+    // weightFns = weightFns.filter(
+    //   ([weightedFn, _retTy]) =>
+    //     fns.findIndex(([fn, _retTy]) => fn === weightedFn) === -1,
+    // );
+    // isDbg && console.log('filtered:', weightFns);
+    // return [...fns, ...weightFns];
```
dfellis, line 622:
```suggestion
```

compiler/src/lnntoamm/Types.ts
```diff
@@ -278,84 +550,127 @@ export class FunctionType extends Type {
...
+        fnsByWeight
+          .get(weight)
+          .forEach(([weightedFn, weightedPTys, weightedRetTy]) => {
+            const alreadyIdx = fns.findIndex((fn) => fn === weightedFn);
+            if (alreadyIdx === -1) {
+              // push to the end of the fns since technically the weight for the function
+              // is indeed higher than what it was before
+              fns.push(weightedFn);
+            } else {
+              fns.push(fns.splice(alreadyIdx, 1)[0]);
```
dfellis, line 634: What the heck is this doing?

compiler/src/lnntoamm/Expr.ts
cdmistman: i simply forgot i had it - fixing

compiler/src/lnntoamm/Types.ts
```diff
@@ -278,84 +550,127 @@ export class FunctionType extends Type {
...
+      } else {
+        throw new Error('somehow got more options when fn selecting');
+      }
     }
+    // const getStack = { stack: '' };
+    // Error.captureStackTrace(getStack);
+    isDbg &&
+      (() => {
+        //   console.log('returning from', getStack.stack);
+        console.dir(ret, { depth: 4 });
```
dfellis, line 666: Why is this in an IIFE?

compiler/src/lnntoamm/Expr.ts
cdmistman: yes, it's just in the wrong spot. fixing

compiler/src/lnntoamm/Types.ts
```diff
@@ -837,88 +1194,150 @@ class Interface extends Type {
...
+      if (that instanceof HasField && !Has.field(that, toCheck)) {
+        throw new Error(`${errorBase} field ${that.name}`);
+      } else if (
+        that instanceof HasOperator &&
+        Has.operator(that, scope, toCheck).length !== 0
+      ) {
+        const opString =
+          that.params.length === 1
+            ? `${that.name} ${that.params[0].ammName}`
+            : `${that.params[0].ammName} ${that.name} ${that.params[1].ammName}`;
```
dfellis, line 1278: This is technically assuming only two parameters

compiler/src/lnntoamm/Expr.ts
cdmistman: this is also useful, so i was planning on keeping it

compiler/src/lnntoamm/Fn.ts
cdmistman: Since type erasure isn't desired, it might be useful to keep a comment summarizing this block. I'll go ahead and leave a TODO for type erasure

compiler/src/lnntoamm/util.ts
```diff
@@ -2,7 +2,9 @@ import { v4 as uuid } from 'uuid';
 import Fn from './Fn';
 import Operator from './Operator';
 
-export const genName = () => '_' + uuid().replace(/-/g, '_');
+let counter = 0;
+export const genName = (name?: string) =>
+  name ? `${name}-n${counter++}` : '_' + uuid().replace(/-/g, '_');
```
dfellis, line 7: Why is both a counter an a UUID necessary for uniqueness?

compiler/src/lnntoamm/Fn.ts
cdmistman: yeah, i had to do a block at one point for debugging purposes, but i'll change it back

compiler/src/lnntoamm/Fn.ts
cdmistman: I use the group to indicate that the next line is associated to the warning. for example:
```
warning: came across TODO from Types.ts when getting result ty for a function:
  TODO: re-tempConstrain Interface
```

compiler/src/lnntoamm/Types.ts
cdmistman: the name would be `Array` in both cases, whereas the `ammName` would be `Array<any>` or `Array<int64>`

compiler/src/lnntoamm/Types.ts
cdmistman: Opaque types by definition don't have any accessible fields - built-in types that expose fields should still use the `Struct` type

compiler/src/lnntoamm/Types.ts
cdmistman: > `g?.fnselectOptions()`

does this work? I can't find any information on the `?` operator when it's not used with `:`

compiler/src/lnntoamm/Types.ts
dfellis: Got it. I called that `baseName` and your `ammName` would be the `typename`.

compiler/src/lnntoamm/Types.ts
cdmistman: it's a function generator - there might be a more convenient way to do this but I don't believe `let of` works with generators

compiler/src/lnntoamm/Types.ts
cdmistman: Yes - I noted specifically that it's because of the ordering of `root.lnn` but realized that it's applicable to *all* code written in Alan, since the definition order of functions is important in determining selection and users may not realize that they need to change the ordering of their function declaration. I'll update the comment to reflect that

compiler/src/lnntoamm/Types.ts
cdmistman: ah, `for of` does work with generators <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#iterating_over_generators>

compiler/src/lnntoamm/Types.ts
cdmistman: it's moving the function from the position with the lower weight to the position with the higher weight. the comment for the other branch should be for the `else` branch

compiler/src/lnntoamm/Types.ts
cdmistman: because otherwise i'd have to write a bunch of `isDbg && ...` when debugging

cdmistman (commented):
ignore this comment/review, github broke and won't let me cancel????

compiler/src/lnntoamm/util.ts
cdmistman: if `name` isn't `undefined`, then that's because it needs to be easier to read. if it is `undefined`, then i use uuid to make easier-to-read names and uuid names easier to separate and optically parse

compiler/src/lnntoamm/Types.ts
dfellis: It's not the `?` operator, it's the [`?.` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

PR #614, opened 2021-12-07
Test out building a debian package for better integration in Ubuntu

dfellis:


Makefile
```diff
@@ -34,8 +34,6 @@ ALAN_STD_FILES=$(wildcard std/*.ln) $(wildcard std/*.lnn)
 	yarn add pkg
 	cd compiler && ../node_modules/.bin/pkg --targets host .
 
-# allow cargo to determine when it's necessary to rebuild
-.PHONY: ./avm/target/release/alan
```
dfellis, line 38: This interferes with `checkinstall` because `rustc` itself has problems with the fake filesystem used to track what files are put where by `make install`. Removing this bit prevents it from being run if `make` is run manually before `make install`.

PR #631, opened 2023-11-10
Get the AVM compiling with current openssl

dfellis:


dfellis:
Shoot, the max version of `ntapi` I can upgrade to still has that [unaligned access bug](https://github.com/alantech/alan/actions/runs/6827270639/job/18569023099?pr=631#step:4:380).

Reviving on Windows is a lower priority for me, and it looks like all uses of `heim` will need to be replaced with something still maintained, so that's a larger project that I'm gonna come back to later.

PR #632, opened 2023-11-10
Fix autopublish

dfellis:


dfellis:
This can't be tested without merging, unfortunately, so here we go.

PR #656, opened 2024-02-21
Alan to Rust transpiler code

dfellis:
This branch has the work on a more "real" transpiler. Feeling out the responsibilities between the lntors, parse, and program sections of the compiler. I'm going to try to make lntors as "dumb" as possible and move things like type and function resolution into the program so it should be more easily reusable for the eventual shader transpiler logic. Also thinking that I may want to revive the js transpile path that was in Alan 0.1, initially to use as the interpreter path in the Alan CLI, but also potentially useful for the browser path if the WASM transpile path with WGPU is too finicky, and *especially* useful for a demo website like the current alan-lang.org where the compiler is running in the browser generating code the browser can run directly. I don't know if you can dynamically generate WASM code and then load it, so that would be an escape hatch if it's not possible.

- **First step in making a more real transpiler: establish a subdirectory for all of the soon-to-be source files**
- **Run cargo fmt**
- **Add logic to properly load the standard library from embedded strings in the compiler (and prove it by parsing the root scope on evey compilation**
- **Move the function generation to a separate file**
- **Move the old root scope to an unused filename and create a new root scope with a single entry for now**
- **Fix formatting**
- **Update TODOs in the resolve functions to make it clearer which way they're going to evolve**
- **Add super simple 'to-rs' sub-command to get the generated Rust code as a file. Good for debugging**


dfellis:
There are things here that need to be addressed, but I want to get this merged to main now and tackle those in other PRs before this gets too far out of hand and difficult to follow in the future.

The move to microstatements and generating code from that has made the recursive-decent code generation I needed available. I will try to tackle the type info stuff in another PR because it'll probably be really tricky.

PR #658, opened 2024-02-26
Get the ExitCode type working and use it on two tests

dfellis:


src/lntors/function.rs
```diff
@@ -18,11 +18,8 @@ pub fn from_microstatement(
         .to_string()),
         Microstatement::Value { representation, .. } => Ok(representation.clone()),
         Microstatement::FnCall { function, args } => {
-            // TODO: Add logic to get the type from the args array of microstatements. For the sake
-            // of keeping the hello world test working for now, adding some magic knowledge that
-            // should not be hardcoded until the microstatement generation adds the type
-            // information needed.
-            match program.resolve_function(scope, function, &vec!["String".to_string()]) {
+            let arg_types = args.iter().map(|arg| arg.get_type(scope, program)).collect();
+            match program.resolve_function(scope, function, &arg_types) {
```
dfellis, line 22: This recursive resolution of function return types should work as long as every resolved function *eventually* is composed of functions with concrete types. So, basically as long as we have concrete types in the standard library, user-written functions can build up on them and get type inference for "free."

src/lntors/mod.rs
```diff
@@ -5,6 +5,8 @@ mod function;
 mod typen;
 
 pub fn lntors(entry_file: String) -> Result<String, Box<dyn std::error::Error>> {
+    // TODO: Figure out a better way to include custom Rust functions that we may then bind
+    let preamble = include_str!("../std/root.rs").to_string();
```
dfellis, line 9: This one is a big TODO so I want to think a bit about options here. It won't block this PR, but it will block the first non-root-scope standard library.

PR #660, opened 2024-02-27
Implement simple i8 arithmetic.

dfellis:
Proper support for generic types will be necessary before further progress can happen here (except if I want to create a bazillion bindings to the `getOrExit`


dfellis:
Now up to 57 passing tests (and 153 still failing, but progress!)

PR #665, opened 2024-03-01
Improve the get_type function to properly fail

dfellis:


src/program.rs
```diff
@@ -510,7 +512,7 @@ fn baseassignablelist_to_microstatements(
                             _ => unreachable!(),
                         }),
                         None => Err(format!("Couldn't find variable {}", var.iter().map(|segment| segment.to_string()).collect::<Vec<String>>().join(""))),
-                    }?;
+                    }??;
```
dfellis, line 515: I can't believe this works.

dfellis: Extra confused code here.

PR #666, opened 2024-03-02
Implement events

dfellis:
Implements threads, unning each event handler on a separate thread and copying the data they use,

Also turns `toI8` into just `i8` for consistency and sets up a new Microstatement type, Arg, to make variable resolution work for arguments, though it doesn't guard against argument names that are valid in Alan and not in Rust.

And several bugfixes along the way.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -103,7 +103,7 @@ macro_rules! status {
 // The only test that works for now
 test!(hello_world => r#"
     export fn main {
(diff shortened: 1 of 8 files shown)


dfellis:
All that work for just one passing test...

And surely it's a coincidence that the PR where I restored the event system and did it with multi-threading is #666 :japanese_ogre:

PR #667, opened 2024-03-02
Add a multi-line Hello, World test

dfellis:
I realized that while the syntax for strings makes them multi-line by default, there was no test to demonstrate that (and the vim highlighting broke on it, so I also fixed that).

And a tiny change on some todo lines so it's more clear what it is to be done in the future.


dfellis:
Screenshot from 2024-03-02 11-24-21
Doesn't that look so very FP? :)

PR #668, opened 2024-03-04
Implement basic method syntax and i16 math functions

dfellis:
Two of the tests for the i16 math used methods, so that's why I went with them to test the method syntax.

It currently only supports "direct" method access, not via properties, so something like `http.request("https://google.com")` would work but calling a method off of a property of a variable name would not (eg `http.request("https://google.com").response.status()` would fail to compile).

I've also set up grouping syntax (eg `(1 + 2)`) but since operators don't exist yet, that's not actually tested, yet.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -30,8 +30,8 @@ pub fn compile(source_file: String) -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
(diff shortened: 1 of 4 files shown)


dfellis:
> 67 passed; 144 failed

Progress! :)

PR #669, opened 2024-03-06
Implement operator support

dfellis:
testing with i64 math, and fix formatting,


dfellis:
Getting close to the halfway mark

> test result: FAILED. 82 passed; 129 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.88s

(though that's a bit misleading as some tests combine a *lot* into a single test script)

PR #670, opened 2024-03-07
Implement nested alan function calling, modifying the function generation to pass around an ordered hash map of functions indexed by the output function name, getting some more tests to pass (and adding an extra one)

dfellis:


dfellis:
> test result: FAILED. 84 passed; 128 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.99s

Inching closer to the halfway point.

PR #680, opened 2024-03-18
Implement a faster parmap

dfellis:
The prior implementation using `Vec::append` nullified *most* of the performance benefit of parallelization because of all of the copying and resizing, likely. (It was only 10-20% faster for large vectors)

This new implementation uses some unsafe shenanigans to write directly into the output vector, and is about 4.25x faster than running sequentially on my 16-thread CPU. There's probably some contention at play since I think there's only 8 *real* cores and it's hyperthreading (or whatever AMD's equivalent is called) causes contention between them, but I might also be hitting some memory bandwidth limits, as all it's doing is multiplying values by 2, wrapping in a `Result`, and storing that in the vector.

I have been hesitant for the generated Rust code to depend on any third-party libraries, but thinking about it, that will be the case once I start using `wgpu`, so I might look into a more advanced library to determine *actual* core count, instead of just using `available_parallelism`, assuming that cutting that value in half keeps the same performance (or potentially improves it). That will be a follow-up PR if the test goes well.

But I am still quite happy with perf now vs the 0.1 days :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,6 +19,7 @@ name = "alan"
 version = "0.2.0"
 dependencies = [
  "clap",
+ "divan",
  "nom",
  "ordered_hash_map",
 ]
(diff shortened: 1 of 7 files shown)


dfellis:
So I did a quick check dividing the `available_parallelism` value by 2 and there's no measurable impact on performance (up or down) so I do think for this test it's memory bandwidth constrained, but also that it may make sense to stick to the number of cpu cores instead of the number of cpu threads.

PR #682, opened 2024-03-22
Compile rust transpile as cargo projects

dfellis:
This took a lot longer than I expected (when attempting to keep the compilation time at least somewhat close to what it was before) because [`wgpu`](https://wgpu.rs), the library I am interested in using in the transpilation, takes >30sec to compile on my machine, and doing that for every single test is madness.

So after digging into how cargo's build cache works (terribly, btw), I'm making a singular project and serializing compilation through it via a lockfile, which isn't great, but at least the test suite finishes within a minute, still.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,11 @@ name = "alan"
 version = "0.2.0"
 dependencies = [
  "clap",
+ "dirs",
  "divan",
  "nom",
  "ordered_hash_map",
+ "sysinfo",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


dfellis:
The test suite is still passing 89 tests, so no regressions here, though it is over a minute on the CI machine.

> test result: FAILED. 89 passed; 127 failed; 0 ignored; 0 measured; 0 filtered out; finished in 63.32s

The benchmark also seems to be taking a bit longer than before because of the build time increase, but it's passing.

PR #683, opened 2024-04-03
Create the GPU type and necessary functions

dfellis:
Currently just confirming that the GPU struct actually builds correctly. Once there's some real usage of it the Debug printing of this type will go away and it'll be wholly opaque in the language

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,4 +1,5 @@
 name: Benchmarks
+runs-on: self-hosted
 
 on:
   push:
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,5 +1,4 @@
 name: Benchmarks
-runs-on: self-hosted
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


dfellis:
Hmm, this is gonna complicate things. The test suite is failing on the `hello_gpu` test in the CI, presumably because there's no GPU available in the CI worker.

I'm gonna have to do some reading up on things to see if/how GPU testing in Github Actions is accomplished.

dfellis:
Digging in on this, there is a waitlist for Github-run GPU runners that I could apply to, but it's focused on businesses, which this no longer is, or I could self-host a runner but they highly recommend that I don't allow that for public repos.

I think if I require manually starting the CI jobs for PRs, then I can do my own review on malicious intent before actually sending any work to the runner and I should be fine, then.

dfellis:
Success! :)

> test result: FAILED. 90 passed; 127 failed; 0 ignored; 0 measured; 0 filtered out; finished in 42.93s

Right now I'm running the runner inside of an ssh connection to my desktop machine.

Screenshot from 2024-04-04 11-11-04

I think I can "poor man's daemonize" this by just running it inside of a screen session, for now.

PR #684, opened 2024-04-06
Implement compound types (structs), and the bare minimum of property access and booleans to get another test working

dfellis:
I realized that I wanted to implement the basic GPU `run` function by passing in a type that has the wgsl shader code as a string and an array of arrays of GPU buffers (to represent the group and binding fields for mutable memory) that I could then build the better abstractions on top of, but that it was impossible because structs and arrays don't really exist, yet.

This PR gets structs working. It also partially implements arrays and the bare minimum of bool and property access to get another test working. I really need to clean up the property/method/etc logic into a properly recursive function instead of the spaghetti I have right now, but I wanted to reserve that refactor for later (both because I want to get a GPGPU benchmark working ASAP, regardless of the quality of that code there, but also because this PR is getting long in the tooth and I didn't want to make it too difficult to follow in the future).

Stopping here for tonight since it's getting late on a Friday.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -1540,22 +1540,18 @@ test!(array_literals => r#"
     stdout "1\n2\n4\n";
 );
 test!(object_literals => r#"
-    from @std/app import start, print, exit
-
     type MyType {
-      foo: string,
+      foo: String,
       bar: bool,
     }
 
-    on start {
+    export fn main {
       const test = new MyType {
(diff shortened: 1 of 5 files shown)


dfellis:
Now at

> test result: FAILED. 91 passed; 126 failed; 0 ignored; 0 measured; 0 filtered out; finished in 18.90s

91/126. One more test passing.

PR #686, opened 2024-04-09
Real GPGPU

dfellis:
Adding bindings and functions for buffer creation and reading from buffers. I'm starting to feel the restrictions this half-baked compilation layer is imposing on me, though.

Tomorrow I'll work on updating the benchmark to use this form. I don't expect it to be any faster than perhaps even the linear scan due to the extra memcpys involved, but I might be surprised.

After that I'll be going deep on syntax (both changes in syntax between Alan v0.1 and v0.2) as well as just implementing syntax that's currently broken (looking at you, conditionals and generics).

But definitely excited about this change :)

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -112,6 +112,7 @@ name = "alan_generated_bin"
 edition = "2021"
 
 [dependencies]
+flume = "0.11.0"
 futures = "0.3.30"
 wgpu = "0.19.3""#;
         let cargo_path = {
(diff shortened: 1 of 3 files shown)


dfellis:
Alright, the modified `hello_gpu` test passes. But I ran into an issue where the cached `Cargo.toml` was out-of-date and *most* code wasn't compiling, so I was getting weird results until I logged onto the server and blew away `~/.config/alan`.

I need to add a cache-invalidation mode if the `Cargo.toml` file in `~/.config/alan` doesn't match the one embedded in the binary.

src/compile.rs
```diff
@@ -746,8 +747,24 @@ test!(string_max => r#"
...
+    export fn main {
+      let g = GPU();
+      let b = g.createBuffer(storageBuffer(), filled(2.i32(), 4));
+      let plan = GPGPU("
+        @group(0)
+        @binding(0)
+        var<storage, read_write> vals: array<i32>;
+
+        @compute
+        @workgroup_size(4)
```
dfellis, line 759: I'll eventually update this to provide controls to let you decide how many workgroups you want to run. Then this length doesn't need to be hardwired in here (rather, it can be 1 and you can dynamically set the number of workgroups by the length of the array instead of modifying the size of the workgroup)

PR #687, opened 2024-04-09
Add some GPGPU benchmarking

dfellis:
I had to make some modifications to handle very large arrays. Surprisingly the default limit for buffers is just 256MB. I changed it from the default to whatever the limits are for the hardware/OS combo, instead and I got ~2GB per buffer instead. This still meant my 1 billion 32-bit integers buffer was too large, so the benchmark has suffered some complexity.

I think I can use the limits object to bolt on some safety to the API, but the fact that wgpu defaults to `panic!` so much makes me uneasy. But perhaps that is impossible to avoid with GPUs right now?


dfellis:
Same set of passing tests.

PR #693, opened 2024-05-10
Drop events from the language, add parsing support for top-level keyword generics

dfellis:
First step on following the plan in the new RFC. Events and Handlers are gone from Alan v0.2, which drops a lot of code (and all tests have been converted from event handler syntax that still had it).

This also adds parsing support for generics on the top-level keywords, `import`, `export`, `type`, `fn`, `const`, `let`, `prefix`, `infix`, (and adding `postfix` that was missing), so we can eventually add conditional compiling of them.

The generics support does not yet support operators or method syntax to unify compile-time and run-time execution syntax, that will come later.


dfellis:
I also found out that `todo!` panics and that causes issues with the test suite when combined with the `.lockfile`, so I converted the ones I could find to returning errors and properly bubbling to the surface, which made the test suite much smoother.

I do wonder *how* it was finishing the test suite, before.

PR #696, opened 2024-05-14
Rewrite the baseassignablelist parsing

dfellis:
I realized that a lot of the maintenance burden with the baseassignablelist parsing function was that the method separator (`.`) was part of the *variable* rather than another element in the baseassignablelist, which significantly complicated any attempts at turning the it into a state machine. Swapping it over has bloated the code a bit, but I have more branches in the code implemented, as well, and fewer weird hacks.


dfellis:
> test result: FAILED. 88 passed; 123 failed; 0 ignored; 0 measured; 0 filtered out; finished in 23.00s

All 88 tests still passing after this reworking.

PR #699, opened 2024-05-14
Convert *most* of the codebase to property-like method syntax

dfellis:
This converts most of the code to drop the empty parens `()` at the end of a method call. A few examples still persist and this is necessary for standard function calling, but it makes getter functions effectively identical to properties from a syntactic perspective, which makes method syntax with constructor functions read very similarly to just specifying units for a value.

Doing this demonstrated a bug in the implementation, which has also been fixed.


dfellis:
All of the currently passing tests still pass with this change.

src/compile.rs
```diff
@@ -1250,7 +1250,7 @@ test!(basic_function_usage => r#"
 
     export fn main {
       foo();
-      'foo'.bar().print();
+      'foo'.bar.print;
```
dfellis, line 1253: This is the line that was causing an extra failure before. It was not properly handling method syntax on constants, actually, not the chaining of two sequential method calls without the parens like I originally thought.

PR #728, opened 2024-05-31
Properly implement generic types and function types

dfellis:
This is a pretty massive refactoring of the internals of the type system, switching large chunks of it from being stringly-typed to properly typed with CType trees.

Now generic type definitions work as you'd expect them to and higher-order functions *properly* check that the function being passed to it actually works!

### DIFF
--- a/src/program.rs
+++ b/src/program.rs
@@ -26,7 +26,8 @@ impl Program {
         // Load the entry file
         p = match p.load(entry_file) {
             Ok(p) => p,
-            Err(_) => {
+            Err(e) => {
+                panic!("{}", e);
                 // Somehow, trying to print this error can crash Rust!? Really not good.
                 // Will need to figure out how to make these errors clearer to users.
                 return Err("Failed to load entry file".into());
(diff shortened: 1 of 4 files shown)


src/program.rs
```diff
@@ -26,7 +26,8 @@ impl Program {
         // Load the entry file
         p = match p.load(entry_file) {
             Ok(p) => p,
-            Err(_) => {
+            Err(e) => {
+                panic!("{}", e);
```
dfellis, line 30: I should revert this change.

src/program.rs
```diff
@@ -2346,8 +2542,8 @@ fn statement_to_microstatements(
 #[derive(Clone, Debug)]
 pub struct Function {
     pub name: String,
-    pub args: Vec<(String, String)>, // Making everything Stringly-typed kinda sucks, but no good way to give an error message in the parser for unknown types otherwise
-    pub rettype: Option<String>,
+    pub args: Vec<(String, CType)>, // Making everything Stringly-typed kinda sucks, but no good way to give an error message in the parser for unknown types otherwise
```
dfellis, line 2545: This comment is out of date.

PR #744, opened 2024-06-09
Rework adapter selection

dfellis:
While working with the new RISC-V SBC, I ran into gpu-related problems. I have not yet resolved them, but along the way I updated to the latest version of `wgpu`, added `wgpu`'s tracing support (eventually will make that a build option, but for now it's just on by default), and figured out that I was choosing an adapter in an incorrect way (not considering whether or not the driver even has everything necessary to support WebGPU; I thought they just wouldn't return those options, but apparently you're allowed to use them if you know what you're doing and avoid the parts of the WebGPU spec that they're not compliant with).

So I still consider it a bit of a win on this front. I do hope to get it actually running GPGPU logic on that SBC, though, but it's a lower priority.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -334,9 +334,9 @@ dependencies = [
 
 [[package]]
 name = "regex-lite"
-version = "0.1.5"
+version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


dfellis:
Hmm... That's unexpectedly failing on my x86 machines, too. Is there a regression in 0.20.0? Or perhaps I've been doing something incorrectly this whole time with their API and it worked by chance before?

dfellis:
Going to move the `wgpu` upgrade to another PR (despite the name of this branch).

PR #745, opened 2024-06-10
Upgrade `wgpu` to `0.20.0` for real

dfellis:
Working around a bug in `copy_buffer_to_buffer` by including `MAP_WRITE` on the "normal" buffer type


dfellis:
The reason why this code works: https://github.com/gfx-rs/wgpu/issues/5795#issuecomment-2159139603

PR #763, opened 2024-06-11
Switch to wrapping arithmetic

### LINKED ISSUE #762: Switch default arithmetic to wrapping
Necessary to be compatible with GPGPU arithmetic, which is wrapping, as well.

The upcoming match RFC will cover where numbers will be going in the future (preview of my plans for the RFC on my Mastodon), but the Result-based arithmetic will be going away.

dfellis:
Resolves #762


dfellis:
...and the planned power outage in my neighborhood happened right now, so the test suite can't run (since I host my own runners to get GPU access).

I guess I'm taking a break for a bit.

PR #770, opened 2024-06-20
Fix benchmarks on windows

dfellis:
Now that I have a little Intel NUC to use for testing Windows, I manually tested and noticed a small problem with the benchmarks. This should fix that and also enable testing with it. Since I have only one Windows machine, I have also trimmed the work done on Windows vs Linux to try and reduce an increase in latency. I have also forced the normal testing path to use Linux so it doesn't accidentally clog up the Windows machine's queue.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -10,10 +10,8 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  build:
-
-    runs-on: self-hosted
-
+  bench:
+    runs-on: [self-hosted, linux]
     steps:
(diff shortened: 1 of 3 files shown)


dfellis:
Screenshot from 2024-06-20 00-14-40

Not sure how much of this slowdown is due to Windows, how much is due to this being a NUC (though it's good a good SSD, 32GB of RAM, etc...), etc, but the benchmarks take 50% more time and the test suite is about 150% slower, and these two have to run sequentially, taking the testing time up from 4 minutes to 11 minutes. :/

But it works!

PR #772, opened 2024-06-21
Implement generic functions (and random cleanup along the way)

### LINKED ISSUE #748: Generic Functions
Generic functions of the form `fn <name>{generic, args} (...` should be added.

Similar to how generic types are implemented, the generic function only exists as an AST waiting to be serialized with a hidden scope mapping the generic type names with the actual type names that will take their place.

This approach does lead to generated binary bloat, but:

1. That's what Rust does, anyways, so even if we kept it generalized, it would still end up that way.
2. It's much easier to typecheck that way.

dfellis:
I'm not 100% liking this result, and since I've been working on it so long it's accumulated other random things, like improvements to the rust library updating logic, that probably should have been separate PRs, but I didn't want to do the work to disentangle it.

There's still some string-based typing in the depths of this codebase, and since I needed to change that string-based typing I touched a lot more files than I *should* have, but it is working and working faster than before.

Resolves #748

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -305,9 +305,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.85"
+version = "1.0.86"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "22244ce15aa966053a896d1accb3a6e68469b97c7f33f284b99f0d576879fc23"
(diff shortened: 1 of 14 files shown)


dfellis:
Hmm... It seems the file locking behavior on Windows is different versus Linux/macOS, so I'll need to mess with that a bit to figure out how to get it working (which I really want to do because this locking improvement cuts the test time almost in half).

PR #800, opened 2024-07-10
Drop ok, err, some (both kinds), none from language

dfellis:
Alan v0.2 doesn't really need these functions for the optional and fallible types. You can just `Maybe(val)` to be equivalent to `some(val)`, `Maybe{myType}(void)` to be equivalent to `none` (a bit more awkward, type inference should eventually support `Maybe(void)` for `none`), `Fallible(val)` for `ok(val)` and `Fallible{myType}(errorVal)` for `err(errorVal)` (again, type inference may help here). And the `Maybe` and `Fallible` "functions" could be bound to `?` and `!` postfix operators to turn that into just `val?`, `void?` (assuming inference works as desired), `val!`, and `errorVal!` (again assuming inference works as desired) which is even shorter.

At the same time, many languages have a different meaning for `some`, as an array method for determining if any of the elements of the array passes the provided check function. Alan since v0.1 includes a `has` function that accepts a value directly and returns true if that value is present in the array. This is like a shorthand for a check function that performs an equality check on the value, so I decided to expand `has` to also accept check functions, which therefore made it equivalent to `some` in those languages.

Since this language is aimed at both crowds, completely eliminating these terms from the language may be initially confusing, but should hopefully be less confusing than having a known term with a different meaning.

### DIFF
--- a/src/compile/integration_tests.rs
+++ b/src/compile/integration_tests.rs
@@ -1546,26 +1546,6 @@ test!(array_every => r#"
     }"#;
     stdout "false\n";
 );
-test_ignore!(array_every_some_del => r#"
-    fn isOdd (val: int64) -> bool = val % 2 == 1;
-
-    export fn main {
-      const test = [ 1, 1, 2, 3, 5, 8 ];
-      test.every(isOdd).print;
-      test.some(isOdd).print;
-      print(test.length);
-      print(test.delete(1));
-      print(test.delete(4));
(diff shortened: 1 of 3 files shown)


dfellis:
The Windows machine decided to install updates (probably) and took itself out, while I accidentally screwed up my ssh access just before going on a long trip, so... I guess I will have to spot-check Windows every so often in a VM?

PR #808, opened 2024-07-13
Array and Buffer (kinda) 'repeat' implemented

dfellis:
As noted in one of the comments, the type system can't currently handle repeating buffers an arbitrary number of times because the value is a runtime value. The type system needs to be improved to be able to distinguish the literal `3` from a variable `three` that may or may not have the value `3` inside of it before this can be improved.

Then we can have both the current implementation for runtime variables, which *has* to convert into an array to have a static type, and a new implementation for static values that can convert into a buffer on the other side, as well.

### DIFF
--- a/src/compile/integration_tests.rs
+++ b/src/compile/integration_tests.rs
@@ -1474,10 +1474,10 @@ test!(array_map => r#"
     }"#;
     stdout "1, 2, 3, 4, 5\n2, 4, 6, 8, 10\n";
 );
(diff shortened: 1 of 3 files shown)


dfellis:
I really need to figure out why Windows occasionally errors out with this message:

> Error: "Failed to compile Os { code: 33, kind: Uncategorized, message: \"The process cannot access the file because another process has locked a portion of the file.\" }"

A second run succeeds, so I suspect something to do with the hackery to get file locking working on Windows.

PR #812, opened 2024-07-20
Add unsigned integers and partial reworking of GPU types

dfellis:


dfellis:
My home lost power, so none of the tests will run right now, so just going off of what my laptop tells me. (It's good to go.)

PR #821, opened 2024-07-29
Closure functions actually enclose outer-scope variables

dfellis:
This fixes an issue with closures actually using variables from the outer scope by making sure they're available as owned values within the closure. This is a super trivial implementation that just `clone`s everything from the outer scope, but it works, and appears to have the unused `clone`s deleted during LLVMs optimization passes (probably just dropping all unused variables in their entirety?) "If it's stupid but it works..."

### DIFF
--- a/benches/bench.ln
+++ b/benches/bench.ln
@@ -14,76 +14,64 @@ fn bench(l: i64) {
   v2[0];
   t2.elapsed.print;
   let t3 = now();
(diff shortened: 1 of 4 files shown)


dfellis:
Whelp, it doesn't quite work as-is. Now to decide if/how to reduce the set of variables to clone...

PR #834, opened 2024-08-06
Better enum construction for void type values

dfellis:
This feels more "Alan" to me (there's no special function to make a void-value Either type anymore, just call the normal constructor function with no values), simplifies the codebase a bit, and fixes a bug in the rust codegen.

### DIFF
--- a/src/lntors/typen.rs
+++ b/src/lntors/typen.rs
@@ -157,7 +157,7 @@ pub fn generate(
             CType::Either(_) => {
                 let res = generate(t, out)?;
                 out = res.1;
-                out.insert(name.clone(), ctype_to_rtype(typen, false)?);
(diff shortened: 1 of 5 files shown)


src/lntors/typen.rs
```diff
@@ -157,7 +157,7 @@ pub fn generate(
             CType::Either(_) => {
                 let res = generate(t, out)?;
                 out = res.1;
-                out.insert(name.clone(), ctype_to_rtype(typen, false)?);
+                out.insert(t.to_callable_string(), ctype_to_rtype(typen, false)?);
```
dfellis, line 160: This was a bug that allowed the same type to accidentally be created twice, because the keys could differ.

PR #937, opened 2024-10-22
Reduce testing latency (especially in Windows/MacOS) by removing child process spawning for compilation

dfellis:


dfellis:
# Before

Screenshot from 2024-10-22 16-02-03

# After

Screenshot from 2024-10-22 16-48-45

That's 33% faster for x86-64 Linux, 38% faster for ARM64 Linux, 27% faster for Windows, and 40% faster for MacOS.

Surprisingly less important for Windows than I expected, but good improvements all around.

PR #945, opened 2024-11-01
Refactor the Program type to be used as a singleton

dfellis:
Also make the environment frozen for the program on startup, which allows the test suite to run "normally" (multithreaded) because it doesn't depend on the single-threaded env type anymore.

This is in preparation for allowing imports of other source files to finally work.


dfellis:
Oh, and I bumped the version of Rust I'm using here to at least 1.80.0, so I need to update Rust on all of my test runners before tests will pass.

PR #949, opened 2024-11-04
Drop the global_id usage from wgpu, since it no longer exists on the latest version

dfellis:
A necessary pre-req before I can upgrade `wgpu`


dfellis:
Oh, right, part of this depends on the update to the library, too. Let me try and attach it to the dependabot PR.

PR #955, opened 2024-11-12
Allow calling higher-order generic functions with a generic function as the argument

dfellis:
This implementation is hacky at the moment, so I'm not sure how well it'll work, but it does pass with the provided test case.

### DIFF
--- a/alan/src/compile/integration_tests.rs
+++ b/alan/src/compile/integration_tests.rs
@@ -2592,6 +2592,24 @@ test_full!(generic_functions => r#"
 "#;
     stdout "[]\n";
 );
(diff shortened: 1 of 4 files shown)


dfellis:
Well crap. I broke a lot of things getting that new test working. I need to dig into how/why that happened. Not sure this'll get merged today.

dfellis:
So I forgot that `Call{N, F}` functions have an auto-generated name that's almost 100% guaranteed to differ from the name the user binds them to, so if the function's `name` property doesn't actually match a function added into the `Scope`, just don't replace the name and that seems to have fixed it.

PR #956, opened 2024-11-13
Implement `clz` (count leading zeroes) in Javascript

dfellis:


dfellis:
Figuring out the binary search implementation for `clz` was a fun puzzle. :)

PR #966, opened 2024-11-20
Convert the arithmetic tests to pure Alan

dfellis:
Not converting everything yet because:

1. Maybe there's a mistake that requires a refactor. This PR is already large enough as it is if that's the case.
2. I don't have a good story for doing the GPGPU tests, so I *can't* convert those yet, and those are the tests I ran into.
3. I'm curious how much just this saves. :)


dfellis:
Even just this small amount has drastically improved things.

From this:

Screenshot From 2024-11-19 19-28-36

To this:

Screenshot From 2024-11-19 20-43-38

That's a *48%* reduction in runtime for Linux, *44%* for Windows, *31%* for MacOS, and *35%* for ARM Linux. And that's only removing the arithmetic tests from the Rust test suite to the new Alan test suite (I will note that it was a kludge that was also intended to end-to-end test the compiler, but still).

That did reduce the number of tests in the Rust CI from 250 to 139, which is a *44.4%* decrease on that side, implying that for my Linux desktop the testing bottleneck is almost 100% due to compilation time, while for my older hardware (the 2014 Mac mini and the Raspberry Pi) there is some impact from actually running the tests, as well.

PR #974, opened 2024-11-21
Implement `ones` method for JS integer types

dfellis:


dfellis:
Well, that implementation didn't *quite* work, because I wasn't mutating the original value. With that fixed it's still faster, but not quite as much.

Screenshot From 2024-11-21 15-08-25

PR #1031, opened 2024-12-28
Eliminate many Function clones

dfellis:
This speeds up compilation of the `Hello, World!` app by ~33% both natively and in the browser.

That still puts it in the seconds range, which isn't great, but baby steps. :)

The majority of the time left is almost evenly split between generating Rust/JS-compatible names for types/functions and actually constructing the function objects. I'm going to dig in a bit on the former to see if I can speed that up (it's kinda crazy how long that's taking), but for the latter the only solution is to make function creation lazy, which I have been avoiding, but maybe I shouldn't?


dfellis:
The test suite perf impact is much more modest at around 5-10% faster. Probably because of the lockfile contention serializing the compilation between runs.

PR #1042, opened 2025-01-08
Implement `storageBarrier`

dfellis:
This supposedly will fail on Metal since each execution is being done in a separate workgroup and Metal doesn't provide `storageBarrier` guarantees across workgroups (despite the naming and there being a `workgroupBarrier` that should be the one constrained per workgroup) but I figured I may as well test it out with my two Apple machines and see what happens.

I may still discourage using `storageBarrier` since making a pipeline with two shaders that is guaranteed to work as expected is nearly identical in Alan instead of a big pain-in-the-butt in normal WebGPU.

Eg

```rs
[
  someGpuCompute(),
  storageBarrier(),
  someMoreCompute()
].build.run;
```

vs

```rs
[
  someGpuCompute().build,
  someMoreCompute().build,
].run;
```

But if this does work, it is theoretically faster and can be used in loops of indeterminant iterations (though I don't expose GPU loops, yet).

### DIFF
--- a/alan_compiler/src/std/root.ln
+++ b/alan_compiler/src/std/root.ln
@@ -4992,7 +4992,7 @@ fn lte(a: gvec4i, b: gvec4i) = glte{gvec4i, gvec4b}(a, b);
 fn lte(a: gvec4f, b: gvec4f) = glte{gvec4f, gvec4b}(a, b);
 
 fn ggt{I, O}(a: I, b: I) {
-  let varName = '('.concat(a.varName).concat(' < ').concat(b.varName).concat(')');
+  let varName = '('.concat(a.varName).concat(' > ').concat(b.varName).concat(')');
   let statements = a.statements.concat(b.statements);
   let buffers = a.buffers.union(b.buffers);
   return {O}(varName, statements, buffers);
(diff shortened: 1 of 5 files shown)


alan_compiler/src/std/root.ln
```diff
@@ -4992,7 +4992,7 @@ fn lte(a: gvec4i, b: gvec4i) = glte{gvec4i, gvec4b}(a, b);
 fn lte(a: gvec4f, b: gvec4f) = glte{gvec4f, gvec4b}(a, b);
 
 fn ggt{I, O}(a: I, b: I) {
-  let varName = '('.concat(a.varName).concat(' < ').concat(b.varName).concat(')');
+  let varName = '('.concat(a.varName).concat(' > ').concat(b.varName).concat(')');
```
dfellis, line 4995: No matter what, this fix (and the one below) will be kept, as it confused the crap out of me until I inspected the generated shader code and realized something was very wrong.

dfellis:
Confirmed that `storageBarrier` works on everything except Apple Silicon GPUs (even Metal on the old Intel Macs works correctly).

So I'm not sure if I should actually not include `storageBarrier` or not.

PR #1072, opened 2025-02-10
Fix some bugs in the JS window logic discovered primarily when changes to the page happen after the initial load

dfellis:


alan_std.js
```diff
@@ -1058,7 +1058,9 @@ export function frameFramebuffer(frame) {
 
 export async function runWindow(initialContextFn, contextFn, gpgpuShaderFn) {
   // None of this can run before `document.body` exists, so let's wait for that
-  await new Promise((r) => document.addEventListener("DOMContentLoaded", () => r()));
+  if (document.readyState !== "complete" && document.readyState !== "loaded") {
+    await new Promise((r) => document.addEventListener("DOMContentLoaded", () => r()));
+  }
```
dfellis, line 1063: If this window is started *after* the initial loading of the page, this `await` would cause it to hang forever before, so now check if we actually need the `await`.

alan_std.js
```diff
@@ -1083,15 +1085,15 @@ export async function runWindow(initialContextFn, contextFn, gpgpuShaderFn) {
     canvas.style['height'] = '100%';
     canvas.width = window.innerWidth;
     canvas.height = window.innerHeight;
-    document.body.addEventListener("resize", () => {
+    window.addEventListener("resize", () => {
```
dfellis, line 1088: Apparently this event never fires on the `body`, only the top-level `window` object. (Especially confusing because I do remember seeing documentation stating otherwise.)

alan_std.js
```diff
@@ -1083,15 +1085,15 @@ export async function runWindow(initialContextFn, contextFn, gpgpuShaderFn) {
...
       canvas.height = window.innerHeight;
     });
-    if (!context.cursorVisible) {
-      canvas.style['cursor'] = 'none';
-    }
     context.canvas = canvas;
   }
+  if (!context.cursorVisible) {
+    context.canvas.style['cursor'] = 'none';
+  }
```
dfellis, line 1096: Moving this out of the other `if` block so it can be applied to a canvas provided by the user, not only one auto-generated by the code here.

PR #1161, opened 2026-04-30
Automatic GPGPU workgroup sizing

dfellis:
For *much* better performance.


dfellis:
Oof. So this PR finally changes the API between `alan_std` and the root scope, which means pinning x86 MacOS to the last working version of `wgpu` on those machines no longer works. I'll need to port these API changes to that branch to make the test suite pass.

So, that doubles the maintenance burden here. Do I

1. Cherry-pick changes from this branch onto that branch and try and get this over the line?
2. Decide to jettison x86 MacOS as unsupported, since Apple themselves are no longer supporting it, and `wgpu` has stopped supporting it a year or so ago.

dfellis:
Salute x86 MacOS

So long x86 MacOS. You served us well. (Well, better than Windows, at least.)

